PR #31, opened 2022-07-13
fix:complex calls

kanej:
To allow `m.call(...)` calls against overloaded functions we now support the following syntax:

```javascript
m.call(resolver, "setAddr(bytes32,address)", {
  args: [resolverNode, resolver],
});
```

## Technical details

The invocation using the `setAddr(bytes32,address)` was fine, but failed the initial validation. This fix is to the validation code.

### DIFF
--- a/packages/core/src/executors/CallExecutor.ts
+++ b/packages/core/src/executors/CallExecutor.ts
@@ -6,6 +6,7 @@ import { Services } from "../services/types";
 import { Tx } from "../types";
 
 import { Executor } from "./Executor";
+import { mapToAddress } from "./utils";
 
 export class CallExecutor extends Executor<CallOptions, Tx> {
   public async execute(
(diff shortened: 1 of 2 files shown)


MatiFalcone:
I have the following module:

```
function createPoap(m: IgnitionModuleBuilder) {

  const proxyAdminOwner = m.getAccount(0);

  const name = m.getParameter('name');
  const symbol = m.getParameter('symbol');

  const poapContract = m.contract('Poap', [
    name,
    symbol,
    proxyAdminOwner
  ]);

  // Initialize
  const baseUri = m.getParameter('baseUri');
  m.call(poapContract, "initialize", [baseUri, []]);

  return { poapContract };
}
```

And I am getting this error:

```
Poap#Poap.initialize:
 - IGN710: Function 'initialize' is overloaded in contract Poap. Please use one of these names instead:

 - initialize()
 - initialize(string,address[])
 - initialize(address)
 ```
 
But when I changed it to the function I want to execute (`initialize(string,address[])`) I get:

```
IgnitionError: IGN702: Module validation failed with reason: The function name "initialize(string,address[])" is invalid, make sure you use a valid identifier.
```

How can I execute that function?

kanej:
Can you paste a version of the Solidity code of your contract?

Are you able to invoke the `initialize(address)` version if you change the parameter list? To exclude the issue being a bug with Ignition dealing with `address[]` incorrectly.

matias-gmg:
The version of Solidity I am using is `0.8.24` and when I call `initialize(address)` everything works as expected.

matias-gmg:
@kanej based on [this issue](https://github.com/NomicFoundation/hardhat-ignition/issues/762), it looks like there is a bug with Ignition dealing with any `[]` type.

SyedAsadKazmi:
Kindly please see to it, @kanej. Thanks.

PR #42, opened 2022-10-07
docs: add a uniswap example

kanej:
Add a uniswap example based on the `uniswap-v3-deployer`.

## TODO

* [x] add multirecipe example
* [x] add interesting tests that work against uniswap


morgansliman (approved):
Nice! LGTM

PR #69, opened 2022-11-18
refactor: make deploy return type explicit

kanej:
This is pulling out a small change from the docs PR as it is a code change.

The return type of deploy on the Ignition wrapper now has an explicit type.

**Note**: the return value of serialized result for string and number has been changed to the value itself. I cannot find a current path that would lead to this code being used, but I am going to defer the cleanup to a later refactor.


packages/hardhat-plugin/src/ignition-wrapper.ts
```diff
@@ -1,26 +1,41 @@
...
-  private _ignition: Ignition;
+interface Ignition {
+  deploy<T extends ModuleDict>(
+    ignitionModule: Module<T>,
+    deployParams?: {
+      parameters?: { [key: string]: ExternalParamValue };
+      ui?: boolean;
+    }
+  ): Promise<DeployResult>;
+}
```
morgansliman, line 24: what's the purpose of this interface?

kanej: Ah good point. I pulled it out so I could document it in the docs, but maybe it should now disappear again. I'll make that change.

PR #70, opened 2022-11-18
feat:restart-deployment

kanej:
Enable a basic form of restarting the deploy (on failed transaction or on hold) by leveraging a journal file.

The journal file is written as a twin of the module file. It uses ndjson to record the messages sent during the execution stage. A rerun of the module rebuilds the execution state based on the journal.

WARNING: there are currently no checks to ensure the structure of the module is "close enough" on the rerun to allow.

## TODO

* [x] Write vertex updates to journal file
* [x] Merge previous journal entries into the state of the executor
* [x] Update the multisig example to show a restart
* [x] Update cli UI to deal with restarts
* [x] Add on hold state
* [x] Make awaitEvent return hold state
* [x] Update cli UI to deal with on hold
* [x] Update the docs


packages/core/src/deployment/Deployment.ts
```diff
@@ -1,177 +1,205 @@
...
+    return Object.values(this.state.execution.vertexes).some(
+      (v) => v.status === "HOLD"
+    );
+  }
+
+  private async _runDeploymentCommand(
+    logMessage: string | [string, any[]],
+    command: DeployStateCommand
+  ): Promise<void> {
+    log.apply(this, typeof logMessage === "string" ? [logMessage] : logMessage);
```
morgansliman, line 194: I'm not sure I really love this as a solution, but I can't think of a better alternative right now either so it's fine

morgansliman (approved):
This is great! not sure what the conflicting files are, but aside from those this looks fine to merge :+1:

packages/core/src/deployment/Deployment.ts
kanej: I agree, we should settle for my mediocrity, while we come up with something better. 😄

PR #71, opened 2022-11-21
multisig example and awaitEvent

morgansliman:
will wait to add docs until we get the on hold state so it works properly

### DIFF
--- a/examples/multisig/test/Multisig.js
+++ b/examples/multisig/test/Multisig.js
@@ -1,32 +1,33 @@
 const { expect } = require("chai");
+const { ethers } = require("ethers");
+
 const Multisig = require("../ignition/Multisig");
-const namehash = require("eth-ens-namehash");
-const labelhash = (label) =>
-  hre.ethers.utils.keccak256(hre.ethers.utils.toUtf8Bytes(label));
 
 const ACCOUNT_0 = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266";
-const ACCOUNT_1 = "0x70997970C51812dc3A010C7d01b50e0d17dc79C8";
 
 describe("Multisig", function () {
-  it("should be able to create new subrecords", async function () {
-    // Arrange
-    const { ens, resolver } = await ignition.deploy(ENSModule);
+  let multisig;
 
-    await ens.setSubnodeOwner(
-      namehash.hash("test"),
-      labelhash("alice"),
-      ACCOUNT_0
+  before(async () => {
+    const moduleResult = await ignition.deploy(Multisig);
+
+    multisig = moduleResult.multisig;
+  });
+
+  it("should store a submitted transaction", async function () {
+    const submittedTx = await multisig.functions.transactions(0);
+
+    expect(submittedTx.destination).to.equal(ACCOUNT_0);
+    expect(submittedTx.value.toString()).to.equal(
+      ethers.utils.parseUnits("50").toString()
     );
+    expect(submittedTx.data).to.equal("0x00");
+    expect(submittedTx.executed).to.equal(false);
+  });
 
-    // Act
-    await resolver["setAddr(bytes32,address)"](
-      namehash.hash("alice.test"),
-      ACCOUNT_1
-    );
+  it("should confirm a stored transaction", async function () {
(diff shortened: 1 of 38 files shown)


examples/multisig/test/Multisig.js
```diff
@@ -0,0 +1,32 @@
...
+const Multisig = require("../ignition/Multisig");
+const namehash = require("eth-ens-namehash");
+const labelhash = (label) =>
+  hre.ethers.utils.keccak256(hre.ethers.utils.toUtf8Bytes(label));
+
+const ACCOUNT_0 = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266";
+const ACCOUNT_1 = "0x70997970C51812dc3A010C7d01b50e0d17dc79C8";
+
+describe("Multisig", function () {
+  it("should be able to create new subrecords", async function () {
```
kanej, line 11: Do we need to update this test?

morgansliman:
Ah! I did forget to come back to this. The test won’t work until we have on
hold state anyways, so I could just fix this test when I add docs for
awaitEvent. if you’d rather have the test fixed for this PR i can do that
too, just lmk!

packages/core/src/services/TransactionsService.ts
```diff
@@ -18,4 +22,25 @@ export class TransactionsService implements ITransactionsService {
...
+  public async waitForEvent(
+    filter: ethers.EventFilter,
+    durationMs: number
+  ): Promise<ethers.providers.Log | null> {
+    const provider = new ethers.providers.Web3Provider(
+      this._providers.ethereumProvider
+    );
+
+    return new Promise((resolve) => {
+      const timeout = setTimeout(() => {
```
kanej, line 35: Can I check my understanding here.

We wait for event, __for some amount of time__. Then fail if it hasn't occurred?

What happens if an event has already happened before `provider.once`, does it return immediately?

packages/core/src/validation/dispatch/validateAwaitEvent.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const eventFragments = iface.fragments
+    .filter((frag) => frag.name === vertex.event)
+    .concat(events);
+
+  if (eventFragments.length === 0) {
+    return {
+      _kind: "failure",
+      failure: new Error(
+        `Contract '${contractName}' doesn't have an event ${vertex.event}`
```
kanej, line 44: Very nice

packages/core/src/validation/dispatch/validateAwaitEvent.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  return {
+    _kind: "success",
+    result: undefined,
+  };
+}
+
+async function resolveArtifactForCallableFuture(
```
kanej, line 77: We should pull this out as a helper function, I think it is a duplicate.

packages/core/src/execution/dispatch/executeAwaitedEvent.ts
```diff
@@ -0,0 +1,53 @@
...
+  resultAccumulator: Map<number, VertexVisitResult | null>,
+  { services, options }: ExecutionContext
+): Promise<VertexVisitResult> {
+  const resolve = resolveFrom(resultAccumulator);
+
+  const resolvedArgs = args.map(resolve).map(toAddress);
+
+  const { address, abi } = resolve(contract);
+
+  let log;
```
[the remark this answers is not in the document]
morgansliman: I'll rename it!

kanej (commented):
This is looking really solid.

A few small comments/todos. Lets update the await section of the docs as well.

packages/core/src/services/TransactionsService.ts
morgansliman: > We wait for event, for some amount of time. Then fail if it hasn't occurred?

correct

> does it return immediately?

also correct. if an event matching the provided filter has happened before that event listener is registered, it will get triggered

packages/core/src/validation/dispatch/validateAwaitEvent.ts
morgansliman: ah yeah you're probably right

kanej (approved):
Looks good. Lets get this merged.

PR #73, opened 2022-11-23
Fix plan bug for uniswap example

morgansliman:


kanej (approved):
Looks good.

We should ponder how to add a safeguard test for when the vertexes list changes - maybe that is really a test of the `isDeploy` that will throw if there are new types. I am not sure.

PR #74, opened 2022-11-24
added params section in cli ui

morgansliman:
Screen Shot 2022-11-24 at 7 53 08 AM

### DIFF
--- a/packages/hardhat-plugin/src/ui/components/execution/ModuleParameters.tsx
+++ b/packages/hardhat-plugin/src/ui/components/execution/ModuleParameters.tsx
@@ -1,0 +1,34 @@
+import type { ModuleParams } from "@ignored/ignition-core";
+import { Text, Newline } from "ink";
+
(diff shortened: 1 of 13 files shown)


packages/hardhat-plugin/src/ui/components/execution/ModuleParameters.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export const ModuleParameters = ({
+  moduleParams,
+}: {
+  moduleParams?: ModuleParams;
+}) => {
+  if (moduleParams) {
+    const params = [];
+
+    for (const [key, value] of Object.entries(moduleParams)) {
```
kanej, line 12: pipe the `.entries` into a map, to produce the `<Text></Text>` array. I wouldn't hold it against you if you did it inline on the template.

kanej (approved):
A few suggested, tiny tweaks, but otherwise good to go.

PR #75, opened 2022-11-24
added config for pollingInterval

morgansliman:


packages/hardhat-plugin/test/fixture-projects/with-config/hardhat.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
   ignition: {
     maxRetries: 1,
     gasIncrementPerRetry: BigNumber.from(1000),
+    pollingInterval: 4,
```
kanej, line 16: Lets show the default of 300. Just in case someone thinks its like 4 seconds.

kanej (commented):
Looks good. Lets add the config property into the documentation as well.

packages/hardhat-plugin/test/fixture-projects/with-config/hardhat.config.js
morgansliman: I was intentionally not using the default here so I could test that the user value was being set and properly overwriting the default. The default is written in the docs section for the config where the other default values are listed as well. 

If you'd still like it changed here I can do it, but that's the reasoning for why it's not currently that way.

morgansliman: nevermind, i just realized i forgot to push the doc changes i added, so take the above comment as what will soon be the case

kanej: Ah that makes complete sense. Ignore my request.

kanej (approved):
LGTM!

PR #76, opened 2022-11-30
getBytesForArtifact implementation

morgansliman:


packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
```diff
@@ -202,7 +202,9 @@ async function convertArgs(
   const resolvedArgs = [];
 
   for (const arg of args) {
-    const resolvedArg = await resolveParameter(arg, transformContext);
+    const resolvedArg = isBytesArg(arg)
```
kanej, line 205: I was going to suggest creating a wrapping function here, but it would end up called `resolved` or something equally awful. I think we do nothing, wait for another case then abstract.

kanej (approved):
This is really great. LGTM!

PR #79, opened 2022-12-14
added ability to send a normal ETH transaction

morgansliman:
Support sending `ETH` to a contract without having to make a call/deploy:

```tsx
m.sendETH(exchange, {
  value: ethers.utils.parseUnits("1"),
});
```

### DIFF
--- a/packages/core/src/execution/dispatch/executeSendETH.ts
+++ b/packages/core/src/execution/dispatch/executeSendETH.ts
@@ -1,0 +1,38 @@
+import type { PopulatedTransaction } from "ethers";
+
+import { ExecutionContext } from "types/deployment";
+import { SentETH } from "types/executionGraph";
+import { VertexVisitResult } from "types/graph";
+
+import { resolveFrom, toAddress } from "./utils";
+
+export async function executeSendETH(
+  { address, value }: SentETH,
+  resultAccumulator: Map<number, VertexVisitResult | null>,
+  { services, options }: ExecutionContext
+): Promise<VertexVisitResult> {
+  const resolve = resolveFrom(resultAccumulator);
+
+  const to = toAddress(resolve(address));
+
+  let txHash: string;
+  try {
+    const tx: PopulatedTransaction = { to, value };
+
+    txHash = await services.contracts.sendTx(tx, options);
+  } catch (err) {
+    return {
+      _kind: "failure",
+      failure: err as any,
+    };
+  }
+
+  await services.transactions.wait(txHash);
+
+  return {
+    _kind: "success",
+    result: {
+      hash: txHash,
+    },
+  };
+}
--- a/packages/core/src/types/future.ts
+++ b/packages/core/src/types/future.ts
@@ -62,6 +62,14 @@ export interface EventFuture {
   params: EventParams;
 }
 
+export interface SendFuture {
+  vertexId: number;
+  label: string;
+  type: "send";
+  subtype: "eth";
+  _future: true;
+}
+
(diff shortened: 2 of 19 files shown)


packages/core/src/execution/dispatch/executeSendETH.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    txHash = await services.contracts.sendTx(tx, options);
+  } catch (err) {
+    return {
+      _kind: "failure",
+      failure: err as any,
+    };
+  }
+
+  await services.transactions.wait(txHash);
```
kanej, line 30: I have been meaning to ask about this. The `sendTx` call is within the catch by the `wait` call is not. Are there any cases where `wait` can blow up?

I was thinking about this when looking over the execute call code. I didn't make a change, but wanted to come back to it.

packages/core/src/utils/guards.ts
```diff
@@ -131,3 +132,23 @@ export function isCallable(
 
   return future.type === "contract" || future.type === "library";
 }
+
+export function isContract(
+  future: DeploymentGraphFuture
+): future is ContractFuture {
+  if (isProxy(future)) {
```
kanej, line 139: Does this need to be recursive? Just saying it makes me wince.

packages/core/src/types/future.ts
```diff
@@ -70,6 +78,7 @@ export interface EventParamFuture {
   vertexId: number;
   label: string;
   type: "eventParam";
+  subtype: string;
```
kanej, line 81: Can you walk me through the different values this can be set to?

kanej (commented):
This is looking good.

packages/core/src/execution/dispatch/executeSendETH.ts
morgansliman: Looking into the ethers code, it looks like it can throw on a timeout exceeded or if the given tx hash is replaced on-chain. I just copied this code from `executeContractCall`. I can update all of these in another PR

packages/core/src/utils/guards.ts
morgansliman: Honestly, no idea, I just copied it from the `isCallable` function above it

packages/core/src/types/future.ts
morgansliman: Any solidity type except `mapping`, I believe. So, `uint8..256`, `int8..256`, `fixed8..256`, `address`, `bool`, `bytes`, and `bytes1..32`, also array versions of all of those as well as `tuple` which is how `struct`s are represented as well.

morgansliman: I think we eventually could use these to make validation for `awaitEvent` much more robust.

packages/core/src/execution/dispatch/executeSendETH.ts
kanej: Agreed. It was looking at the `executeContractCall` last week that got me wondering. I will open a linear ticket for this.

packages/core/src/utils/guards.ts
kanej: I will add a linear ticket with a potential test (passing params through nested modules so param wrapped in param).

packages/core/src/types/future.ts
kanej: Ah I see. So we use the abi to derive the subtype and then we can use that in validation - the abi says this is a uint256 but your trying to passin it as an arg to a function that takes strings?

PR #97, opened 2023-01-09
Fix for sendETH graphing

morgansliman:
fixing a bug that didn't properly map dependencies when sending ETH


morgansliman:
Note the base branch this is set to merge into is #96

kanej (approved):

PR #98, opened 2023-01-10
Fixed plan task for send and event vertexes

morgansliman:
added `value` field in vertex info page and made the whole thing a bit more stable and future-proof


morgansliman:
again, merging into #96

kanej (approved):
Looks good.

PR #103, opened 2023-01-13
Confirm deploy to non-hardhat network

### LINKED ISSUE #95: Ask for confirmation on deploy to mainnet
There should be an override flag, but if you are spending money due to an ignition deploy we should show a confirmation message.

morgansliman:
closes #95 

## Preview

Jan-13-2023 11-04-34

### DIFF
--- a/packages/hardhat-plugin/package.json
+++ b/packages/hardhat-plugin/package.json
@@ -46,6 +46,7 @@
     "@types/mocha": "^9.0.0",
     "@types/ndjson": "2.0.1",
     "@types/node": "12.20.25",
+    "@types/prompts": "^2.4.2",
     "@types/sinon": "^10.0.13",
     "@typescript-eslint/eslint-plugin": "4.31.2",
     "@typescript-eslint/parser": "4.31.2",
@@ -80,6 +81,7 @@
     "ink": "3.2.0",
     "ink-spinner": "4.0.3",
     "ndjson": "2.0.0",
+    "prompts": "^2.4.2",
     "react": "18.2.0",
     "serialize-error": "8.1.0"
   }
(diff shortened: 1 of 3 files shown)


packages/hardhat-plugin/package.json
```diff
@@ -80,6 +81,7 @@
     "ink": "3.2.0",
     "ink-spinner": "4.0.3",
     "ndjson": "2.0.0",
+    "prompts": "^2.4.2",
```
kanej, line 84: I think we should leverage the ink framework. I know it is more hassle, but better one ui approach than many.

Maybe leverage something like: https://github.com/kevva/ink-confirm-input

morgansliman: Hmm I kind of disagree in this one instance, but first I think there's a decision to be made. At what point do we want to prompt the user?

My inclination was to prompt immediately because anything else would be a waste if the user cancelled. I could also see the case for not wanting compilation to run on a cancelled deployment, not to mention clearing the terminal for a cancellation. If we go this route, then I think ink would not be the best decision because of how we encapsulated the ui renderer, although I guess we could just import and call `render` directly, but I personally think that's less clean than keeping ink in one place.

kanej: Nice move. I am forced to agree that we should short circuit, which would imply the choice is between:

1. an early prompt using the prompt library
2. rewriting the prompt in ink out of a sense of consistency
3. refactoring the existing ink rendering code to an architecture that can support this use case

Lets go with the prompt library (1), your approach. I suspect more edge cases will push us to 3, but lets wait for them to appear as there is no difference in cost to doing them now or doing them later when we have more info.

kanej (approved):

PR #107, opened 2023-01-16
fix(journal): categorise commands by chainId

### LINKED ISSUE #99: Seperate out network in journal
Currently the journal is not recording the concept of network. A run against mainnet will read the state of a previous run against testnets.

A design question is whether we want a single file for the journal or one journal file per network?

kanej:
Record the `chainId` along with the execution command, so that reruns are scoped by chain.

Fixes #99.


morgansliman (approved):
LGTM

PR #110, opened 2023-01-18
refactored awaitEvent to just event

### LINKED ISSUE #108: Rename `awaitEvent` to `event`
This is based on feedback from Anatol.

The reasoning is that `await` is loaded given the js context.

morgansliman:
closes #108


kanej (approved):
LGTM!

PR #115, opened 2023-01-18
fix: guard against reruns with changes

kanej:
We intend to add a reconciliation phase to allow for small changes in modules between runs. For the moment we will ban all changes and report an error if any are found.


morgansliman (approved):
:+1:

PR #116, opened 2023-01-19
Catch potential error when confirming tx

### LINKED ISSUE #84: Catch `transactions.await`
Both executeSendEth, executeContractCall and potentialially other execute calls have a pattern of send a transaction within a try catch, but then waiting the txHash outside of the try catch. This is likely unsafe.

Morgan took a look at the ethers docs:

> Looking into the ethers code, it looks like it can throw on a timeout exceeded or if the given tx hash is replaced on-chain.

We should add a general try catch for the wait but timeouts should lead to an on-hold state.

morgansliman:
closes #84

### DIFF
--- a/packages/core/src/execution/dispatch/executeContractCall.ts
+++ b/packages/core/src/execution/dispatch/executeContractCall.ts
@@ -34,7 +34,13 @@ export async function executeContractCall(
     };
   }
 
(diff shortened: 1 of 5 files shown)


kanej (approved):
Looks good.

Issue #102, opened 2023-01-11
Allow loading of modules by full path


kanej:
For the moment, there is no special meaning to the ignition folder, but we assume it on the cli.

~~We should remove that constraint so that modules can be loaded from anywhere.~~

We should allow the path to the module to be specified fully and fallback to looking in the ignition folder. This will allow the support of cli file completions.

PR #142, opened 2023-02-08
feat: specify module by path

kanej:
At the cli, allow the relative path to the module to be given (rather than just the name). This allows the user to leverage shell completions to the file.

The module file must be within the modules folder still.

Fixes https://github.com/NomicFoundation/ignition/issues/102.

## Preview

Feb-08-2023 15-38-55

## Details

We now support:

```
npx hardhat deploy LockModule
npx hardhat deploy LockModule.js
npx hardhat deploy LockModule.ts
npx hardhat deploy ./ignition/LockModule.js
```

The docs now default to the filename of the module without the extension (option 1 above).

The path patterns apply to plan as well.

The codepaths that supported deploying multiple modules at once have been removed.


zoeyTM (approved):
LGTM

packages/hardhat-plugin/src/load-module.ts
```diff
@@ -12,7 +12,7 @@ export function loadModule(
   debug(`Loading user modules from '${modulesDirectory}'`);
 
   if (!fsExtra.existsSync(modulesDirectory)) {
-    throw new Error(`Directory ${modulesDirectory} not found.`);
+    throw new IgnitionError(`Directory ${modulesDirectory} not found.`);
```
kanej, line 15: Someone should write a plugin to save us from this numpty. 🤦

PR #145, opened 2023-02-09
Fix/missing error msg on no passed module params

### LINKED ISSUE #141: Bad message on not passing module parameters
Feedback from @antico5:

> using getParam, didnt specify them, output message was `Ignition starting .`
    - writing a .config.json file and passing it with --params works well

We should improve our validation to determine which modules require parameters and display a useful error at the cli and within scripts/tests.

kanej:
General errors during transform were not being reported back, instead leaking out to the top level but not being shown.

A top level handler has been added to the deploy process in ignition core that moves the phase to `failed-unexpectedly`. The cli has been modified to display errors in this state.

A caveat to the general handling of unexpected errors is that any exception happening in the validation phase is treated as a validation error rather than an unexpected error.

The top level ui display component has been refactored to show a type error if it is not covering all phases.

An additional change is that the ink ui is now turned off if you have `DEBUG` set, so:

```shell
DEBUG=ignition:* npx hardhat deploy ExampleModule
```

Fixes #141.

## Preview

Feb-09-2023 15-47-00

### DIFF
--- a/packages/core/src/Ignition.ts
+++ b/packages/core/src/Ignition.ts
@@ -67,41 +67,69 @@ export class Ignition {
       this._uiRenderer
     );
 
-    const chainId = await this._services.network.getChainId();
-    await deployment.setChainId(chainId);
-    await deployment.setNetworkName(options.networkName);
+    try {
+      const chainId = await this._services.network.getChainId();
+      await deployment.setChainId(chainId);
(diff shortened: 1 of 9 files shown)


zoeyTM (approved):
This is awesome!! Love it. LGTM :+1:

PR #146, opened 2023-02-10
Fix/error on async to buildmodule

### LINKED ISSUE #138: Forbid async callbacks when creating a module
We silently fail if `buildModule` is passed an async function. Detect and warn if that happens.

Test the fix against:

* cli usage
* test usage
* script usage

kanej:
Due to typescript transpilation we can't check that the function passed into `buildModule` is async, instead we check whether the returned result is a promise or not.

If we detect a promise being returned Ignition throws and relays on the general error handler for display in the UI.

Fixes #138.

## Preview

image


zoeyTM (approved):
I'm trying to think of a case where running the action before throwing could cause side effects, but I can't think of any. LGTM!

PR #149, opened 2023-02-15
Typescript support

### LINKED ISSUE #101: Support for typescript modules
Allow projects to specify their ignition module files in typescript.

### LINKED ISSUE #140: Fix example of getOptionalParam
`getOptionalParam` doesn't accept BigNumbers, at least in the signature, but the example uses a BigNumber value for ONE_GWEI

zoeyTM:
- fully support using Ignition with typescript projects
- properly overloading `m.contract(...)` and `m.library(...)` for automatic return value typing
- refactored internal serialization to be more straightforward and type-friendly
- refactored module return types to ensure proper typing in typescript tests
- enforce only allowing contracts (and proxies/virtuals) to be returned from a module 
- added a typescript example project 

Closes #101 
Closes #140

### DIFF
--- a/examples/create2/test/create2.test.js
+++ b/examples/create2/test/create2.test.js
@@ -12,20 +12,4 @@ describe("Create2", function () {
   it("should return an instantiated factory", async function () {
     assert.isDefined(deployResults.create2);
   });
-
-  it("should have deployed the foo contract", async () => {
-    const address = deployResults.fooEvent.deployed;
-    const FooFactory = await hre.ethers.getContractFactory("Foo");
-    const foo = FooFactory.attach(address);
-
-    assert.equal(await foo.name(), "Foo");
-  });
-
-  it("should have deployed the bar contract", async () => {
-    const address = deployResults.barEvent.deployed;
-    const BarFactory = await hre.ethers.getContractFactory("Bar");
-    const bar = BarFactory.attach(address);
-
-    assert.equal(await bar.name(), "Bar");
-  });
 });
--- a/examples/sample/ignition/LockModule.js
+++ b/examples/sample/ignition/LockModule.js
@@ -5,7 +5,7 @@ const currentTimestampInSeconds = Math.round(new Date(2023, 01, 01) / 1000);
 const TEN_YEAR_IN_SECS = 10 * 365 * 24 * 60 * 60;
 const TEN_YEARS_IN_FUTURE = currentTimestampInSeconds + TEN_YEAR_IN_SECS;
 
-const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei");
+const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei").toString();
 
 module.exports = buildModule("LockModule", (m) => {
   const unlockTime = m.getOptionalParam("unlockTime", TEN_YEARS_IN_FUTURE);
--- a/examples/ts-sample/ignition/LockModule.ts
+++ b/examples/ts-sample/ignition/LockModule.ts
@@ -20,11 +20,7 @@ const LockModule = buildModule("LockModule", (m) => {
     value: lockedAmount,
   });
 
-  const c = m.call(lock, "test", { args: [] });
-
-  const event = m.event(lock, "Withdrawal", { args: [], after: [c] });
-
-  return { lock, event };
+  return { lock };
 });
 
 export default LockModule;
--- a/examples/ts-sample/ignition/LockModule.ts
+++ b/examples/ts-sample/ignition/LockModule.ts
@@ -1,0 +1,28 @@
+import hre from "hardhat";
+import { buildModule } from "@ignored/hardhat-ignition";
+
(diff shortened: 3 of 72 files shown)


alcuadrado (commented):
I left some minor comments. Can you explain this further?

> refactored internal serialization to be more straightforward and type-friendly

packages/tsconfig.settings.json
```diff
@@ -11,6 +11,7 @@
     "esModuleInterop": true,
     "moduleResolution": "node",
     "declaration": true,
+    "declarationMap": true,
```
alcuadrado, line 14: I don't think PRs should modify tooling config without an explanation. I started a conversation on slack to normalize our settings across projects.

Having said that, this change is correct, this should have always been `true`.

examples/sample/ignition/LockModule.js
```diff
@@ -5,7 +5,7 @@ const currentTimestampInSeconds = Math.round(new Date(2023, 01, 01) / 1000);
 const TEN_YEAR_IN_SECS = 10 * 365 * 24 * 60 * 60;
 const TEN_YEARS_IN_FUTURE = currentTimestampInSeconds + TEN_YEAR_IN_SECS;
 
-const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei");
+const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei").toString();
```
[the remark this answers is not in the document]
zoeyTM: Intentional. parameters don't support BigNumbers as input, so we should've been converting this to a string the whole time (#140)

examples/ts-sample/ignition/LockModule.ts
```diff
@@ -0,0 +1,26 @@
...
+const currentTimestampInSeconds = Math.round(
+  new Date(2023, 1, 1).valueOf() / 1000
+);
+const TEN_YEAR_IN_SECS = 10 * 365 * 24 * 60 * 60;
+const TEN_YEARS_IN_FUTURE = currentTimestampInSeconds + TEN_YEAR_IN_SECS;
+
+const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei").toString();
+
+const LockModule = buildModule("LockModule", (m) => {
+  const artifact = require("../artifacts/contracts/Lock.sol/Lock.json");
```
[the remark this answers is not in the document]
zoeyTM: We have a to-do for adding a utility function, similar to `m.getBytesForArtifact` that would return the entire artifact instead so we wouldn't need to do this. At one point, this module had an event being awaited in it, which requires an artifact contract, but i removed the event and forgot to remove this artifact. I'll make that fix

packages/tsconfig.settings.json
zoeyTM: Agreed. I think I was looking at the hardhat repo actually when i fixed this.

zoeyTM:
> I left some minor comments. Can you explain this further?
> 
> > refactored internal serialization to be more straightforward and type-friendly

Right, so we were serializing for a lot of other return types than what we support, and it was making the types a nightmare internally. I refactored the serialization and streamlined it, removing unnecessary return types and code paths.

examples/sample/ignition/LockModule.js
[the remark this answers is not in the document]
alcuadrado: got it! thanks.

I think we should come up with a way of supporting typed params, but that'd be a separate issue.

examples/ts-sample/ignition/LockModule.ts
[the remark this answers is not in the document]
alcuadrado: thanks for the explanation!

packages/core/test/deploymentBuilder/useModule.ts
```diff
@@ -491,6 +491,7 @@ describe("deployment builder - useModule", () => {
   });
 
   describe("returning non contract/library futures from within a module", () => {
+    // @ts-ignore
```
kanej, line 494: Victory!

examples/create2/test/create2.test.js
```diff
@@ -12,20 +12,4 @@ describe("Create2", function () {
   it("should return an instantiated factory", async function () {
     assert.isDefined(deployResults.create2);
   });
-
-  it("should have deployed the foo contract", async () => {
-    const address = deployResults.fooEvent.deployed;
```
kanej, line 17: Am I following right?

We need to remove these tests because we no longer return events, which would allow us to get the deployed contracts address. We intend to implement a way of converting a returned event param to a ContractFuture within a module, but that doesn't exist yet.

kanej (approved):
Looks good.

I have pushed commits to:

* allow returning of a library future from a module
* making the visit result generic, which allows the validation functions to return `undefined` without an `any` cast

examples/create2/test/create2.test.js
zoeyTM: No, it does exist and that's what we are returning instead, but the rest is correct

PR #151, opened 2023-02-21
feat: force flag on deploy

### LINKED ISSUE #132: Include `--force` flag in deploy to reset the journal for the current network
For localhost in particular, it should be possible to restart the deployment as if no previous runs had occurred.

In practice this should likely take the form of an override command that clears the execution state rather than deleting old messages, so that the transaction log remains append only.

kanej:
Support `--force` on deploy to ignore previous deployment runs and start again as if from a clean deployment. The history in the journal is kept but each action in the execution graph will be executed.

Resolves #132.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -22,7 +22,7 @@
     {
       "name": "Debug core tests",
       "internalConsoleOptions": "openOnSessionStart",
-      "program": "${workspaceFolder}/packages/core/node_modules/mocha/bin/_mocha",
+      "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
       "request": "launch",
       "skipFiles": ["<node_internals>/**"],
       "type": "node",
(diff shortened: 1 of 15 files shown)


zoeyTM:
Worth noting that this will overwrite the existing journal with each new (forced) deployment. I think this is intended, but I wanted to mention it regardless.

zoeyTM (approved):
LGTM

kanej:
The journal shouldn't get rewritten. The start execution message now has a `force` flag, so if you ran a deployment and it failed and you rerun with force, the journal will show the history for the first run and the history for the second, but the second's `execution::start` message will have `force: true` set against it.

I initially thought we should clear the journal, but I suspect it is better to keep the history but append a message saying, "START DEPLOYMENT OVER". We can't show that history right now, but better to have a full log file, than want the previous logs and not be able to get it.

Issue #87, opened 2023-01-04
Improve validation errors


kanej:
If we throw due a validation fail, we want the stack trace to lead to the failing line.

This means merging the validation code from the validation phase into the deployment builder.

We need to consider how this affects return results in tests. Presumably we should change to just throw the exception if there is a validation failure, rather than returning a failed results?

PR #160, opened 2023-03-03
improvement: validation errors

kanej:
Show the file/line/column against each validation error. The display allows cmd+click in the vscode terminal.

## Preview

image

## Implementation Details

To support this we capture a potential validation error when the deployment builder action is called. These potential errors are passed to the validation phase, one for each vertex. If a validation check fails we use the pre-captured error so that the stack trace shows the correct point in the module.

We have to do some magic in IgnitionValidationError. Firstly we need to externally rewrite the stack via a function call `resetStackFrom`, this is because Hardhat Errors manipulate the  stack themselves. To allow calling a custom function on the error, we need to manipulate the prototype because ... javascript.

Resolves https://github.com/NomicFoundation/ignition/issues/87.

### DIFF
--- a/packages/core/src/dsl/DeploymentBuilder.ts
+++ b/packages/core/src/dsl/DeploymentBuilder.ts
@@ -58,6 +58,15 @@ import { resolveProxyDependency } from "utils/proxy";
 import { DeploymentGraph } from "./DeploymentGraph";
 import { ScopeStack } from "./ScopeStack";
 
+type DeploymentApiPublicFunctions =
+  | InstanceType<typeof DeploymentBuilder>["contract"]
+  | InstanceType<typeof DeploymentBuilder>["library"]
+  | InstanceType<typeof DeploymentBuilder>["contractAt"]
+  | InstanceType<typeof DeploymentBuilder>["call"]
+  | InstanceType<typeof DeploymentBuilder>["event"]
+  | InstanceType<typeof DeploymentBuilder>["sendETH"]
+  | InstanceType<typeof DeploymentBuilder>["useModule"];
+
 const DEFAULT_VALUE = ethers.utils.parseUnits("0");
 
 function parseEventParams(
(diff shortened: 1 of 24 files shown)


packages/core/src/validation/dispatch/helpers.ts
```diff
@@ -68,9 +89,33 @@ export async function validateBytesForArtifact(
...
+export function buildValidationError(
+  vertex:
+    | ArtifactContractDeploymentVertex
+    | HardhatContractDeploymentVertex
+    | EventVertex
+    | CallDeploymentVertex
+    | ArtifactLibraryDeploymentVertex
+    | HardhatLibraryDeploymentVertex
+    | DeployedContractDeploymentVertex
+    | SendVertex,
```
zoeyTM, line 108: Can we define a `ValidationVertex` or similar that is this union and use that here instead

packages/core/src/dsl/DeploymentBuilder.ts
```diff
@@ -567,10 +573,26 @@ export class DeploymentBuilder implements IDeploymentBuilder {
...
+  ) {
+    const potentialValidationError = new IgnitionValidationError("");
+    potentialValidationError.resetStackFrom(f);
+    callPoints[vertexId] = potentialValidationError;
+  }
+
   private static _addVertex(
     graph: DeploymentGraph,
+    callPoints: CallPoints,
+    // eslint-disable-next-line @typescript-eslint/ban-types
```
zoeyTM, line 590: What is the error being suppressed here? Is it because `f` is typed as a generic `Function` and not something more descriptive?

zoeyTM (changes requested):

packages/core/src/dsl/DeploymentBuilder.ts
zoeyTM: will `f: () => any` fix the error?

packages/core/src/validation/dispatch/helpers.ts
kanej: Good point. I have switched out both lists. One is really just a any deployment graph vertex, the other is that with a constraint that it has args as well.

packages/core/src/dsl/DeploymentBuilder.ts
kanej: That doesn't quite work due to ts generics. I reworked to be explicit about the public functions should be passed to add vertex and then elided the type to any to pass it to the underlying function.

Take a look at the commit, or if you think the original approach is simpler, let me know.

packages/core/src/dsl/DeploymentBuilder.ts
```diff
@@ -58,6 +58,15 @@ import { resolveProxyDependency } from "utils/proxy";
...
 import { ScopeStack } from "./ScopeStack";
 
+type DeploymentApiPublicFunctions =
+  | InstanceType<typeof DeploymentBuilder>["contract"]
+  | InstanceType<typeof DeploymentBuilder>["library"]
+  | InstanceType<typeof DeploymentBuilder>["contractAt"]
+  | InstanceType<typeof DeploymentBuilder>["call"]
+  | InstanceType<typeof DeploymentBuilder>["event"]
+  | InstanceType<typeof DeploymentBuilder>["sendETH"]
+  | InstanceType<typeof DeploymentBuilder>["useModule"];
```
zoeyTM, line 68: Ooh that's a handy helper. I like that

zoeyTM (approved):
LGTM

PR #165, opened 2023-03-08
refactor(core): remove dep on hardhat

### LINKED ISSUE #163: refactor: remove hh as dep from core
Ignition core should be usable without any reference to HH. This task is to remove the dependency. HH can be used from plugin. This will mean reworking things like error handling so that core throws pure ignition errors, and plugin catches and rethrows HH exceptions.

kanej:
Instead of throwing hh errors, we throw our own ignition core errors, this allows us to remove hardhat as a dependency. This is ideal as ignition core should not force you to use hardhat even implicitly.

Resolves #163.

### DIFF
--- a/packages/core/.eslintrc.js
+++ b/packages/core/.eslintrc.js
@@ -8,7 +8,6 @@ module.exports = {
     createDefaultProgram: true,
   },
   rules: {
-    "@nomiclabs/hardhat-internal-rules/only-hardhat-plugin-error": "error",
     "no-console": "error",
   },
   ignorePatterns: ["post-build.js"],
--- a/config/eslint/.eslintrc.js
+++ b/config/eslint/.eslintrc.js
@@ -20,12 +20,7 @@ module.exports = {
     project: "./tsconfig.json",
     tsConfigRootDir: "./",
   },
-  plugins: [
-    "@nomiclabs/eslint-plugin-hardhat-internal-rules",
-    "eslint-plugin-import",
-    "@typescript-eslint",
-    "mocha",
-  ],
+  plugins: ["eslint-plugin-import", "@typescript-eslint", "mocha"],
   rules: {
     "import/no-unused-modules": [
       1,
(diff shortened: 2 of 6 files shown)


kanej:
@alcuadrado this removes HH as a dep in core. As a warning it is based off of the validation improvements PR.

I haven't wrapped and rethrown the core error as a hardhat plugin error as we take control of the cli ui with ink and control the display of errors. Other than display are there other reasons to throw a HH error?

zoeyTM:
> @alcuadrado this removes HH as a dep in core. As a warning it is based off of the validation improvements PR.
> 
> I haven't wrapped and rethrown the core error as a hardhat plugin error as we take control of the cli ui with ink and control the display of errors. Other than display are there other reasons to throw a HH error?

display was the big one

packages/core/.eslintrc.js
```diff
@@ -8,7 +8,6 @@ module.exports = {
     createDefaultProgram: true,
   },
   rules: {
-    "@nomiclabs/hardhat-internal-rules/only-hardhat-plugin-error": "error",
```
zoeyTM, line 11: I understand why we need to remove hh as a dependency and agree, but it's unfortunate that we're losing this linting to enforce ignition errors. I would love to find a solution that would solve both problems, but aside from writing our own eslint plugin to accomplish the same thing, I'm not sure what that solution would actually be.

zoeyTM (approved):

packages/core/.eslintrc.js
alcuadrado: Maybe parametrize the rule? `only-custom-error` or something like that?

alcuadrado:
Looks good.

I think there are still many things named `Hardhat` in `core`'s codebase that could have a more generic name.

kanej:
> Looks good.
> 
> I think there are still many things named `Hardhat` in `core`'s codebase that could have a more generic name.

Agreed, but I think that Hardhat is part of Ignition, in the sense that we have integration to load artifacts based on HH.

I think the answer to this is that we separate out more clearly the module api that has hh links and the execution engine that has no deps on hh at all. Technically we could separate the module api and the execution engine into different packages.

PR #166, opened 2023-03-08
Adding support for accounts from configured networks

### LINKED ISSUE #135: Add accounts to the Module API
Currently there is no in-built way of working with different accounts in ignition. We want users of ignition to be able to:

* leverage multiple accounts in test fixtures from hh
* use multiple accounts in a production deployment

## Potential considerations

* specify an account per action (i.e. deploy, call)
* ~~passing accounts as parameters to modules~~
* ui display of accounts in use
* validation?
* compatibility with hh's approach to hardware wallets

zoeyTM:
- [x] add `m.accounts: string[]` to the deployment builder for easy access inside a module (no more hardcoding required!)
- [x] updated examples to remove hardcoded accounts in favor of `m.accounts`
- [x] add a `from` option to relevant deployment actions to allow specifying which account to send the transaction from
- [x] added docs and tests for the new features
- [x] (also fixed the `--force` flag)

fixes #135


packages/hardhat-plugin/src/buildIgnitionProvidersFrom.ts
```diff
@@ -49,6 +49,20 @@ export function buildIgnitionProvidersFrom(hre: HardhatRuntimeEnvironment) {
...
     },
     config: new ConfigWrapper(),
+    accounts: {
+      getAccounts: async () => {
+        try {
+          return (await hre.network.provider.request({
+            method: "eth_accounts",
+          })) as string[];
+        } catch {
+          return [];
```
kanej, line 53: I would be tempted not to catch this. I suspect if we can't get accounts we should stop, whether in tests or a full deploy.

packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
```diff
@@ -97,13 +99,17 @@ async function convertHardhatContractToContractDeploy(
...
+    signer,
   };
 }
 
 async function convertArtifactContractToContractDeploy(
   vertex: ArtifactContractDeploymentVertex,
   transformContext: TransformContext
 ): Promise<ContractDeploy> {
+  const signer: ethers.Signer =
+    await transformContext.services.accounts.getSigner(vertex.from);
```
kanej, line 111: What is the thinking between doing the from -> signer conversion in the graph conversion as against doing it in the execution engine?

kanej (commented):
This is looking great. The test coverage is excellent. Well done.

packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
kanej: We discussed this out of band. The conversion has to happen and the conversion phase seems like a sensible location.

There are future questions about what should be in the execution graph, because we may be persisting it to disk as part of a deployment "artifact". That will be the point to consider whether it is signers or a pointer to a signer that is in the execution graph.

packages/hardhat-plugin/src/buildIgnitionProvidersFrom.ts
kanej: I pushed the removal of this catch as it was the last outstanding change.

kanej:
I am going to merge this PR to reduce the number of open PRs that will be affected by the tsconfig refactor PR.

PR #167, opened 2023-03-10
refactor: change imports

kanej:
Several repo wide refactors/restructuring:

* remove `baseUrl` - this allows us to remove post processing but makes imports ugly again, but brings us in line with HH
* fix some prettier coverage issues (turns out it wasn't applying to ts files, who knew)
* copy across the latest eslint config from HH (though keep the harsher import reorder)
* update examples to use hardhat toolbox
* replace `test:examples` with plain `test` but use workspaces at root to differentiate between `test` (only runs on packages) and `test:examples` (only runs on examples)
* update ts-node which fixes nyc code coverage reports (finally!)
* rework tsconfig based on HH:
  - have a tsconfig in the package root, but have it include both src and test (rather than hacks like an additional `./test/tsconfig.json`)
  - have a shared tsconfig under `./config/typescript/tsconfig.json`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,8 +108,9 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
+      "integrity": "sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==",
       "dev": true,
-      "license": "Apache-2.0",
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.1.0",
         "@jridgewell/trace-mapping": "^0.3.9"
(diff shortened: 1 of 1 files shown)


zoeyTM (approved):
LGTM once the tests are fixed

packages/core/src/execution/dispatch/utils.ts
```diff
@@ -31,7 +32,7 @@ function resolveFromContext(
 
   const entry = context.get(arg.vertexId);
 
-  if (!entry) {
+  if (entry === undefined || entry === null) {
```
alcuadrado, line 35: Not-blocking: I think in general returning both `undefined` and `null` can signal that you should be returning a special type from that function.

packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
```diff
@@ -324,5 +324,5 @@ async function resolveBytesForArtifact(
 }
 
 function assertNeverParamResult(hasParamResult: never) {
```
alcuadrado, line 326: I think there's a misunderstanding about what `never` means. The nerdy name for it is bottom type, and it means the absence of value (note: not undefined nor null, those are values), so if an argument has type `never` that signals to the compiler that that function will never be called.

packages/core/test/deploymentBuilder/value.ts
```diff
@@ -1,15 +1,20 @@
...
-import type { Artifact } from "types/hardhat";
-import { isArtifactContract, isHardhatContract, isCall } from "utils/guards";
+} from "../../src/types/deploymentGraph";
+import type { Artifact } from "../../src/types/hardhat";
+
+import { assert } from "chai";
+import { ethers } from "ethers";
+
+import { buildModule } from "../../src/dsl/buildModule";
+import { generateDeploymentGraphFrom } from "../../src/process/generateDeploymentGraphFrom";
```
alcuadrado, line 12: Noticed this in many files. Why doesn't the linter auto-orders this? There's a pretty random order here, wheres before external dependencies were first, and then the local ones.

alcuadrado:
I left a few comments, mostly unrelated to the PR itself, except for this: https://github.com/NomicFoundation/ignition/pull/167/files#r1136232827

Also, I think doing `Boolean(value)` to avoid the linter error is equivalent to disabling it and should be avoided. `Boolean(value)` has all the same pitfalls as `if (value)`.

packages/core/test/deploymentBuilder/value.ts
kanej: I think the rule has changed slightly with a newer version. It appears to sort `import type` before `import` even if that means internal before external, or more correctly I think it is:

* import type external
* import type internal
* import external
* import internal

But I assume we can change the config settings if we want?

packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
kanej: This is my default pattern for enforcing all values of a type in a switch statement, so if a new option is added to the type the compiler will error unless a new case is added.

Do you have a preferred pattern for achieving that without the never boilerplate above?

packages/core/test/deploymentBuilder/value.ts
alcuadrado: Ohhh, it must be a versioning issue! I hate when this happens 😫

I hadn't noticed the `type` in the first import. I guess it makes sense now. thanks

packages/core/src/execution/dispatch/utils.ts
kanej: Taking a look at the root cause of the entry possibly being undefined/null, the undefined comes from the `map.get`, so undefined is always a possibility. The value of the map can sometimes be null, as the internal execution state result can be `null`, the value it will have before it is populated by a dispatch.

In general I prefer using `null` to `undefined` for values that have not yet been set, but will be set, to avoid the confusion of doing a lookup with the wrong property and getting undefined and doing a lookup early in the process and the value not being set yet. But that might just be my javascript roots showing through.

The execution state that has result possibly null, is entirely internal and controlled by us, so I have no problem switching to undefined in this case to clean up the check (I have pushed a commit to that effect.

packages/core/test/deploymentBuilder/value.ts
kanej: For future reference, the config is in global eslint config:

image

You can see that type imports take priority. We can reconfigure this and `eslint:fix` will take care of the rest.

packages/core/src/execution/dispatch/utils.ts
alcuadrado: Thanks for the explanation, and sorry, as I was confused here. I didn't realize `context` was a builtin `Map`.

packages/core/src/process/transform/convertDeploymentVertexToExecutionVertex.ts
kanej: I have enabled that eslint rule and changed the code to leverage it. Two things to note:

1. It won't give you the exhaustive check if there is a default option (so I removed those)
2. You are relying on the type system, so anyone who leveraged js to sneak weird data past won't trip a coherent error

I think both of those are acceptable tradeoffs for the code simplification. The code where we had more aggressive checks is deep in core, which we should control; if we get bugs in those switch checks it suggests we need better validation at the public api. Though I reserve the right to do a complete 180 if users start reporting bugs that an explicit throw would have caught.

kanej:
I have pushed commits that remove the `Boolean(random)` pattern and @alcuadrado's suggested fix for the tests.

alcuadrado (approved):
🚀

PR #169, opened 2023-03-14
fix: support lib vertexes in plan

### LINKED ISSUE #131: plan is broken when you include a library
Including a local library in the ignition module causes an error on running `npx hardhat plan`:

```
Error: Unknown vertex type: HardhatLibrary
    at parseType (node_modules/@ignored/hardhat-ignition/src/plan/utils.ts:20:13)
    at Object.getSummaryLists (node_modules/@ignored/hardhat-ignition/src/plan/utils.ts:60:18)
    at Renderer.write (node_modules/@ignored/hardhat-ignition/src/plan/index.ts:64:32)
    at SimpleTaskDefinition.action (node_modules/@ignored/hardhat-ignition/src/index.ts:204:16)
    at async Environment._runTaskDefinition (node_modules/hardhat/src/internal/core/runtime-environment.ts:308:14)
    at async Environment.run (node_modules/hardhat/src/internal/core/runtime-environment.ts:156:14)
    at async main (node_modules/hardhat/src/internal/cli/cli.ts:272:7)
```

kanej:
Enforce the vertex types support through the type system. Include the library vertexes as a type of deploy.

Change the templating to include the new type strings.

Fixes #131.

## Preview

image

### DIFF
--- a/packages/hardhat-plugin/src/plan/utils.ts
+++ b/packages/hardhat-plugin/src/plan/utils.ts
@@ -1,30 +1,35 @@
-import {
+import type {
   VertexGraph,
   VertexDescriptor,
-  IgnitionError,
+  DeploymentGraphVertex,
 } from "@ignored/ignition-core";
 
-export function parseType(v: VertexDescriptor): string {
-  let type: string;
-  switch (v.type) {
+type DisplayType =
+  | "deploy-contract"
+  | "deploy-library"
+  | "call"
+  | "transfer"
+  | "event";
+
+export function parseType(v: VertexDescriptor): DisplayType {
+  const depVertex = v as DeploymentGraphVertex;
+  switch (depVertex.type) {
     case "HardhatContract":
     case "ArtifactContract":
-      type = "deploy";
-      break;
+    case "DeployedContract":
+      return "deploy-contract";
+    case "HardhatLibrary":
+    case "ArtifactLibrary":
+      return "deploy-library";
     case "Call":
-      type = "call";
-      break;
+      return "call";
     case "SendETH":
-      type = "transfer";
-      break;
+      return "transfer";
     case "Event":
-      type = "event";
-      break;
-    default:
-      throw new IgnitionError(`Unknown vertex type: ${v.type}`);
+      return "event";
+    case "Virtual":
+      throw new Error("Virtual vertex unexpected in plan");
   }
-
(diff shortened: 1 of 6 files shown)


packages/hardhat-plugin/src/plan/utils.ts
```diff
@@ -1,30 +1,41 @@
...
+  | "event";
+
+export function parseType(v: VertexDescriptor): DisplayType {
+  const depVertex = v as DeploymentGraphVertex;
+  switch (depVertex.type) {
     case "HardhatContract":
     case "ArtifactContract":
-      type = "deploy";
-      break;
+    case "DeployedContract":
```
zoeyTM, line 19: I'm not sure it's appropriate to handle `DeployedContract`s the same as the other two since it doesn't create a transaction or interact with the chain at all

zoeyTM: I kind of think they shouldn't be on the `plan` diagram at all

zoeyTM (changes requested):

packages/hardhat-plugin/src/plan/utils.ts
kanej: Agreed. We shouldn't be seeing the deployed contract in the execution graph as nothing is executed.

I think my ideal would be to include the deployed contract vertex in the deployment graph and validate it (that the address makes sense, maybe at some point in the future that the on chain address exists and the abi matches or some other useful check?), then eliminate it in the execution graph.

I can think of two approaches to the elimination. In transformation we simplify the deployment graph after validation to create the execution graph. The execution of deployed contract just passes the input abi/address to the result context. To replicate this we could:

1. Modify transform to allow prepopulating the result context, so pull the deployed contract logic currently in execution forward to the graph transform step
2. Change future result resolution so that if a deployed contract future is passed in, the abi/address is resolved from the data on the future rather than from the result context.

I am tempted by 2 currently, any thoughts?

kanej: I am going to separate this out into its own issue: https://github.com/NomicFoundation/ignition/issues/170

I think the PR does fix the lib display issue, it is just that it has led us to thinking about what is displayed across all vertex types.

zoeyTM (approved):

PR #172, opened 2023-03-17
clarify api

kanej:
We want to clarify what is and is not in the public api.

## TODO

* [x] remove subgraph from the api
* [x] enforce export paths through package.json
* [-] simplify types + functions passed from core to plugin
* [ ] document public functions and types


package.json
```diff
@@ -19,6 +19,7 @@
     "test:coverage": "npm run test:coverage -w packages",
     "test:examples": "npm run test -w examples --if-present",
     "watch": "tsc --build --watch packages/core packages/hardhat-plugin",
+    "typedocs": "typedoc --entryPointStrategy packages --out ./typedocs ./packages/core",
```
alcuadrado, line 22: What's this?

packages/core/package.json
```diff
@@ -18,8 +18,19 @@
...
+    ".": "./dist/src/index.js",
+    "./helpers": "./dist/src/helpers.js"
+  },
+  "typesVersions": {
+    "*": {
+      "helpers": [
+        "./dist/src/helpers"
+      ]
+    }
+  },
```
alcuadrado, line 33: What does this do?

packages/core/src/deployment/deployExecutionStateReducer.ts
```diff
@@ -79,17 +77,17 @@ function convertTo(
   vertexVisitResult: ExecutionVertexVisitResult
 ): VertexExecutionState {
   switch (vertexVisitResult._kind) {
-    case VertexResultEnum.SUCCESS:
+    case "success":
```
alcuadrado, line 80: What's the reasoning behind this change?

alcuadrado:
I think this PR is missing an important point, which is separating the public and private apis at the FS level. 

Right now it's not possible for someone reading the codebase to understand if something is public or not without reading a big chunk of the code. I know because that's my situation. Take, for example, `packages/core/src/types/serialization.ts`, are these types meant to be public or not? 


I think using the file system structure to reflect that is very handy and easy to understand. 

I believe the confusion comes from how I have been communicating the need to separate the private and public apis: mostly focused users importing private modules. I failed to communicate that that's just half of the story, and this is also great to help understand the codebase. It has the added benefit of making some architectural errors easier to understand in a code review (e.g. a new export of a private type is easy to spot in a diff).

I'd change the structure into something like:

```
packages/core/src/
  index.ts
  types/
     <public types>
  <modules that need to be plublic, like the dsl>
  internal/
    types/
         <private types>
    <everything else>
```

packages/core/package.json
kanej: It is what is allowing the reading of type information for the `@ignored/ignition-core/helpers` entry point.

It is why `@hardhat-ignition` is able to resolve the type information. I haven't checked its impact on `.js.map` and `.d.ts.map` files though (vscode was navigating but let me double check)

packages/core/package.json
```diff
@@ -18,8 +18,19 @@
...
     "node": "^14.0.0 || ^16.0.0 || ^18.0.0"
   },
-  "main": "dist/src/index.js",
-  "types": "dist/src/index.d.ts",
+  "main": "./dist/src/index.js",
+  "types": "./dist/src/index.d.ts",
+  "exports": {
+    ".": "./dist/src/index.js",
+    "./helpers": "./dist/src/helpers.js"
+  },
```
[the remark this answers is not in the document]
kanej: The type constraint is being worked around by the `typeVersions` hack.

packages/core/src/deployment/deployExecutionStateReducer.ts
kanej: I was under the impression that we prefer union types to enums, have I misunderstood that?

Is the issue that I have used the string literal instead of a `const SUCCESS = 'SUCCESS'`

package.json
kanej: That is a documentation generator that reads the typescript source. I am using it to quickly visualise the types in the package and their documentation.

I suspect it will not make it into the final version of the PR once it has done its job.

kanej:
> I think this PR is missing an important point, which is separating the public and private apis at the FS level.
> 
> Right now it's not possible for someone reading the codebase to understand if something is public or not without reading a big chunk of the code. I know because that's my situation. Take, for example, `packages/core/src/types/serialization.ts`, are these types meant to be public or not?
> 
> I think using the file system structure to reflect that is very handy and easy to understand.
> 
> I believe the confusion comes from how I have been communicating the need to separate the private and public apis: mostly focused users importing private modules. I failed to communicate that that's just half of the story, and this is also great to help understand the codebase. It has the added benefit of making some architectural errors easier to understand in a code review (e.g. a new export of a private type is easy to spot in a diff).
> 
> I'd change the structure into something like:
> 
> ```
> packages/core/src/
>   index.ts
>   types/
>      <public types>
>   <modules that need to be plublic, like the dsl>
>   internal/
>     types/
>          <private types>
>     <everything else>
> ```

Ha, I had a version of the fs separation and reverted it when I got typescript to play nice with the package.json exports. It should be quick to recreate though. I had misunderstood it as an either or; we would prefer to block people importing internal code using package.json exports but if we couldn't make that work we would fall back to a fs separation.

I am happy to make that change, it is just not what my intuition was. I would always assume that the top level package exports are the public api; if on top of that I get errors when trying to import something under '@ignored/ignition-core/dist/src/secret.js` then I would hope I would get the point. In terms of the architectural review point, is this not equivalent to paying special attention to any new/changes exports in the `index.ts` and `helpers.ts` entry points?

The structure you outlined made me pause on the distinction between Public and Private types. It hadn't occurred to me that we would distinguish between public and private types. I suspect that my javascript tendency is showing again, to treat the code as the important thing and the types as hints that have been layered on top. But is there a sense in which the private types can be missused - or is it that if we can minimise the exposed surface, we should minimise the exposed surface?

kanej:
I will hunt out the stackoverflows/blogs on the `typesVersions` I was using for Monday. I am not committed to the approach, but wanted to give it a try and see what people thought.

kanej:
This PR has been replaced by a set of PRs that cover the same changes.

PR #174, opened 2023-03-20
refactor: limit exports from npm package

kanej:
Deliniate between public/private using package.json exports on core.

Part of https://github.com/NomicFoundation/ignition/pull/172/commits


packages/hardhat-plugin/test/use-module.ts
```diff
@@ -33,10 +32,6 @@ describe("useModule", function () {
           args: [],
         });
 
-        if (!isCallable(foo)) {
```
kanej, line 36: Our changes to the useModule return types removes the need for this check.

kanej: Zoey's enhancements to support typescript included constraining the allowed returnable futures. An attempt to return a non-contract future from the module will error. This case is covered by a core test.

The upshot of these changes is we no longer need this if check to satisfy typescript, and the error case would never be hit because useModule would have thrown earlier.

PR #178, opened 2023-03-22
Add support for retrieving artifacts

### LINKED ISSUE #155: Replace artifact api calls with single destructuring `getArtifact`
We currently support `m.getBytesForArtifact("Foo")` within the api but need a similar function for getting an abi.

Instead of adding functions for each property of artifact we should add a `getArtifact` that returns an artifact future that can be destructured into bytes/abi futures or others as we need to add them.

zoeyTM:
- added `m.getArtifact`
- removed `m.getBytesForArtifact` 

fixes #155


docs/creating-modules-for-deployment.md
```diff
@@ -95,31 +96,41 @@ const b = m.contract("B", {
...
 m.call(uniswap, "swap", { ... })
 ```
 
 ### Deploying from an artifact
 
-To allow you to use your own mechanism for getting the contract artifact, `contract` supports passing an `Artifact` as an optional second parameter:
+Additionally, `contract` supports passing an `Artifact` as an optional second parameter:
 
 ```javascript
-const artifact = hre.artifacts.readArtifactSync("Foo");
```
kanej, line 114: I think we should keep the `hre.artifacts.readArtifactSync("Foo")` as the example; to show it can be done and also because you should just use `m.contract("Foo")` in this case and our docs shouldn't show bad practice, even if it demonstrates the point.

docs/creating-modules-for-deployment.md
```diff
@@ -244,7 +257,8 @@ Because `totalSupply` is not a number, it is a future.
 A deployment can be put `on-hold` until an on-chain event has been emitted (for instance a timelock or multisig approval):
 
 ```tsx
-const multisig = m.deploy("Multisig");
+const artifact = m.getArtifact("Multisig");
+const multisig = m.contract("Multisig", artifact, { args: [] });
```
kanej, line 261: By the same argument this should probably just be `m.contract("Multisig")`.

examples/create2/ignition/Create2FactoryModule.js
```diff
@@ -1,14 +1,14 @@
 const { buildModule } = require("@ignored/hardhat-ignition");
 
-const Create2Artifact = require("../artifacts/contracts/Create2Factory.sol/Create2Factory.json");
-const BarArtifact = require("../artifacts/contracts/Bar.sol/Bar.json");
-const FooArtifact = require("../artifacts/contracts/Foo.sol/Foo.json");
-
 module.exports = buildModule("Create2Factory", (m) => {
+  const Create2Artifact = m.getArtifact("Create2Factory");
```
kanej, line 4: This is nice.

kanej (approved):
I like the approach, reducing the amount of complexity around future management by ... avoiding it, is great.

I left two comments against the docs, but this is good to merge after that.

docs/creating-modules-for-deployment.md
zoeyTM: I actually disagree on this one. Multisig is _the_ example where it does matter that we pass in an artifact because `m.event` requires it to be an `ArtifactContract`. I was just lazy on the initial docs for this because it was pseudocode anyways and the full example does show it.

kanej: Ah and that is because we need to lessen the `m.event` constraint to also include `HardhatContract`s?

zoeyTM: Originally it needed to be this way because it used the artifact to generate the event param futures. Now that I think about it, though, since we're front-loading the artifacts now, we might not need this constraint anymore

kanej: Do you want to tackle that here or move that out to a different issue?

zoeyTM: Created #187 to address this concern

Issue #171, opened 2023-03-16
Emphasis public/private api distinction


kanej:
We want to clarify the public api, which consists of the module api and the public ignition object.

## TODO

- [x] https://github.com/NomicFoundation/ignition/pull/173
- [x] https://github.com/NomicFoundation/ignition/pull/174
- [x] https://github.com/NomicFoundation/ignition/pull/175
- [x] https://github.com/NomicFoundation/ignition/pull/176
- [x] https://github.com/NomicFoundation/ignition/pull/179
- [x] https://github.com/NomicFoundation/ignition/pull/180
- [x] https://github.com/NomicFoundation/ignition/pull/181
- [x] https://github.com/NomicFoundation/ignition/pull/184
- [x] https://github.com/NomicFoundation/ignition/pull/182
- [x] Simplify API and tsdocs usage with `unknown` types
- [ ] https://github.com/NomicFoundation/ignition/issues/192
- [ ] ~Look again at deploy options - check different constructors make sense and all options are needed~
- [ ] ~Clarify types and code exported to support implementing journal (clean up old journal types)~

kanej:
This part of the refactoring is complete.

PR #181, opened 2023-03-22
refactor: pass providers to Ignition constructor

kanej:
In place of services, pass providers then immediately convert them. This has the advantage of only exporting one set of concepts (i.e. low level providers) rather than two (providers and higher level services), particularly as one leverages the other.

The consequence of this approach though is that the providers are too low level for clean mocking in our core tests. The services are the more appropriate seam. By switching to providers in the `Ignition` constructor we lose that injection seam. This has been worked around with a utility function that sets up Ignition but overrides the services so that mocked services can continue to be used in tests.

Part of #171.

### DIFF
--- a/packages/core/src/Ignition.ts
+++ b/packages/core/src/Ignition.ts
@@ -38,18 +38,34 @@ export class Ignition {
   private _uiRenderer: UpdateUiAction;
   private _journal: ICommandJournal;
 
-  constructor({
+  public static create({
     providers,
-    uiRenderer,
-    journal,
+    uiRenderer = () => {},
+    journal = new NoopCommandJournal(),
   }: {
     providers: Providers;
     uiRenderer?: UpdateUiAction;
     journal?: ICommandJournal;
   }) {
-    this._services = createServices(providers);
-    this._uiRenderer = uiRenderer ?? (() => {});
-    this._journal = journal ?? new NoopCommandJournal();
+    return new Ignition({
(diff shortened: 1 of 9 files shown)


kanej:
I am not fully committed to this approach yet. If I could have reworked the core tests to work purely in terms of providers I would have, but the mocking started exploding in complexity - the interaction between mocked code and ethers was deeply painful.

At the same time, stating "here is what you need to implement to provide your own ignition, or if you have hre hanging around here is a helper" seems like the right approach. Or maybe we should consider services our adapter for ethers + hh network?

packages/core/src/Ignition.ts
```diff
@@ -37,15 +39,15 @@ export class Ignition {
   private _journal: ICommandJournal;
 
   constructor({
```
alcuadrado, line 41: A better solution would be:

1) Keep receiving services here.
2) Make the constructor `protected`.
3) Add a `static` factory that receives providers.
4) In the tests extend `Ignition` (e.g. `MockableIgnition`) with a public `constructor` or another factory.

alcuadrado: This saves us from fiddling with private fields later

kanej: I have added a static factory method and made the Ignition constructor protected.

The tests use a `TestIgnition` with its own Services orientated constructor.

PR #182, opened 2023-03-23
refactor: update enum usage

kanej:
We have opted to leverage standard enums across the code base, this PR switches out several union types for enums, it also changes the process steps to use one unified result type rather than a result type per step.

Part of #171.

## Archived

This attempts to get a new tradeoff in our single use of an enum. We used `VertexResultEnum` as a discriminator on messages streamed to the journal. The definition of a journal is going to be external to `ignition-core`, so the enum has to be exported.

Using an enum in the consuming library is an overhead for the outside developer.

A secondary issue is that `import type {...}` breaks on enums as they are both typesystem and runtime entities.

An ideal setup would be:

1. Use of enum like constants in core
2. Enforcement of enum like constants within core by ts
3. Do not require third parties to have access to the enum to use ignition core
4. Allow for `import type {...}` freely within the `./**/types/` folder

Our current enum loses us 3 & 4. We could move the enum out to a constants file to resolve 4. But we continue to need third parties to use our exported enum to correctly type commands.

A second approach - that was tried and dropped - was to use `const enum` which does not do the generation of runtime objects. This fixes 4, but retains 3, moreover it adds a further issue, in that it breaks transpilation which works at a file level not across the entire typesystem of all relevant modules.

The third approach was to move the enum to a `./constants` file that can be accessed as an entry point by third parties. The enum is replaced witha a const (readonly) object. The command messages are typed with the keys of the object i.e. `_kind: typeof VertexResult.SUCCESS`. This approach allows consumers to just use the string "success" and have it type correctly. So we get 1, 3 and 4 but we lose 2 - you can use the const object to clean up your code but ts isn't going to say anything if you don't.

Part of #171.


kanej:
My suspicion is that the best approach is to continue to use standard enums but move them to a `./constants` or `./internal/constants` file. This would fix the import issue, but we retain the need to export the enum; but maybe that is just a pain that should be accepted, let third parties use js if they want to dodge the typesystem.

This PR shows an alternative, but I am unsure if the gains justify the non-standardness.

kanej:
I have pushed a new version of this given our decision to leverage standard enums more liberally, and accept that we will need to export the enum and expect our consumers to use them.

alcuadrado:
Merging so I can base my PR off of this

PR #183, opened 2023-03-23
Hardhat task for listing ignition deployments

### LINKED ISSUE #111: Command to list previous deployments against a module
We do currently do show the deployed contracts if a successful module deploy is rerun. However, a user may be understandably reluctant to run the `deploy` command when only information is being sought. We should add an `ignition-info` command that displays the success/unsuccessful deployed contracts, their status and address.

We would look to leverage hh scopes to get subtasks when they become available and improve the command to `npx hardhat ignition info`.

The display should be scoped if a network is provided, but show each network one after another and the contracts deployed to each.

Maybe something like `npx hardhat ignition-info --network polygon`:

```shell

polygon
-------

Contract A   deployed   0x0000123
Contract B   errored
Contract C   pending

```

and `npx hardhat ignition-info`:

```shell
mainnet
-------
Contract A   deployed   0x0000123
Contract B   deployed   0x0000124
Contract C   deployed   0x0000125


polygon
-------

Contract A   deployed   0x0000123
Contract B   errored
Contract C   pending
```

We should investigate table layouts in ink.

zoeyTM:
`hardhat ignition-info` task will display info for all saved Ignition deployments

fixes #111 

Screen Shot 2023-03-23 at 3 33 57 PM


packages/hardhat-plugin/src/index.ts
```diff
@@ -206,6 +212,64 @@ task("plan")
...
+        console.warn("No Ignition modules found");
+        process.exit(0);
+      }
+
+      const journalPath = resolveJournalPath(
+        userModule?.name,
+        hre.config.paths.ignition
+      );
+
+      const deployments = await hre.ignition.info(userModule.name, journalPath);
```
kanej, line 235: Can we narrow this api between core and the plugin.

If the code here is moved down into core so that only the final calculated network to deployed contract names/addresses mapping is passed back.

packages/core/src/Ignition.ts
```diff
@@ -191,7 +193,48 @@ export class Ignition {
...
+      [k: string]: Array<DeployStateExecutionCommand & { chainId: number }>;
+    } = {};
+    for await (const command of this._journal.readAll()) {
+      const network = networkNameByChainId[command.chainId] ?? "unknown";
+
+      if (journalData[network] === undefined) {
+        journalData[network] = [];
+      }
+
+      journalData[network].push(command);
```
kanej, line 216: Can we restructure the journalData to be id'ed by chainId. I realise this makes the value need to be a more complex object than an array. It has an advantage though in that two different `chainId`s that are not in the network map, will be kept separate but labelled 'unknown' rather than both being merged into 'unknown'.

kanej (changes requested):
This is looking good.

Unfortunately the instruction to simplify the core api task has come in before it was merged!

We should add a section to the readme to document the new info task.

packages/core/src/Ignition.ts
zoeyTM: oh good catch

zoeyTM:
Example output for multiple unknown network deployments:

Screen Shot 2023-04-04 at 1 49 54 PM

packages/core/src/internal/types/journal.ts
```diff
@@ -49,4 +49,9 @@ export interface ICommandJournal {
    * Read out the stored deployment update commands.
    */
   read(): AsyncGenerator<DeployStateExecutionCommand, void, unknown>;
+  readAll(): AsyncGenerator<
+    DeployStateExecutionCommand & { chainId: number },
```
kanej, line 53: This doesn't need changed, but I suspect it means we should rework the structure of the journal (probably multiple files - one per chain).

packages/core/src/Ignition.ts
```diff
@@ -227,7 +234,89 @@ export class Ignition {
...
+      const deployment = new Deployment(
+        moduleName,
+        this._services,
+        this._journal
+      );
+
+      const [networkName, chainId] = networkTag.split(":");
+      await deployment.setDeploymentDetails({
+        networkName,
+        chainId: +chainId,
```
kanej, line 274: In the same situation, I would do a `parseInt`. This is the sort of thing that should be an eslint rule, but I am will ping the typescript channel and see what the collective wisdom is.

kanej: Alright, the channel has spoken (Franco) and it is neither, `Number(chainId)` wins!

kanej (approved):
One minor casting tweak, but after that this is good to merge!

packages/core/src/Ignition.ts
kanej: The channel speaks again. `parseInt` for anything that might come from a user, internally we don't care. Given this is entirely internal, I think it can be left as it is.

kanej (approved):

PR #184, opened 2023-03-24
refactor: add tsdocs for public api

kanej:
We are adding tsdocs to the methods and types of the public api to help with discoverability by intellisense.

## Preview

image

## Details

To support this goal we are adding two tools:

* typedocs - to generate html documentation from the tsdocs (currently only for internal use)
* api-extractor - to enforce linting rules on tsdocs like making sure every type used in the public api is exported.

We are also exporting (and documenting) all types that users could encounter through our api. This rule is now enforced in linting via `api-extractor`.

All exported class/function/types are marked with either `@alpha` or `@internal`. Any exported typed that an end Ignition user can use, is marked with `@alpha`. Any exported typed meant only to be used by tool developers (so us) is marked with `@internal`.

Currently that breaks down as:

image

A similar distinction is now present at the file-system level with an explicit `./internal` folder make it easy to see violations during Ignition development.

### Typedocs

To see the documentation for the exported classes/functions/types, run from root:

```sh
npm run typedocs
```

This will auto-open your browser with the generated documentation.

### Api-extractor

The `api-extractor` is only enabled for core. Within core you can run it with:

```sh
npm run api-extractor
```

It runs as part of core's linting.

A task has been added to `./.vscode`, called `api-extractor` that will run npm script but also leverage a problem matcher to highlight any issues inline within vscode.

Part of #171

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "test:examples": "npm run test -w examples --if-present",
     "watch": "tsc --build --watch packages/core packages/hardhat-plugin",
     "fullcheck": "npm run build && npm run lint && npm run test:coverage && npm run test:examples",
+    "typedocs": "typedoc --entryPointStrategy packages --out ./typedocs ./packages/core && open-cli ./typedocs/index.html",
     "clean": "npm run clean --workspaces --if-present"
   },
   "devDependencies": {
@@ -31,7 +32,9 @@
     "eslint-plugin-import": "2.24.2",
     "eslint-plugin-mocha": "^9.0.0",
     "eslint-plugin-prettier": "4.0.0",
+    "open-cli": "7.2.0",
     "prettier": "2.4.1",
+    "typedoc": "0.23.28",
     "typescript": "^4.7.4"
   }
 }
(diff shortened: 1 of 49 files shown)


package.json
```diff
@@ -20,9 +20,11 @@
     "test:examples": "npm run test -w examples --if-present",
     "watch": "tsc --build --watch packages/core packages/hardhat-plugin",
     "fullcheck": "npm run build && npm run lint && npm run test:coverage && npm run test:examples",
+    "typedocs": "typedoc --entryPointStrategy packages --out ./typedocs ./packages/core && open-cli ./typedocs/index.html",
```
alcuadrado, line 23: If I run `npm run typedocs` in the root I get:

```
% npm run typedocs

> typedocs
> typedoc --entryPointStrategy packages --out ./typedocs ./packages/core && open-cli ./typedocs/index.html

warning Could not determine the entry point for "/Users/pato/projects/nomic/ignition/packages/core/package.json". Package will be ignored.
error Unable to find any entry points. Make sure TypeDoc can find your tsconfig
```

PR #189, opened 2023-03-30
fix: support arrays as arg option on call

### LINKED ISSUE #186: Support arrays as args in `call`
Currently the `m.call` action does not support arrays as args, which needs fixed:

```solidity
const someData = [1, 2, 3];

m.call(Contract, “func”, { args: [someData] });
```

kanej:
Support arrays of bools, numbers and strings as args in `m.call`.

Fixes #186.


zoeyTM (approved):
LGTM

alcuadrado (approved):

PR #196, opened 2023-04-07
Minimize the exposed API

alcuadrado:
This plugin minimizes the public API of core, while introducing a temporary bypass for the plugin.

The majority of the refactor is contained in the first commit.

PR #197, opened 2023-04-11
Implementing static calls

### LINKED ISSUE #85: Piping result of a static call into arg/values of other deployment functions
Implement static calls in the user facing api.

### Static calls

A contract might need the result of some other contract method as an input. Ideally, this would be done by the contract itself, receiving the address of the other contract instead of the result of calling it. But if that's not the case, or if for any other reason you need to make a static call to a contract, you can do it like this:

```tsx
const token = m.contract("Token")
const totalSupply = m.call(token, "totalSupply")

const someContract = m.contract("ContractName", {
  args: [totalSupply]
})
```

Notice that the same `m.call` method is used both for sending transactions and for static calls. We use a convention similar to what ethers.js does, where the behavior is different for view and non-view functions. If the user wants to statically call a non-view function, they can do it with some flag in the options.

In this example, `totalSupply` is called a **deferred value**. Similar to how a contract binding is a contract that will eventually be deployed, a deferred value is some value that will eventually be available. That means you can't do this:

```tsx
if (totalSupply > 0) {
  ...
}
```

Because `totalSupply` is not a number.

Deferred values are **resolved each time they are used**. That is:

```tsx
const totalSupply = m.call(token, "totalSupply")

const foo = m.contract("Foo", {
  args: [totalSupply]
})
const bar = m.contract("Bar", {
  args: [totalSupply]
})
```

Here there's no guarantee that both deployments will receive the same value. If a user wants that behavior, they can do it with a custom deferred value that caches the result.

zoeyTM:
- [x] implemented `m.staticCall`
- [x] allow static call futures to be passed as arguments for other module builder functions
- [x] tests
- [x] docs

fixes #85

### DIFF
--- a/docs/creating-modules-for-deployment.md
+++ b/docs/creating-modules-for-deployment.md
@@ -17,7 +17,7 @@
 - [Calling contract methods](./creating-modules-for-deployment.md#calling-contract-methods)
   - [Transferring _Eth_ as part of a call](./creating-modules-for-deployment.md#transferring-eth-as-part-of-a-call)
   - [Transferring _Eth_ outside of a call](./creating-modules-for-deployment.md#transferring-eth-outside-of-a-call)
-  - [Using the results of a call with a deferred value (TBD)](./creating-modules-for-deployment.md#using-the-results-of-a-call-with-a-deferred-value-tbd)
+  - [Using the results of statically calling a contract method](./creating-modules-for-deployment.md#using-the-results-of-statically-calling-a-contract-method)
   - [Waiting for on-chain events](./creating-modules-for-deployment.md#waiting-for-on-chain-events)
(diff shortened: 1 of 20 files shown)


zoeyTM:
Note that this branches off #196

packages/core/test/deploymentBuilder/calls.ts
```diff
@@ -249,7 +249,7 @@ describe("deployment builder - calls", () => {
       assert.isDefined(deploymentGraph);
     });
 
-    it("should have four nodes", () => {
+    it("should have two nodes", () => {
       assert.equal(deploymentGraph.vertexes.size, 2);
```
kanej, line 253: 😄

kanej (approved):
This looks good. Before merging I just want to check that we are happy with `staticCall` as the name.

PR #200, opened 2023-04-27
allow HardhatContracts for events

### LINKED ISSUE #187: Support passing HardhatContracts into `m.event`
Currently, there is a restriction when listening to events that the contract be an ArtifactContract, but that restriction is no longer needed now that we're front-loading all the artifacts.

zoeyTM:
fixes #187

### DIFF
--- a/packages/core/src/types/future.ts
+++ b/packages/core/src/types/future.ts
@@ -216,17 +216,6 @@ export interface ProxyFuture {
 }
 
 /**
- * A future representing an artifact deployment or an already existing deployed
- * contract.
- *
- * @privateRemarks
- * TODO: is this needed?
- *
- * @alpha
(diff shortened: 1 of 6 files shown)


packages/core/src/types/future.ts
```diff
@@ -215,17 +215,6 @@ export interface ProxyFuture {
...
-/**
- * A future representing an artifact deployment or an already existing deployed
- * contract.
- *
- * @privateRemarks
- * TODO: is this needed?
- *
- * @alpha
- */
-export type ArtifactFuture = ArtifactContract | DeployedContract;
```
kanej, line 227: Well, that answered that.

kanej (approved):
LGTM!

PR #201, opened 2023-04-28
Library Refactor

### LINKED ISSUE #177: Refactor Libraries
There are currently some incorrect assumptions made about libraries that aren't harming anything, per se, but we should refactor them for clarity regardless.

- [ ] inclusion in `CallableFuture` type (you can't call functions on a library)
- [ ] unnecessary arg length validation (you can't give constructor args for a library)

The list above may not be exhaustive, those are just the things I've noticed (and can remember). I'll add to it if I find more, but this ticket should probably just look at Libraries across the code base and fix any similar issues in case I forget to add any here.

zoeyTM:
There are currently some incorrect assumptions made about libraries that aren't harming anything, per se, but we should refactor them for clarity regardless.

- [x] inclusion in CallableFuture type (you can't call functions on a library)
- [x] unnecessary arg length validation (you can't give constructor args for a library)

fixes #177

### DIFF
--- a/docs/creating-modules-for-deployment.md
+++ b/docs/creating-modules-for-deployment.md
@@ -325,7 +325,7 @@ module.exports = buildModule("`TEST` registrar", (m) => {
 
 Calls to `useModule` memoize the results object, assuming the same parameters are passed. Multiple calls to the same module with different parameters are banned.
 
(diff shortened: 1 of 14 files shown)


kanej (approved):
Looks great, nice catch on the docs update.

LGTM!

PR #208, opened 2023-05-03
Move application state from core to plugin

zoeyTM:


kanej:
We moved in another direction!

PR #212, opened 2023-05-08
support plan (New API PR 9)

kanej:
Add plan feature for the new api.

This involved:
1. Adding a new website as a package under `./packages/ui`
2. Adding serialization/deserialization for IgnitionModules
3. Update the plugin to write a plan folder under cache that interpolates the serialized version of the IgnitionModule being planned

The website is a React SPA using Vite. Everything is bundled into `index.html` so it can be opened locally without any CORs issues.

Serialization is done by flattening a module into a json document where each module/submodule is represented, but references to futures/modules are replaced by tokens. On deserialization, a first pass generates all of the relevant objects, and a second pass rewires them together.

## Preview

image

## Running locally

After a successful build a dev version of the plan website can be started with:

```sh
cd ./packages/ui
npm run dev
```

There is an example Module under `./examples/ComplexModule.js` that can be edited and used to update the website with:

```sh
npm run regenerate-deployment-example
```

Additionally, the `./examples/sample` project has been converted to use `defineModule` for testing:

```sh
cd ./examples/sample
hardhat plan ./ignition/LockModule.js
```


packages/core/src/new-api/stored-deployment-serializer.ts
```diff
@@ -0,0 +1,659 @@
...
+      Object.entries(libraries).map(([key, token]) => [
+        key,
+        partialFutureLookup[token.futureId] as ContractFuture<string>,
+      ])
+    );
+  }
+
+  /**
+   * Oh you think you can defeat me typesystem. I don't acknowledge
+   * your _readonly_.
```
zoeyTM, line 330: LOL brilliant

packages/ui/public/deployment.json
```diff
@@ -0,0 +1,238 @@
+{
```
zoeyTM, line 1: is this meant to be included or ignored?

zoeyTM (approved):
I still have reservations about doing serialization/deserialization this way as opposed to simpler JSON, but this looks good. :+1:

packages/ui/public/deployment.json
kanej: I had it there to help the development experience, but are you suggesting adding something like a "predev" script that runs `npm run regenerate-development-example` before running `npm run dev`?

packages/core/src/ui-helpers.ts
```diff
@@ -0,0 +1,4 @@
+export { StoredDeploymentSerializer } from "./new-api/stored-deployment-serializer";
+export * from "./new-api/types/module";
+export { FutureType } from "./new-api/types/module";
```
alcuadrado, line 3: Isn't this already exported in the line above?

```suggestion
```

packages/core/src/new-api/types/serialized-deployment.ts
```diff
@@ -0,0 +1,239 @@
...
+ * @beta
+ */
+export type SerializedSolidityParamsType = SerializedSolidityParamType[];
+
+/**
+ * The serialized version of the artifact.
+ *
+ * @beta
+ */
+export type SerializedArtifactType = any[];
```
alcuadrado, line 22: This can be the same type as the artifact, as they are already serializable.

alcuadrado (approved):

packages/core/src/ui-helpers.ts
kanej: Good catch. Thanks.

packages/ui/public/deployment.json
kanej: I removed the `./public/deployment.json` from git and added the generation of it from the example to a prescript for `npm run dev`.

PR #223, opened 2023-05-14
Lazy module construction (New API PR 2)

alcuadrado:
This PR modifies how `buildModule` works. Instead of creating `IgnitionModule`s directly, it creates `IgnitionModuleDefinition` objects, which are in turn used to create `IgnitionModule`s.

The reason for this is that some of the `ModuleBuilder` APIs are meant to return values that are only present during deployment, and not during `buildModule`'s execution. An example of this is the chain's id. Now we keep track of the definition, and only create the `IgnitionModule` once those values are present.

This PR introduces the necessary infrastructure to add support for those kind of concrete values, but doesn't directly add support for them.

### DIFF
--- a/packages/core/test/new-api/contract.ts
+++ b/packages/core/test/new-api/contract.ts
@@ -141,29 +141,39 @@ describe("contract", () => {
     });
 
     it("should throw if the same contract is deployed twice without differentiating ids", () => {
-      assert.throws(() => {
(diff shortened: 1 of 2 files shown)


zoeyTM (approved):
LGTM!

kanej (approved):

PR #224, opened 2023-05-16
Add accounts to new API (New API PR 7)

### LINKED ISSUE #220: Support `m.accounts` in new-api

zoeyTM:
fixes #220

### DIFF
--- a/packages/core/test/new-api/call.ts
+++ b/packages/core/test/new-api/call.ts
@@ -15,7 +15,7 @@ describe("call", () => {
       return { contract1 };
     });
 
-    const constructor = new ModuleConstructor(0);
+    const constructor = new ModuleConstructor(0, []);
     const moduleWithASingleContract = constructor.construct(
       moduleWithASingleContractDefinition
     );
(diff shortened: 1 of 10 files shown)


packages/core/test/new-api/call.ts
```diff
@@ -15,7 +15,7 @@ describe("call", () => {
       return { contract1 };
     });
 
-    const constructor = new ModuleConstructor(0);
+    const constructor = new ModuleConstructor(0, []);
```
kanej, line 18: I have no problem with duplication in tests, I think you have to justify DRY in tests due to the cognitive burden of abstractions.

However, I am wondering if we are over the line and a helper should be added to support our usage of construct here?

What do people think? @alcuadrado @zoeyTM

alcuadrado: I'm normally ok with some duplication in tests, so no strong preference here.

zoeyTM: also no strong preference from me; i'm fine either way.

PR #229, opened 2023-05-17
Add value and from to new API

### LINKED ISSUE #221: Add `value` back as an option in new-api
Several actions accept `value` as an option in the old api. Bring it back for the `new-api`.

Lets consider only supporting js BigInt, as that is what ethers/viem are moving to.

zoeyTM:
Added `value` and `from` as options to relevant futures

fixes #221

### DIFF
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -133,6 +133,7 @@ export class IgnitionModuleBuilderImplementation<
     const futureId = `${this._module.id}:${id}`;
     options.libraries ??= {};
     options.value ??= BigInt(0);
+    options.from ??= this.accounts[0];
 
     this._assertUniqueContractId(futureId);
 
@@ -142,7 +143,8 @@ export class IgnitionModuleBuilderImplementation<
       contractName,
       args,
       options.libraries,
-      options.value
+      options.value,
+      options.from
     );
 
     for (const arg of args.filter(isFuture)) {
@@ -174,6 +176,7 @@ export class IgnitionModuleBuilderImplementation<
     const futureId = `${this._module.id}:${id}`;
     options.libraries ??= {};
     options.value ??= BigInt(0);
+    options.from ??= this.accounts[0];
 
     this._assertUniqueArtifactContractId(futureId);
 
@@ -184,7 +187,8 @@ export class IgnitionModuleBuilderImplementation<
       args,
       artifact,
       options.libraries,
-      options.value
+      options.value,
+      options.from
     );
 
     this._module.futures.add(future);
@@ -213,6 +217,7 @@ export class IgnitionModuleBuilderImplementation<
     const id = options.id ?? libraryName;
     const futureId = `${this._module.id}:${id}`;
     options.libraries ??= {};
+    options.from ??= this.accounts[0];
 
     this._assertUniqueLibraryId(futureId);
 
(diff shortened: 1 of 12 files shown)


zoeyTM:
also removed some duplicated interface definitions that sneaked through my rebases

alcuadrado (changes requested):
Nice PR, @zoeyTM!

Just a few small comments:

1. I'm pretty sure that the type that we want to use is `bigint` and not `BigInt`. The first one is the primitive, the latter the boxed value. Like `number` and `Number`. 
2. Static call can also have value, so maybe you can move value and from to `FunctionCallFuture`
3. I think we can always default to `value` `0` if none is present, so we can make the `value` field in the futures that have them non-optional.
4. With `from` I'd say it would be better to leave the `undefined` as `undefined` though, so that the execution strategy can have more flexibility about what to do. We just let the user choose if they really want to, but if they don't, that's better for us.

zoeyTM:
> Static call can also have value, so maybe you can move value and from to FunctionCallFuture

Are you sure about this @alcuadrado ? I'm pretty sure sending value is considered modifying state and isn't allowed in a static call. Plus static calls don't create transactions on-chain, and sending value would create one.

zoeyTM:
> I think we can always default to value 0 if none is present, so we can make the value field in the futures that have them non-optional.

They're optional because we default the entire options to an empty object if it isn't given. It is still being defaulted to zero if it isn't given, but it should stay optional just like `after` and the rest of the options.

zoeyTM:
> With from I'd say it would be better to leave the undefined as undefined though, so that the execution strategy can have more flexibility about what to do. We just let the user choose if they really want to, but if they don't, that's better for us.

I would argue that defaulting to `accounts[0]` is pretty standard. We can still allow execution strategies to override that if they choose, but leaving `from` as `undefined` would **require** them to set it every time, rather than only overriding when relevant. I think leaving it the way it is currently written _is_ the more flexible choice.

alcuadrado:
> > Static call can also have value, so maybe you can move value and from to FunctionCallFuture
> 
> Are you sure about this @alcuadrado ? I'm pretty sure sending value is considered modifying state and isn't allowed in a static call. Plus static calls don't create transactions on-chain, and sending value would create one.

Yes, you can think of static calls as dry-runs of transactions, and it's normally implemented like that. For example, many people do that to see if a tx would succeed and only send it if it does.

> They're optional because we default the entire options to an empty object if it isn't given. It is still being defaulted to zero if it isn't given, but it should stay optional just like after and the rest of the options.

ok, let's keep it optional

> I would argue that defaulting to accounts[0] is pretty standard. We can still allow execution strategies to override that if they choose, but leaving from as undefined would require them to set it every time, rather than only overriding when relevant. I think leaving it the way it is currently written is the more flexible choice.

I think `from` here should represent the sender that does what the future represents. For example, in a deployment future with `from` the contract's constructor should get that value as `msg.sender`.

In the simplest execution strategy (1 tx per future) that would always be possible and simple to achieve. You just send the transaction that executes the future from that account.

Some other execution strategies will want to deploy contracts from other contracts, like a create2 factory, and those won't be able to support `from`, as the `msg.sender` of the constructor will be the factory. There's still going to be an account sending the transactions, but that will probably be an ignition-wide setting that will default to `acccount[0]`.

Letting `undefined` flow to the execution strategy let's us implement this.

PR #231, opened 2023-05-19
add getParameter to new API

### LINKED ISSUE #226: Add module parameters into the new api
For our initial pass we are going to avoid explicit module -> submodule parameter passing. Instead a module will be able to access the global parameters context. If the parameter context is accessed a default value has to be provided as well (this is a constraint we can loosen later with validation of required parameters).

Only solidity parameter value types can be passed via `getOptionalParameter`

Example: 

```
const supply = m.getOptionalParameter<number>("supply", 1000);
const ticker = m.getOptionalParameter<string>("ticker", "ETH");

const c = m.contract("Token", [ticker, supply]);
```

The return value should be a value, not a future. The parameters provided will be injected into the constructor constructor constructor.

zoeyTM:
Included both required and optional parameters in the same implementation because it seemed like a clean and simple solution to me. Happy to refactor back to just the `getOptioinalParam` version if we prefer separate implementations like in the old API. 

fixes #226

### DIFF
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -62,7 +62,7 @@ export class ModuleConstructor {
   constructor(
     public readonly chainId: number,
     public readonly accounts: string[],
-    public readonly parameters: ModuleParameters = {}
+    public readonly parameters: { [moduleId: string]: ModuleParameters } = {}
   ) {}
 
   public construct<
@@ -96,7 +96,7 @@ export class ModuleConstructor {
         mod,
         this.chainId,
         this.accounts,
-        this.parameters
+        this.parameters[moduleDefintion.id]
       )
     );
 
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -2,7 +2,7 @@ import assert from "assert";
 import { inspect } from "util";
 
 import { IgnitionValidationError } from "../../errors";
-import { ArtifactType, SolidityParamsType } from "../stubs";
+import { ArtifactType, SolidityParamType, SolidityParamsType } from "../stubs";
 import {
   ArtifactContractDeploymentFuture,
   ArtifactLibraryDeploymentFuture,
@@ -10,6 +10,7 @@ import {
   ContractFuture,
   IgnitionModule,
   IgnitionModuleResult,
+  ModuleParameters,
   NamedContractCallFuture,
   NamedContractDeploymentFuture,
   NamedLibraryDeploymentFuture,
@@ -60,7 +61,8 @@ export class ModuleConstructor {
 
   constructor(
     public readonly chainId: number,
-    public readonly accounts: string[]
+    public readonly accounts: string[],
+    public readonly parameters: ModuleParameters = {}
   ) {}
 
   public construct<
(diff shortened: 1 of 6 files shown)


packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -363,6 +367,22 @@ export class IgnitionModuleBuilderImplementation<
     return future;
   }
 
+  public getParameter<ParamType extends SolidityParamType>(
```
alcuadrado, line 370:
```suggestion
  public getParameter<ParamType extends SolidityParamType = any>(
```

I checked what we discussed earlier today, and if we add this it can lead to a nicer UX. Otherwise you'd have to cast `unknowns` into other types when using params without a default value.

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -59,7 +60,8 @@ export class ModuleConstructor {
 
   constructor(
     public readonly chainId: number,
-    public readonly accounts: string[]
+    public readonly accounts: string[],
+    public readonly parameters: ModuleParameters = {}
```
alcuadrado, line 64: I think this should get a `ModuleParameters` per module. More like:
```suggestion
    public readonly parameters: {[moduleId: string]: ModuleParameters} = {}
```

packages/core/test/new-api/getParameter.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    assert.equal(contractFuture.constructorArgs.length, 1);
+    assert.equal(contractFuture.constructorArgs[0], "arg1");
+  });
+
+  it("should allow a parameter to have a default value", () => {
+    const moduleWithParamsDefinition = defineModule("Module1", (m) => {
+      const arg1 = m.getParameter<string>("param1", "arg1");
+      const arg2 = m.getParameter<number>("param2", 42);
```
alcuadrado, line 44: These `<types>` can be removed and they are infered

packages/core/test/new-api/getParameter.ts
```diff
@@ -0,0 +1,87 @@
...
+describe("getParameter", () => {
+  it("should record given parameters", () => {
+    const constructor = new ModuleConstructor(0, [], { param1: 42 });
+
+    assert.equal(constructor.parameters.param1, 42);
+  });
+
+  it("should allow a parameter to be passed as an arg", () => {
+    const moduleWithParamsDefinition = defineModule("Module1", (m) => {
+      const arg1 = m.getParameter<string>("param1");
```
alcuadrado, line 16: If we do the `= any` that we discussed this `<string>` can be removed, and same in the other non-optional params.

packages/core/src/new-api/internal/module-builder.ts
zoeyTM: done!

kanej:
Can I check, with the use of `id` against parameters in the constructor, are we saying that `getParameter` is scoped to only that module?

So I don't pass in one set of parameters available to all modules and submodule instead I pass parameters for a paticular module?

zoeyTM:
> So I don't pass in one set of parameters available to all modules and submodule instead I pass parameters for a particular module?

correct @kanej

alcuadrado (approved):
lgtm

PR #232, opened 2023-05-19
Add readEventArgument action

alcuadrado:
This PR adds ModuleBuilder#readEventArgument.

It also fixes an error message that I believe got outdated due to a merging order issue.

I intentionally left out of this PR using the values read by these futures, as I think that requires some work across the whole module api.

### DIFF
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -599,7 +599,7 @@ export class IgnitionModuleBuilderImplementation<
   private _assertUniqueReadEventArgumentId(futureId: string) {
     return this._assertUniqueFutureId(
       futureId,
-      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.readEventArgument("MyContract", "0x123...", artifact, { id: "MyId"})\``,
+      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.readEventArgument(myContract, "MyEvent", "eventArg", { id: "MyId"})\``,
       this.contractAt
     );
   }
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -446,12 +446,12 @@ export class IgnitionModuleBuilderImplementation<
   ): ReadEventArgumentFuture {
     const eventIndex = options.eventIndex ?? 0;
 
-    const futureToReadFromsContract =
+    const contractToReadFrom =
       "contract" in futureToReadFrom
         ? futureToReadFrom.contract
         : futureToReadFrom;
 
-    const emitter = options.emitter ?? futureToReadFromsContract;
+    const emitter = options.emitter ?? contractToReadFrom;
 
     const id =
       options.id ??
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -16,6 +16,7 @@ import {
   NamedContractDeploymentFuture,
   NamedLibraryDeploymentFuture,
   NamedStaticCallFuture,
+  ReadEventArgumentFuture,
 } from "../types/module";
 import {
   CallOptions,
@@ -26,6 +27,7 @@ import {
   IgnitionModuleDefinition,
   LibraryFromArtifactOptions,
   LibraryOptions,
+  ReadEventArgumentOptions,
 } from "../types/module-builder";
 
 import {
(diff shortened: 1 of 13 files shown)


packages/core/test/new-api/readEventArgument.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    it("Should accept an explicit emitter", () => {
+      const defintion = defineModule("Module1", (m) => {
+        const contract = m.contract("Contract");
+        const call = m.call(contract, "fuc");
+
+        const contract2 = m.contract("Contract2");
+
+        m.readEventArgument(contract, "EventName1", "arg1");
+        m.readEventArgument(call, "EventName2", "arg2", { emitter: contract2 });
```
kanej, line 73: When do I want to change the contract, can you walk me through the use case?

packages/core/test/new-api/readEventArgument.ts
```diff
@@ -0,0 +1,211 @@
...
+
+      const [read1, read2] = Array.from(mod.futures).filter(
+        (f) => f.type === FutureType.READ_EVENT_ARGUMENT
+      ) as ReadEventArgumentFuture[];
+
+      assert.equal(read1.emitter, mod.results.contract);
+      assert.equal(read2.emitter, mod.results.contract2);
+    });
+
+    it("Should set the right eventName and argumentName", () => {
```
kanej, line 89: Are you wanting all `Should`'s capitalized in this case? Or is it a _don't care, its a convention not worth enforcing_?

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -364,6 +367,52 @@ export class IgnitionModuleBuilderImplementation<
...
+      | NamedContractDeploymentFuture<string>
+      | ArtifactContractDeploymentFuture
+      | NamedContractCallFuture<string, string>,
+    eventName: string,
+    argumentName: string,
+    options: ReadEventArgumentOptions = {}
+  ): ReadEventArgumentFuture {
+    const eventIndex = options.eventIndex ?? 0;
+
+    const futureToReadFromsContract =
```
zoeyTM, line 381: for readability and clarity, can we change this variable name to `contractToReadFrom` or something similar?

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -364,6 +367,52 @@ export class IgnitionModuleBuilderImplementation<
     return future;
   }
 
+  public readEventArgument(
+    futureToReadFrom:
+      | NamedContractDeploymentFuture<string>
+      | ArtifactContractDeploymentFuture
+      | NamedContractCallFuture<string, string>,
+    eventName: string,
+    argumentName: string,
```
zoeyTM, line 376: what if a user wants to read more than one argument from the event?

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -459,7 +508,15 @@ export class IgnitionModuleBuilderImplementation<
...
-      `Contracts must have unique ids, ${futureId} has already been used, ensure the id passed is unique \`m.contractAt("MyContract", "0x123...", artifact, { id: "MyId"})\``,
+      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.contractAt("MyContract", "0x123...", artifact, { id: "MyId"})\``,
+      this.contractAt
+    );
+  }
+
+  private _assertUniqueReadEventArgumentId(futureId: string) {
+    return this._assertUniqueFutureId(
+      futureId,
+      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.readEventArgument("MyContract", "0x123...", artifact, { id: "MyId"})\``,
```
zoeyTM, line 519:
```suggestion
      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.readEventArgument(myContract, "MyEvent", "eventArg", { id: "MyId"})\``,
```

zoeyTM (changes requested):

zoeyTM:
Also I think this needs a rebase and will need to add support to the serialization, ui, and batching code

packages/core/test/new-api/readEventArgument.ts
kanej: The answer to this is that a call to a contract can delegate on to a whole chain of further contract function calls. Hence we need a way of saying which contract in the call chain is the emitter, not necessarily the top level contract that was called.

packages/core/src/new-api/internal/module-builder.ts
alcuadrado: I'd say they create two features for now. If this becomes a recurrent complain we extend this api.

alcuadrado:
> the serialization, ui, and batching code

Yeah, I realized that after seeing your PR. I'll propose a change to the types that will make this mistake a compilation or linter error.

alcuadrado:
> Also I think this needs a rebase and will need to add support to the serialization, ui, and batching code

Done

zoeyTM (approved):
lgtm

PR #235, opened 2023-05-23
Support passing futures as addresses in contractAt

### LINKED ISSUE #225: enhance contractAt to support passing in futures as addresses
The first step here is to drop up an implementation plan.

We want to support:

```

const contract = m.contract("Contract");

const address = m.call(contract, "getAnotherContractAddress"); // This could be a call or event future.

const anotherContract = m.contractAt("AnotherContract", address, artifact);

return { anotherContract };
```

If contract at is provided with a future that resolves to a usable address, this should work.

zoeyTM:
- added support for passing static call futures into `contractAt` as an address
- added named and artifact variants for `contractAt` futures
- updated plan UI to handle new features

fixes #225

### DIFF
--- a/packages/core/src/new-api/stored-deployment-serializer.ts
+++ b/packages/core/src/new-api/stored-deployment-serializer.ts
@@ -273,8 +273,8 @@ export class StoredDeploymentSerializer {
             partialFutureLookup[
               (
                 (serializedFuture as SerializedNamedContractAtFuture)
-                  .address as unknown as NamedStaticCallFuture<string, string>
-              ).id
+                  .address as unknown as FutureToken
+              ).futureId
             ] as NamedStaticCallFuture<string, string>
           );
         }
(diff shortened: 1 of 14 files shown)


packages/core/src/new-api/stored-deployment-serializer.ts
```diff
@@ -262,8 +265,32 @@ export class StoredDeploymentSerializer {
...
-        // no future specific tokens to resolve
+      } else if (future instanceof NamedContractAtFutureImplementation) {
+        if (typeof future.address !== "string") {
+          this._overwriteReadonly(
+            future,
+            "address",
+            partialFutureLookup[
+              (
+                (serializedFuture as SerializedNamedContractAtFuture)
+                  .address as unknown as FutureToken
```
zoeyTM, line 276: i am all ears if there's a better solution than this :pray:

alcuadrado (approved):
LGTM, but it'd be better if @kanej can review the comment in the serialization module.

PR #236, opened 2023-05-23
Introduce Future, ContractFuture, DeploymentFuture, and FunctionCallFuture as union types and remove ResultT from Future.

alcuadrado:
Apart from introducing the union types that the title mentions, it makes all base types / classes private to the module that defines them.

When updating the UI I removed the `Ui*` types, as discussed with @kanej. I also removed some unneeded type guards because they were always `true`.

I recommend reviewing this commit by commit.


zoeyTM (approved):
This is looking really great!

packages/core/test/new-api/types/module.ts
```diff
@@ -0,0 +1,16 @@
...
+} from "../../../src/new-api/types/module";
+
+interface BaseFuture {
+  id: string;
+  type: FutureType;
+  module: IgnitionModule;
+  dependencies: Set<Future>;
+}
+
+const _testThatEveryFutureIsBaseFuture: Future extends BaseFuture
```
kanej, line 14: Living in the future.

PR #237, opened 2023-05-23
feat/reconciliation (PR 2)

kanej:
Add reconciliation stage to processing in prep for execution. Key elements are mocked out, but this will provide a base for adding further reconciliation rules.

## TODO

- [x] Per type reconciliation
- [x] Dependency reconciliation
- [x] swap to lodash


packages/core/package.json
```diff
@@ -90,6 +91,7 @@
   "dependencies": {
     "@ethersproject/address": "5.6.1",
     "debug": "^4.3.2",
+    "deep-equal": "2.2.1",
```
alcuadrado, line 94: I'd rather use `lodash` and import `lodash/isEqual`, as it has 0 deps (vs 18)

kanej:
I am holding off on artifact reconciliation until we have the execution engine (and the artifact use cases it will need).

packages/core/src/new-api/deployer.ts
```diff
@@ -0,0 +1,60 @@
...
+  public async deploy(
+    moduleDefinition: IgnitionModuleDefinition<
+      string,
+      string,
+      IgnitionModuleResult<string>
+    >,
+    moduleParameters: ModuleParameters,
+    accounts: string[]
+  ) {
+    const module = this._moduleConstructor.construct(moduleDefinition);
```
zoeyTM, line 28: Why are we doing this here? My understanding was that we wanted `construct` and `deploy` to be entirely separate and distinct processes, so that you could take any output of `construct` and pass it into a `deploy`.

packages/core/src/new-api/types/module.ts
```diff
@@ -376,4 +376,5 @@ export interface IgnitionModule<
   futures: Set<Future>; // Future created in this module — All of them have to be deployed before returning the results. Note that not all of them are in results
   submodules: Set<IgnitionModule>; // Modules used by this module — Note that there's only one instance of each module
   results: IgnitionModuleResultsT; // The futures returned by the callback passed to buildModule
+  getFutures(): Future[];
```
alcuadrado, line 379: Not sure if I'd add this here, as it restrain us a bit from creating other (static, like yml) representations of the modules.

Or maybe I'm overreacting? I don't have a clear mental model of how having logic here and a static representation would work, nor how/why this method is exposed to the users instead of being an internal utility.

packages/core/src/new-api/deployer.ts
alcuadrado: Agreed, but as we are removing the `ModuleConstructor` I wouldn't block on this.

packages/core/src/new-api/types/execution-engine.ts
```diff
@@ -0,0 +1,8 @@
+import { ExecutionStateMap } from "./execution-state";
+import { IgnitionModule, IgnitionModuleResult } from "./module";
+
+export interface ExecutionEngineState {
```
alcuadrado, line 4: This should be an internal type

packages/core/src/new-api/types/execution-state.ts
```diff
@@ -120,32 +120,63 @@ export interface DeploymentExecutionState
   > {
   storedArtifactPath: string; // As stored in the deployment directory.
   storedBuildInfoPath?: string; // As stored in the deployment directory. Optional as it's not always present
+  contractName: string;
```
alcuadrado, line 123: These types should be internal

packages/core/src/new-api/types/execution-state.ts
```diff
@@ -120,32 +120,63 @@ export interface DeploymentExecutionState
...
+  from: string | undefined;
 }
 
-type CallExecutionState = BaseExecutionState<FutureType.NAMED_CONTRACT_CALL>;
+export interface StaticCallExecutionState
+  extends BaseExecutionState<FutureType.NAMED_STATIC_CALL> {
+  contractAddress: string;
+  functionName: string;
+  args: ArgumentType[];
+  result?: ArgumentType;
```
alcuadrado, line 145: We may need to use a Buffer here and store the non-decoded version, as we may get errors or invalid ABI encoded values. 

I guess we also want to store if the execution was successful or not. In the other execution states we can check if they failed by looking at their txs, but not in this case.

packages/core/src/new-api/types/execution-state.ts
```diff
@@ -120,32 +120,63 @@ export interface DeploymentExecutionState
...
+  contractName: string;
+  address: string;
+}
 
-type ContractAtExecutionState = BaseExecutionState<
-  FutureType.NAMED_CONTRACT_AT | FutureType.ARTIFACT_CONTRACT_AT
->;
+export interface ReadEventArgumentExecutionState
+  extends BaseExecutionState<FutureType.READ_EVENT_ARGUMENT> {
+  eventName: string;
```
alcuadrado, line 159: This lacks the information about which transaction the event should be read from.

packages/core/src/new-api/types/execution-state.ts
alcuadrado: > I guess we also want to store if the execution was successful or not
 self-answer: that's part of the `BaseExecutionState`

packages/core/src/new-api/internal/reconciliation/reconcilier.ts
```diff
@@ -0,0 +1,115 @@
...
+  ReconciliationContext,
+  ReconciliationFailure,
+  ReconciliationFutureResult,
+  ReconciliationFutureResultFailure,
+  ReconciliationResult,
+} from "./types";
+
+export class Reconcilier {
+  public static reconcile(
+    module: IgnitionModule,
```
alcuadrado, line 20: Maybe worth making it explicit if this includes the futures of submodules

alcuadrado: I'm pretty sure it does, but while reviewing this file I found myself having to double check it by reading the code.

packages/core/src/new-api/internal/reconciliation/reconcileDependencyRules.ts
```diff
@@ -0,0 +1,44 @@
...
+    const additionalExecutionState = context.executionStateMap[additionalDep];
+
+    if (additionalExecutionState === undefined) {
+      return fail(
+        future,
+        `A dependency from ${future.id} to the a new future ${additionalDep} has been added`
+      );
+    }
+
+    if (additionalExecutionState.status === ExecutionStatus.SUCCESS) {
```
alcuadrado, line 33:
```suggestion
    // TODO: Check that is was successfully executed before `executionState` was created.
    if (additionalExecutionState.status === ExecutionStatus.SUCCESS) {
```

packages/core/src/new-api/internal/reconciliation/reconcileDependencyRules.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const additionalDeps = difference(currentDeps, previousDeps);
+
+  for (const additionalDep of additionalDeps) {
+    const additionalExecutionState = context.executionStateMap[additionalDep];
+
+    if (additionalExecutionState === undefined) {
+      return fail(
+        future,
+        `A dependency from ${future.id} to the a new future ${additionalDep} has been added`
```
alcuadrado, line 29:
```suggestion
        `A dependency from ${future.id} to ${additionalDep} has been added. The former has started executing before the latter started executing, so this change is incompatible.`
```

alcuadrado: I added a bit more of context to the error message here, as they weren't obvious to me, so I guess it would be less clear to users.

packages/core/src/new-api/internal/reconciliation/reconcilier.ts
```diff
@@ -0,0 +1,115 @@
...
+    executionStateMap: ExecutionStateMap,
+    moduleParameters: ModuleParameters,
```
alcuadrado, line 22: I think this should be a `Map` from module id to `ModuleParameters`. I believe this got lost in the PR that introduced runtime values.

packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractAt.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function reconcileArtifactContractAt(
+  future: ArtifactContractAtFuture,
+  executionState: ContractAtExecutionState,
+  context: ReconciliationContext
+): ReconciliationFutureResult {
+  if (!isEqual(future.contractName, executionState.contractName)) {
+    return fail(
+      future,
+      `Contract name has been changed from ${executionState.contractName} to ${future.contractName}`
```
alcuadrado, line 17: Maybe these errors should include the id. In general, all of them.

You could have the same contract name in multiple modules.

alcuadrado: Oh, the future id is not lost.

packages/core/src/new-api/internal/reconciliation/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function resolveFromAddress(
+  from: string | AccountRuntimeValue | undefined,
+  context: ReconciliationContext
+): string | undefined {
+  if (from === undefined || typeof from === "string") {
+    return from;
+  }
+
+  if (typeof from === "string") {
```
alcuadrado, line 34: This is never true because of the if in line 30

packages/core/src/new-api/internal/reconciliation/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return from;
+  }
+
+  if (!isRuntimeValue(from)) {
+    throw new IgnitionError(`Could not resolve from address: ${from as any}`);
+  }
+
+  const runtimeAddress = context.accounts[from.accountIndex];
```
alcuadrado, line 46: I think this could also be a param

alcuadrado: Oh, no, that's not the case with `from`.

packages/core/src/new-api/internal/reconciliation/futures/reconcileNamedStaticCall.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  if (!isEqual(future.functionName, executionState.functionName)) {
+    return fail(
+      future,
+      `Function name has been changed from ${executionState.functionName} to ${future.functionName}`
+    );
+  }
+
+  if (!isEqual(future.args, executionState.args)) {
```
alcuadrado, line 31: Won't these comparison of `args` fail if they have a future or runtime value? Same for the ones that receive `constructorArgs`.

I think I'd try to avoid using `isEqual` unless its really necessary, as it can lead to this situations of incomplete comparison.

packages/core/src/new-api/internal/reconciliation/reconcileFutureSpecificReconciliations.ts
```diff
@@ -0,0 +1,94 @@
...
+export function reconcileFutureSpecificReconciliations(
+  future: Future,
+  executionState: ExecutionState,
+  context: ReconciliationContext
+): ReconciliationFutureResult {
+  switch (future.type) {
+    case FutureType.NAMED_CONTRACT_DEPLOYMENT:
+      return reconcileNamedContractDeployment(
+        future,
+        executionState as DeploymentExecutionState,
```
alcuadrado, line 33: Is this what you'd like to type more precisely?

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    const contractAddress = this.resolveFromExecutionState(
+      address,
+      executionStateMap,
+      (exState: DeploymentExecutionState) => exState.contractAddress
+    );
+
+    if (contractAddress === undefined) {
+      throw new IgnitionError("Previous deployment without contractAddress");
```
alcuadrado, line 54: I think we should try to give more contextual information in these error messages. Is it added later? Otherwise this would be really hard to debug.

packages/core/test/new-api/reconciliation/futures/reconcileArtifactContractDeployment.ts
```diff
@@ -0,0 +1,222 @@
...
+        failure:
+          "Contract name has been changed from ContractUnchanged to ContractChanged",
+      },
+    ]);
+  });
+
+  it("should find changes to constructors unreconciliable", () => {
+    const moduleDefinition = defineModule("Module", (m) => {
+      const contract1 = m.contractFromArtifact("Contract1", fakeArtifact, [
+        "changed",
```
alcuadrado, line 109: It would be cool to have a test with more complex arguments, including a future.

alcuadrado: Same for the positive tests

packages/core/test/new-api/reconciliation/reconciler.ts
```diff
@@ -0,0 +1,244 @@
...
+      "Module1:ContractMissed": {
+        ...exampleDeploymentState,
+        status: ExecutionStatus.STARTED,
+      },
+    });
+
+    assert.deepStrictEqual(
+      reconiliationResult.missingExecutedFutures,
+      ["Module1:ContractMissed"],
+      "Exected one missing previous executed future"
```
alcuadrado, line 56:
```suggestion
      "Expected one missing previous executed future"
```

packages/core/test/new-api/reconciliation/reconciler.ts
```diff
@@ -0,0 +1,244 @@
...
+          ...exampleDeploymentState,
+          futureType: FutureType.NAMED_CONTRACT_DEPLOYMENT,
+          status: ExecutionStatus.SUCCESS,
+          dependencies: new Set(["Module:Contract1", "Module:Contract2"]),
+          contractName: "Contract3",
+        },
+      });
+    });
+
+    it("should reconcile the addition of a dependency to a completed future", () => {
```
alcuadrado, line 153: Note that this will be more strict in the future. Maybe add a todo?

alcuadrado:
I left multiple comments, but I like this PR.

Do we have an issue to track the reconciliation of artifacts?

packages/core/src/new-api/types/module.ts
kanej: Ha! This was my one concession to object-orientation - I wanted to `getFutures` on modules just about everywhere. More than happy to move it back to being a utility function.

packages/core/src/new-api/deployer.ts
kanej: I have been using `deployer` to capture the complete deployment process, but it is just a shim to let me test from the cli/tests.

Feel free to replace it with the `ModuleConstructor` refactor.

packages/core/src/new-api/types/execution-state.ts
kanej: I am not following. Do you mean the execution state should also record the transactionId for the transaction that emits the event?

packages/core/src/new-api/internal/reconciliation/reconcilier.ts
kanej: Agreed, would changing the name of the function to `reconcileModuleAndSubmodules` clarify, or is there a different approach (e.g. a dreaded comment)?

packages/core/src/new-api/internal/reconciliation/reconcileFutureSpecificReconciliations.ts
kanej: Yes, the types between future and execution state have a mirroring hierarchy. I would prefer to avoid the casting; however it is not a big risk given the tests.

packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractAt.ts
```diff
@@ -0,0 +1,32 @@
...
+    );
+  }
+
+  const resolvedAddress: string =
+    ExecutionStateResolver.resolveStaticCallToAddress(future.address, context);
```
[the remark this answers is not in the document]
kanej: I have renamed it to `resolveToAddress`, as you are right. It was refactored for multiple usages but not renamed.

packages/core/src/new-api/internal/reconciliation/utils.ts
kanej: Nice catch. I have removed the string comparison on line 30, so this check is now hit.

packages/core/src/new-api/internal/reconciliation/futures/reconcileNamedStaticCall.ts
kanej: I have pulled out the serialization replacement logic and used it to make the future comparison work with nested futures and runtime args.

kanej:
> I left multiple comments, but I like this PR.
> 
> Do we have an issue to track the reconciliation of artifacts?

I have added the artifact reconciliation here: https://github.com/NomicFoundation/ignition/issues/269

packages/core/test/new-api/reconciliation/reconciler.ts
kanej: Can you elaborate how it will become more strict?

packages/core/test/new-api/reconciliation/futures/reconcileArtifactContractDeployment.ts
kanej: I have added to the tests more complex argument example - including some futures, just for fun.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: I have changed a few of these to assertions (they should be picked up by validation or execution state shouldn't get in the state).

I have broadened out some of the errors with more info.

I also added a catch all for exceptions to `reconcileFutureSpecificReconciliations`. The only exceptions should be coming from the supporting reconciliation utility functions. The exception is now caught and returned as a reconciliation failure, but paired with the future's id for context.

packages/core/src/internal/utils/guards.ts
```diff
@@ -160,6 +166,23 @@ export function isLibrary(
...
 
+export function isDeploymentExecutionState(
+  executionState: ExecutionState
+): executionState is DeploymentExecutionState {
+  return [
+    FutureType.NAMED_CONTRACT_DEPLOYMENT,
+    FutureType.ARTIFACT_CONTRACT_DEPLOYMENT,
+    FutureType.NAMED_LIBRARY_DEPLOYMENT,
+    FutureType.ARTIFACT_LIBRARY_DEPLOYMENT,
+  ].includes(executionState.futureType);
```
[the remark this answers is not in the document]
alcuadrado: I forgot to mention this, but I couldn't find an easy way.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -14,12 +15,59 @@ import {
...
 // TODO: consider merging this into the execution state map
 export class ExecutionStateResolver {
+  public static resolveArgsFromExectuionState(
+    constructorArgs: ArgumentType[],
+    context: ReconciliationContext
+  ): ArgumentType[] {
+    return replaceWithinArg<ArgumentType>(constructorArgs, {
+      bigint: (bi) => bi.toString(),
+      future: (f) => {
+        if (!isContractFuture(f)) {
```
alcuadrado: There's also static call and event argument futures

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -14,12 +15,59 @@ import {
...
+          throw new IgnitionError(
+            `Cannot replace future in args, for non-deployable futures ${f.id}`
+          );
+        }
+
+        return ExecutionStateResolver.resolveContractToAddress(f, context);
```
alcuadrado: I believe this can try to resolve futures that were not successfully executed. I couldn't find a check for that.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -14,12 +15,59 @@ import {
...
 import { resolveModuleParameter, safeToString } from "./utils";
 
 // TODO: consider merging this into the execution state map
 export class ExecutionStateResolver {
+  public static resolveArgsFromExectuionState(
+    constructorArgs: ArgumentType[],
+    context: ReconciliationContext
+  ): ArgumentType[] {
+    return replaceWithinArg<ArgumentType>(constructorArgs, {
+      bigint: (bi) => bi.toString(),
```
alcuadrado: This could be left as is, couldn't it?

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -14,12 +15,59 @@ import {
...
+              `No default value provided for module parameter ${mprv.moduleId}/${mprv.name}`
+            );
+          }
+
+          return mprv.defaultValue;
+        }
+
+        return parameter;
+      },
+    }) as ArgumentType[];
```
alcuadrado: Is this really needed? The cast I mean.

packages/core/src/new-api/internal/utils/replace-within-arg.ts
```diff
@@ -0,0 +1,63 @@
...
+    return arg.map((a) => replaceWithinArg(a, replacers));
+  }
+
+  if (typeof arg === "object" && arg !== null) {
+    return Object.fromEntries(
+      Object.entries(arg).map(([k, v]) => [k, replaceWithinArg(v, replacers)])
+    );
+  }
+
+  return arg as any;
```
alcuadrado, line 62: I think this `as any` will hide bugs.

packages/core/src/new-api/internal/utils/replace-within-arg.ts
```diff
@@ -0,0 +1,63 @@
...
+export function replaceWithinArg<T>(
+  arg: ArgumentType,
+  replacers: Replacers<T>
+): Composable<T> {
+  if (typeof arg === "bigint") {
+    return replacers.bigint(arg);
+  }
+
+  if (isFuture(arg)) {
+    return replacers.future(arg);
```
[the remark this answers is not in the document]
alcuadrado: Oh, maybe they are not needed actually.  Is that intentional?

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
```diff
@@ -26,22 +27,25 @@ export class ExecutionStateResolver {
...
+
+          return ExecutionStateResolver.resolveContractToAddress(f, context);
+        },
+        accountRuntimeValue: (arv) => context.accounts[arv.accountIndex],
+        moduleParameterRuntimeValue: (mprv) => {
+          return resolveModuleParameter(mprv, context);
+        },
+      });
+
+    return constructorArgs.map(replace);
```
alcuadrado, line 48: I left a comment about the `as ArgumentType` in a previous commit. It's cool that TS can infer it like this.

packages/core/src/new-api/types/execution-state.ts
alcuadrado: Yes, or maybe just the id of the future that lead to that transaction, but right now there's no way to extract the event's arg because you can't access the event.

packages/core/src/new-api/types/execution-state.ts
```diff
@@ -120,32 +120,63 @@ export interface DeploymentExecutionState
   > {
   storedArtifactPath: string; // As stored in the deployment directory.
   storedBuildInfoPath?: string; // As stored in the deployment directory. Optional as it's not always present
+  contractName: string;
   constructorArgs: ArgumentType[];
   libraries: Record<string, string>; // TODO: Do we need to store their future ids for the reconciliation process?
   value: bigint;
+  from: string | undefined;
```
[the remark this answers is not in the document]
alcuadrado: > the history doesn't apply for ContractAtExecutionState and ReadEventArgumentExecutionState

I think so. We can reintroduce some flexibility there in the future if needed.

> And just to further clarify, I am expecting history to get populated with either onchain interactions themselves, or simplified versions of onchain interactions. Is that how you see it?

What would the simplified versions be? Other than that, yes.

packages/core/test/new-api/reconciliation/reconciler.ts
alcuadrado: You should only be able to add a dependency if it was successfully executed before the dependent started executing. That mimics how the execution without restarts works.

alcuadrado:
I re-reviewed this and resolved most conversations. I left some new comments, some of them in a commit that's now outdated (sorry), but the conversations are still relevant.

packages/core/src/new-api/internal/utils/replace-within-arg.ts
kanej: Agreed. The arg is `string | number | boolean` at that point so I have enhanced the return type to include those possibilities: 

```ts
): Composable<T | string | number | boolean> {
```

I suspect if I was better at typescript I would be able to express that the T generic parameter must be a union type that includes `string | number | boolean`.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: Yes I will replace it with identity.

packages/core/src/new-api/internal/utils/replace-within-arg.ts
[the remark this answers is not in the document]
kanej: I am not expecting to need to recurse _within futures_. The three use cases I have for this function are:

1. Serialize the constructor/call arguments of a future - any futures are replaced with a future token, there is no further recursion into the future
2. Reconciliation to say if constructor/call args are equivalent to what is recorded in the execution state - I am defining equality for futures there to be equality of the future id and relying on the reverse topoligical sort of reconiliation to ensure the equivalence of the future itself
3. Resolution of constructor/call arguments during execution based on the execution state, similarly no recursion happens in to the future, the result/contractAddress are read directly.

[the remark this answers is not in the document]
kanej: The execution time resolution (3 above) is not in this PR, I will roll it into one of the later PRs.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: The `resolveContractToAddress` function has an _assertion_ that a contract address is returned, so we will exception for a missing address with a relevant error:

https://github.com/NomicFoundation/ignition/blob/a65a88667d763b1f7f4e08fe416423861584ec68/packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts#L84

But my use of an assertion implies a stronger check earlier on - that you can't have a dependency on an unstarted future from an already started future. We have this check, but I am now worried about the ordering of the checks.

Let me add a test for this to ensure we check dependencies before we need to do this resolution.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: I like to pretend they don't exist.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: Resolution of arguments happens after dependency checks:

https://github.com/NomicFoundation/ignition/blob/82ecb97f1db8cac47f545d8d8ea61cf372775193/packages/core/src/new-api/internal/reconciliation/reconciler.ts#L32

I have assumed we can't get a request of resolution of an address on a future that has started but not completed.

If `A <- B`, where A is a call and b depends on its result as a constructor arg, B isn't allowed to start until A is complete. B cannot have execution state, hence we never reconcile it. You could change a dep to create this situation, but our dependency check will ensure we don't have both A and B having started.

packages/core/src/new-api/internal/reconciliation/execution-state-resolver.ts
kanej: I have added tests and conditionals for these two cases. It is the last commit. 

Note this required two refactorings worth paying attention to:

1. Add a SolidityParameterType (identical to ModuleParameterType)
2. Add a result to read event state with type SolidityParameterType

It seemed weird to use ModuleParameterType so far outside a module, though the answer may be to rename it. I also needed a result for ReadEventArgumentState, I am happy to make it a more complex object depending on how we expand out the event api.

alcuadrado (approved):
LGTM

PR #238, opened 2023-05-24
Add send to new API

### LINKED ISSUE #234: send - an arbitrary evm call

zoeyTM:
fixes #234

### DIFF
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -486,13 +486,13 @@ export class IgnitionModuleBuilderImplementation<
   }
 
   public send(
+    id: string,
     to: string | AddressResolvableFuture,
-    data: string,
+    value?: bigint,
+    data?: string,
     options: SendDataOptions = {}
   ): SendDataFuture {
-    const id = options.id ?? (typeof to === "string" ? to : to.id);
-    const futureId = `${this._module.id}:${id}`;
-    options.value ??= BigInt(0);
+    const futureId = `${this._module.id}:${options.id ?? id}`;
 
     this._assertUniqueSendId(futureId);
 
@@ -500,8 +500,8 @@ export class IgnitionModuleBuilderImplementation<
       futureId,
       this._module,
       to,
+      value ?? BigInt(0),
       data,
-      options.value,
       options.from
     );
 
--- a/packages/core/src/new-api/internal/module-builder.ts
+++ b/packages/core/src/new-api/internal/module-builder.ts
@@ -644,7 +644,7 @@ export class IgnitionModuleBuilderImplementation<
   private _assertUniqueSendId(futureId: string) {
     return this._assertUniqueFutureId(
       futureId,
-      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.send(myContract, "0xabcd", { id: "MyId"})\``,
+      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.send("MyId", "0xabcd")\``,
       this.send
     );
   }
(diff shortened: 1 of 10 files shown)


packages/core/src/new-api/types/module-builder.ts
```diff
@@ -211,6 +225,12 @@ export interface IgnitionModuleBuilder {
     options?: ReadEventArgumentOptions
   ): ReadEventArgumentFuture;
 
+  send(
+    to: string | AddressResolvableFuture,
+    data: string,
+    options?: SendDataOptions
+  ): SendDataFuture;
```
alcuadrado, line 232: @zoeyTM what if we do
```suggestion
  send(
    id: string,
    to: string | AddressResolvableFuture,
    value?: bigint,
    data?: string,
    options?: SendDataOptions
  ): SendDataFuture;
```

and only leave `after` and `from` in the options?

My reasoning:

1) required `id`, as it's unclear how to infer an useful one.
2) `value` and `data` are both "first-class" arguments, as they are both intrinsic to the operation the user is asking to do, while options are more metadata-like.
3) `value` and `data` are both optional because you may want to use one and/or the other (e.g. send ETH, send data to a non-payable contract)
4) `value` before `data` because it's simpler to omit with a `0` if you only want to send data.

packages/core/src/new-api/types/module.ts
```diff
@@ -225,6 +239,22 @@ export interface ReadEventArgumentFuture {
...
+ *
+ * @beta
+ */
+export interface SendDataFuture {
+  type: FutureType.SEND_DATA;
+  id: string;
+  module: IgnitionModule;
+  dependencies: Set<Future>;
+  to: string | AddressResolvableFuture;
+  data: string;
```
alcuadrado, line 253: Shouldn't this be optional? Or are we going to use `"0x"` as default?

packages/core/src/new-api/types/module-builder.ts
zoeyTM: > required `id`, as it's unclear how to infer an useful one.

definitely agree with this one

> `value` and `data` are both "first-class" arguments, as they are both intrinsic to the operation the user is asking to do, while options are more metadata-like.

i agree and see your point, but i'm torn because i also like the consistency of keeping `value` as an option like it is everywhere else. Ultimately I think I agree with making it a first class param for this function though

zoeyTM: My other thought was actually moving `data` into the options object for this function as well, to better facilitate that any of those fields were optional

packages/core/src/new-api/types/module.ts
zoeyTM: Yep, good catch

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -600,7 +637,15 @@ export class IgnitionModuleBuilderImplementation<
...
       `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.readEventArgument(myContract, "MyEvent", "eventArg", { id: "MyId"})\``,
-      this.contractAt
+      this.readEventArgument
+    );
+  }
+
+  private _assertUniqueSendId(futureId: string) {
+    return this._assertUniqueFutureId(
+      futureId,
+      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.send(myContract, "0xabcd", { id: "MyId"})\``,
```
alcuadrado, line 647:
```suggestion
      `Duplicated id ${futureId} found in module ${this._module.id}, ensure the id passed is unique \`m.send("MyId", "0xabcd")\``,
```

alcuadrado (approved):

alcuadrado:
I left a small fix to an error message, but this LGTM

packages/core/src/new-api/internal/module-builder.ts
zoeyTM: ah good catch!

PR #239, opened 2023-05-24
Introduce RuntimeValues

alcuadrado:
Draft PR introducing the concept of `RuntimeValue`s.

Things missing:

* [x] Test `ModuleBuilder#getAccount`
* [x] Adapt the tests that use `ModuleBuilder#accounts`
* [x] Add support and tests for `AccountRuntimeValue` as `from`.
* [x] Add support and tests for `AccountRuntimeValue` as argument.
* [x] Adapt the UI to support `AccountRuntimeValue`s
* [x] Update `ModuleBuilder#getParameter`
* [x] Test the new `ModuleBuilder#getParameter`
* [x] Adapt the tests that use the old `ModuleBuilder#getParameter`
* [x] Add support and tests for `ModuleParameterRuntimeValue` as argument.
* [x] Adapt the UI to support `ModuleParameterRuntimeValue`s


packages/core/src/index.ts
```diff
@@ -9,6 +9,7 @@ export * from "./new-api/types/artifact";
 export * from "./new-api/types/module";
 export * from "./new-api/types/module-builder";
 export * from "./new-api/types/serialized-deployment";
+export * from "./new-api/type-guards";
```
alcuadrado, line 12: As discussed with John, I made the type guards public.

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -132,13 +132,25 @@ export class IgnitionModuleBuilderImplementation<
...
+  public getParameter<ParamTypeT extends ModuleParameterType>(
+    parameterName: string,
+    defaultValue?: ParamTypeT
+  ): ModuleParameterRuntimeValue<ParamTypeT> {
+    return new ModuleParameterRuntimeValueImplementation(
+      parameterName,
+      defaultValue
+    );
+  }
+
```
alcuadrado, line 153: We return `RuntimeValue`s now.

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -357,7 +369,7 @@ export class IgnitionModuleBuilderImplementation<
 
   public contractAt<ContractNameT extends string>(
     contractName: ContractNameT,
-    address: string | NamedStaticCallFuture<string, string>,
+    address: string | AddressResolvableFuture,
```
alcuadrado, line 372: I updated some types that were wrong. It had no impact in the code apart from the types.

packages/core/src/new-api/stored-deployment-serializer.ts
```diff
@@ -80,21 +91,16 @@ export class StoredDeploymentSerializer {
   ): SerializedStoredModule {
     return {
       id: userModule.id,
-      futures: Object.fromEntries(
-        Array.from(userModule.futures).map((future) => [
-          future.id,
-          this._serializeFuture(future),
-        ])
+      futures: Array.from(userModule.futures).map((future) =>
```
alcuadrado, line 94: I modified the serialization to preserver the other of:

1. Module's futures
2. Module's submodules
3. Module's results
4. Future's libraries

This allowed me to make the tests stricter with `deepEqual`. They failed otherwise, as sets and objects have an order in javascript.

packages/core/src/new-api/stored-deployment-serializer.ts
```diff
@@ -312,10 +334,48 @@ export class StoredDeploymentSerializer {
...
+  private static _serializeModuleParamterRuntimeValue(
+    arg: ModuleParameterRuntimeValue<ModuleParameterType>
+  ): SerializedModuleParameterRuntimeValue {
+    return {
+      _kind: "ModuleParameterRuntimeValue",
+      name: arg.name,
+      defaultValue:
+        arg.defaultValue === undefined
+          ? undefined
+          : this._jsonStringifyWithBigint(arg.defaultValue),
```
alcuadrado, line 362: The serialization of the this default value has to be fairly generic. I decided to use a JSON string, as it can handle everything we need, except `bigint`. For those, I used [a replacer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer).

packages/core/src/new-api/stored-deployment-serializer.ts
```diff
@@ -407,16 +465,21 @@ export class StoredDeploymentDeserializer {
     for (const serializedModule of Object.values(
       serializedDeployment.modules
     )) {
-      for (const [name, futureToken] of Object.entries(
-        serializedModule.results
-      )) {
-        const mod = this._lookup(modulesLookup, serializedModule.id);
+      const mod = this._lookup(modulesLookup, serializedModule.id);
+
```
alcuadrado, line 469: This was making multiple lookups

packages/core/src/new-api/types/module-builder.ts
```diff
@@ -42,7 +44,7 @@ export interface ContractOptions {
   after?: Future[];
   libraries?: Record<string, ContractFuture<string>>;
   value?: bigint;
-  from?: string;
+  from?: string | AccountRuntimeValue;
```
alcuadrado, line 47: `from` can accept an account obtained with `m.getAccount()` or `string`. In theory it could accept a `Future` or module param, but that doesn't seem like a realistic use-case.

packages/core/src/new-api/internal/module-builder.ts
alcuadrado: `getAccount` instead of `accounts`, as we can't pre-populate an array of `AccountRuntimeValue`.

packages/core/src/new-api/types/module.ts
```diff
@@ -31,16 +32,16 @@ export type ArgumentType =
...
-  ARTIFACT_CONTRACT_DEPLOYMENT,
-  NAMED_LIBRARY_DEPLOYMENT,
-  ARTIFACT_LIBRARY_DEPLOYMENT,
-  NAMED_CONTRACT_CALL,
-  NAMED_STATIC_CALL,
-  NAMED_CONTRACT_AT,
-  ARTIFACT_CONTRACT_AT,
-  READ_EVENT_ARGUMENT,
-  SEND_DATA,
+  NAMED_CONTRACT_DEPLOYMENT = "NAMED_CONTRACT_DEPLOYMENT",
```
alcuadrado, line 35: I replaced enums to use strings, as the type guards were getting confused between `RuntimeValueType` and `FutureType`. 

I added a test that ensures that values have to be a valid key.

packages/core/src/new-api/types/module.ts
```diff
@@ -13,7 +13,8 @@ export type BaseArgumentType =
   | boolean
   | ContractFuture<string>
   | NamedStaticCallFuture<string, string>
-  | ReadEventArgumentFuture;
+  | ReadEventArgumentFuture
+  | RuntimeValue;
```
alcuadrado, line 17: `RuntimeValue`s can be used as arguments.

packages/core/src/new-api/types/module.ts
```diff
@@ -295,6 +296,48 @@ export type ModuleParameterType =
...
+  accountIndex: number;
+}
+
+/**
+ * A module parameter.
+ *
+ * @beta
+ */
+export interface ModuleParameterRuntimeValue<
+  ParamTypeT extends ModuleParameterType
```
alcuadrado, line 334: This can't receive any `ArgumentType`, for example, it can't receive a `Future`. That's why we need a different type.

packages/core/test/new-api/call.ts
```diff
@@ -1,10 +1,16 @@
...
-import { NamedContractCallFutureImplementation } from "../../src/new-api/internal/module";
+import {
+  AccountRuntimeValueImplementation,
+  ModuleParameterRuntimeValueImplementation,
+  NamedContractCallFutureImplementation,
+} from "../../src/new-api/internal/module";
 import { ModuleConstructor } from "../../src/new-api/internal/module-builder";
 import { FutureType } from "../../src/new-api/types/module";
 
+import { assertInstanceOf } from "./helpers";
```
alcuadrado, line 12: I created this helper that does an `assert.instaceOf(v, Class)` and also lets TS know that after that line `v` has type `Class`.

@zoeyTM I think you'll like it.

packages/core/test/new-api/call.ts
```diff
@@ -186,6 +192,228 @@ describe("call", () => {
     assert.equal(callFuture.from, "0x2");
   });
 
+  it("Should be able to pass an AccountRuntimeValue as from option", () => {
```
alcuadrado, line 195: I added a `string` and `AccountRuntimeValue` of this test to every future that takes a `from`.

packages/core/test/new-api/call.ts
```diff
@@ -186,6 +192,228 @@ describe("call", () => {
...
+
+    if (!(callFuture instanceof NamedContractCallFutureImplementation)) {
+      assert.fail("Not a named contract deployment");
+    }
+
+    assertInstanceOf(callFuture.from, AccountRuntimeValueImplementation);
+    assert.equal(callFuture.from.accountIndex, 1);
+  });
+
+  describe("Arguments", () => {
```
alcuadrado, line 226: I added test of `ArgumentType`s for every future that takes either constructorArguments or arguments.

packages/core/test/new-api/getAccount.ts
```diff
@@ -0,0 +1,28 @@
+import { assert } from "chai";
+
+import { defineModule } from "../../src/new-api/define-module";
+import { AccountRuntimeValueImplementation } from "../../src/new-api/internal/module";
+import { ModuleConstructor } from "../../src/new-api/internal/module-builder";
+
+import { assertInstanceOf } from "./helpers";
+
+describe("getAccount", () => {
```
alcuadrado, line 9: Just a simple test verifying that getAccount returns the right object.

packages/core/test/new-api/getParameter.ts
```diff
@@ -1,93 +1,128 @@
...
 
 import { defineModule } from "../../src/new-api/define-module";
-import { NamedContractDeploymentFutureImplementation } from "../../src/new-api/internal/module";
+import { ModuleParameterRuntimeValueImplementation } from "../../src/new-api/internal/module";
 import { ModuleConstructor } from "../../src/new-api/internal/module-builder";
+import { ModuleParameterType } from "../../src/new-api/types/module";
+
+import { assertInstanceOf } from "./helpers";
 
 describe("getParameter", () => {
```
alcuadrado, line 10: Replaced these tests with a pretty simple version. They just test that it returns the right object, and that it supports all the types.

packages/core/test/new-api/stored-deployment-serializer.ts
```diff
@@ -572,6 +572,54 @@ describe("stored deployment serializer", () => {
   });
 
   describe("Complex arguments serialization", () => {
+    it("Should support base values as arguments", () => {
```
alcuadrado, line 575: These tests are based on `m.contract`'s ones.

packages/core/test/new-api/stored-deployment-serializer.ts
```diff
@@ -632,21 +744,31 @@ describe("stored deployment serializer", () => {
...
     2
   );
 
   assert.equal(
     serialized,
     reserialized,
+    "Module serialization not the same across serialization/deserialization"
+  );
+
+  assert.deepEqual(
```
alcuadrado, line 769: I really wanted to add this test, as it was hard to convince myself that this worked correctly.

This lead me to changing the serialization to preserve the order of a few things.

packages/core/src/new-api/types/module.ts
kanej: For debugging purposes, I prefer string enums anyway. But then no one ever accused me of being memory efficient.

packages/core/test/new-api/getParameter.ts
```diff
@@ -1,93 +1,128 @@
...
 
-      return { contract1 };
+      const param = mod.results.contract.constructorArgs[0];
+      assertInstanceOf(param, ModuleParameterRuntimeValueImplementation);
+      assert.equal(param.name, "p");
+      assert.deepEqual(param.defaultValue, defaultValue);
     });
 
-    const constructor = new ModuleConstructor(0, []);
+    it("Should accept complex combinations as deafult", () => {
```
kanej, line 107:
```suggestion
    it("Should accept complex combinations as default", () => {
```

kanej (approved):
Looks good.

kanej:
Can I check the intention on runtime values. Those will be checked and resolved during reconciliation/execution by resolving the passed accounts/module params?

PR #240, opened 2023-05-25
Replace stub types

alcuadrado:


kanej (approved):
Looks good. The reverse topological sort is a nice touch for removing the overwrites.

We should probably add tests to cover complex argument serialization/deserialization.

PR #254, opened 2023-05-31
Add validation to new API (PR 4)

### LINKED ISSUE #228: ModuleBuilder validations and Context-free module validations
Port existing validations that are possible at the level of internal checks within the module builder - without accessing the file system or external data.

Any validation that exists in the current phase but cannot be done with the passed in future information should be done in a new separate module validation phase.

A validation failure should throw a `IgnitionValiditionError`.

For the moment, the module builder validation can be added in directly to the module builder. The module validation phase can be encapsulated in a `./src/new-api/validation/validate.ts` function that operates on IgnitionModule.

To take an example, `validateDeployedContract`, checking that value is a `bigint` (rather than BigNumber) can be done in the builder, checking the named contract exists would be done in the validate function as part of module validation.

zoeyTM:
fixes #228


packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -396,8 +436,6 @@ export class IgnitionModuleBuilderImplementation<
       future.dependencies.add(address);
     }
 
-    // TODO: Validate the the runtime value's default type is string
```
zoeyTM, line 399: I want to note that currently, enforcing this the way I did has introduced the problem that module params without default values are considered invalid due to them getting assigned an `any` type by default.

packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
 export * from "./errors";
 export * from "./initialization";
 export { defineModule } from "./new-api/define-module";
+export { Deployer } from "./new-api/deployer";
```
alcuadrado, line 5: This should be private

packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
...
 /* TODO: how is module constructor getting exposed? */
 export { ModuleConstructor } from "./new-api/internal/module-builder";
+export { FileJournal, MemoryJournal } from "./new-api/journal";
```
alcuadrado, line 8: Same here.

alcuadrado: Maybe we should clean up the repository and this file, because its getting tricky to understand what should be or not public and why.

(for example, there's a TODO I missed in line 6)

alcuadrado: Oh, I think the comments I left here should have been submitted in #255

alcuadrado: I changed the base branch of this pr to reflect that it's based on #255

packages/core/src/new-api/internal/module-builder.ts
alcuadrado: Not sure if I understand this. But this TODO, which I thing I wrote, is incorrect, right? Default values can have types other than strings.

packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -651,4 +698,86 @@ export class IgnitionModuleBuilderImplementation<
...
+  private _assertValidAddress(
+    address:
+      | string
+      | AddressResolvableFuture
+      | ModuleParameterRuntimeValue<string>,
+    func: (...[]: any[]) => any
+  ) {
+    if (
+      typeof address !== "string" &&
+      !isModuleParameterRuntimeStringValue(address) &&
```
alcuadrado, line 768: I think I understand your comment now. I'd say we accept module params without default here, and then validate that they actually exist and are a `string` in the second phase.

packages/core/src/new-api/internal/validation/futures/validateNamedContractDeployment.ts
```diff
@@ -0,0 +1,30 @@
...
+import { IgnitionValidationError } from "../../../../errors";
+import { isArtifactType } from "../../../type-guards";
+import { ArtifactResolver } from "../../../types/artifact";
+import { NamedContractDeploymentFuture } from "../../../types/module";
+
+export async function validateNamedContractDeployment(
+  future: NamedContractDeploymentFuture<string>,
+  artifactLoader: ArtifactResolver
+) {
+  const artifact = await artifactLoader.load(future.contractName);
```
alcuadrado, line 12: Should we handle the case were the name doesn't exist? Or maybe just let the adapter throw an error, letting it do whatever it wants.

packages/core/src/new-api/internal/validation/futures/validateNamedContractCall.ts
```diff
@@ -0,0 +1,53 @@
...
+import { IgnitionValidationError } from "../../../../errors";
+import { isArtifactType } from "../../../type-guards";
+import { ArtifactResolver } from "../../../types/artifact";
+import { NamedContractCallFuture } from "../../../types/module";
+
+export async function validateNamedContractCall(
+  future: NamedContractCallFuture<string, string>,
+  artifactLoader: ArtifactResolver
+) {
+  const artifact = await artifactLoader.load(future.contract.contractName);
```
alcuadrado, line 12: `future.contract` can may not be a named contract, but have the artifact inline instead.

packages/core/src/new-api/internal/validation/futures/validateNamedStaticCall.ts
```diff
@@ -0,0 +1,62 @@
...
+import { IgnitionValidationError } from "../../../../errors";
+import { isArtifactType } from "../../../type-guards";
+import { ArtifactResolver } from "../../../types/artifact";
+import { NamedStaticCallFuture } from "../../../types/module";
+
+export async function validateNamedStaticCall(
+  future: NamedStaticCallFuture<string, string>,
+  artifactLoader: ArtifactResolver
+) {
+  const artifact = await artifactLoader.load(future.contract.contractName);
```
alcuadrado, line 12: Same here. The contract may have the artifact, and in that case the name doesn't need to be related to the artifact loader.

packages/core/src/new-api/internal/validation/futures/validateReadEventArgument.ts
```diff
@@ -0,0 +1,35 @@
...
+import { IgnitionValidationError } from "../../../../errors";
+import { isArtifactType } from "../../../type-guards";
+import { ArtifactResolver } from "../../../types/artifact";
+import { ReadEventArgumentFuture } from "../../../types/module";
+
+export async function validateReadEventArgument(
+  future: ReadEventArgumentFuture,
+  artifactLoader: ArtifactResolver
+) {
+  const artifact = await artifactLoader.load(future.emitter.contractName);
```
alcuadrado, line 12: Same comment about artifacts here.

packages/core/src/new-api/internal/validation/validate.ts
```diff
@@ -0,0 +1,48 @@
...
+      case FutureType.NAMED_CONTRACT_CALL:
+        return validateNamedContractCall(future, artifactLoader);
+      case FutureType.NAMED_STATIC_CALL:
+        return validateNamedStaticCall(future, artifactLoader);
+      case FutureType.READ_EVENT_ARGUMENT:
+        return validateReadEventArgument(future, artifactLoader);
+      case FutureType.SEND_DATA:
+        return validateSendData(future, artifactLoader);
+      default:
+        throw new IgnitionValidationError(`Unknown future type`);
```
alcuadrado, line 45: I think that we our linter rules this is not necessary, as the linter fail if the switch is not exhaustive.

alcuadrado (commented):
This looks good. I couldn't review the tests in depth yet, but they also look good.

I left a few comments, and I think I caught a bug.

packages/core/src/new-api/internal/validation/futures/validateNamedContractDeployment.ts
zoeyTM: In this specific case, our adapter (i.e. `hre`) does throw an error. I'm on board with continuing to delegate that responsibility to the adapter

alcuadrado (approved):

Issue #109, opened 2023-01-17
Named deployment to support multiple deployments per module


kanej:
Currently a module can be deployed to completion once. It can be rerun if only partially executed or errored; but while display the results of the previous run if rerun on a successful deploy.

To support using a module multiple time we could introduce the concept of a deployment id. So a module could have multiple deployments, each with a unique module id.

This raises design questions about composing sections of the module that refer to singleton instances of contracts on a deploy relative to contracts which are redeployed as new instances each run.

alcuadrado:
I think we can have a "global" deployment id, so that we scope everything by it. That is, the journal files, deployment artifacts, or whatever we want to store.

We should default de deployment id to something like `"network-${chainId}"`.

kanej:
This is something we want to support. Right now there is one journal file per module. We want to switch it to a folder structure that supports different deployment ids and chains and whatever other feature the journal needs scoped by.

The next step is to confirm the list of features to scope by (currently chainId and deploymentId) and design a folder layout that will support them.

PR #255, opened 2023-06-01
feat/basic-execution (PR 3)

kanej:
Sketch out the components of the execution engine:

* execution engine - orchestrator
* journal - record and recover transaction log
* execution strategy - an extension point for turning futures into
  multiple on-chain interactions
* transaction service - manages on-chain interactions
* chain-dispatcher - low-level calls to the blockchain

Resolves https://github.com/NomicFoundation/ignition/issues/256

## Preview

deploy2

## Details 

An sketch of the components is in place, but only named contract deploy has been driven through all the steps.

A `deploy2` command has been added to the plugin to allow testing, in prep for it replacing `deploy` proper.

## Testing

After a local build you should be able to test the `./examples/sample` with:

```sh
npx hardhat deploy2 ./ignition/LockModule.js
```


packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
 export * from "./errors";
 export * from "./initialization";
 export { defineModule } from "./new-api/define-module";
+export { Deployer } from "./new-api/deployer";
```
alcuadrado, line 5: As mentioned in #254 by mistake, many of the newly exported things here should be private.

It's probably time to clean up the repo, as it's getting tricky to understand what should or should not be exported and why.

packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
...
 /* TODO: how is module constructor getting exposed? */
 export { ModuleConstructor } from "./new-api/internal/module-builder";
+export { FileJournal, MemoryJournal } from "./new-api/journal";
```
alcuadrado, line 8: Should be private

packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
...
 export { StoredDeploymentSerializer } from "./new-api/stored-deployment-serializer";
+export * from "./new-api/type-guards";
+export * from "./new-api/types/adapters";
 export * from "./new-api/types/artifact";
+export * from "./new-api/types/deployer";
+export * from "./new-api/types/journal";
```
alcuadrado, line 14: Also private

packages/core/src/index.ts
```diff
@@ -2,14 +2,20 @@ export { buildModule } from "./buildModule";
...
 export * from "./new-api/types/module";
 export * from "./new-api/types/module-builder";
 export * from "./new-api/types/serialized-deployment";
-export * from "./new-api/type-guards";
+export * from "./new-api/types/transaction-service";
```
alcuadrado, line 18: Also private

packages/core/src/index.ts
alcuadrado: I marked the new things that I believe should be private. I think it would be more clear if they were defined in `internal/`

packages/core/src/new-api/internal/reconciliation/utils.ts
```diff
@@ -68,7 +67,7 @@ export function resolveModuleParameter(
   return moduleParamValue;
 }
 
-export function safeToString(potential: ModuleParameterType | undefined) {
+export function safeToString(potential: unknown) {
```
alcuadrado, line 70: TBH I don't like this api. What does `safe` mean? When can I use this? What should I expect to get? The `unknown` makes it harder to accept, as it now can accept anything, leading to mistakes.

I'd rather have one or more api with a clear semantics, like `moduleArgumentToErrorString`.

packages/core/src/new-api/deployer.ts
```diff
@@ -1,17 +1,69 @@
...
+  private _strategy: ExecutionStrategy;
+
+  constructor(
+    options:
+      | { journal: Journal; transactionService: TransactionService }
+      | {
+          journal: Journal;
+          adapters: Adapters;
+          artifactLoader: ArtifactResolver;
+        }
```
alcuadrado, line 42: Why the union type here? It's confusing.

alcuadrado: The main thing that confuses me is that they are not equivalent things.

alcuadrado: Have you considered creating multiple static factories and making the constructor private instead?

packages/core/src/new-api/deployer.ts
```diff
@@ -1,17 +1,69 @@
...
+  ) {
+    this._journal = options.journal;
+    this._strategy = new BasicExecutionStrategy();
+
+    if ("adapters" in options && isAdapters(options.adapters)) {
+      if (options.artifactLoader === undefined) {
```
alcuadrado, line 48: Why can this be undefined if it's required?

packages/core/src/new-api/internal/execution/executionStateReducer.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (action.type === "execution-start") {
+    return {
+      ...executionStateMap,
+      [action.futureId]: initialiseExecutionStateFor(action),
+    };
+  }
+
+  if (action.type === "execution-success") {
+    const updatedExecutionState: DeploymentExecutionState = {
+      ...(executionStateMap[action.futureId] as DeploymentExecutionState),
```
alcuadrado, line 23: Why is this `as` required? I think it would be invalid without some extra checks.

packages/core/src/new-api/internal/execution/executionStateReducer.ts
```diff
@@ -0,0 +1,59 @@
...
+      status: ExecutionStatus.SUCCESS,
+      contractAddress: action.contractAddress,
+    };
+
+    return {
+      ...executionStateMap,
+      [action.futureId]: updatedExecutionState,
+    };
+  }
+
```
alcuadrado, line 33: Are the intermediate events between start and success missing on purpose?

packages/core/src/new-api/types/journal.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * A journal message indicating a future's execution was not completed within
+ * the timeout.
+ *
+ * @beta
+ */
+export interface ExecutionHold {
+  type: "execution-hold";
+}
```
alcuadrado, line 193: These feel incomplete. Is is because they were left out of scope for this PR?

packages/core/src/new-api/types/journal.ts
```diff
@@ -0,0 +1,197 @@
+import { ArgumentType, FutureType } from "./module";
+
+/**
+ * Store a deployments execution state as a transaction log.
+ *
+ * @beta
+ */
+export interface Journal {
+  record(message: JournalableMessage): Promise<void>;
```
alcuadrado, line 9: I don't think this should be async. To ensure some reasonable durability of the journal we should append synchronously making sure that everything is flushed before continuing the execution. I've already did some research about how to achieve this in node.

packages/core/src/new-api/types/journal.ts
```diff
@@ -0,0 +1,197 @@
...
+export type TransactionMessage = OnchainInteraction | OnchainResult;
+
+// #region "OnchainInteraction"
+
+/**
+ * A on-chain interaction request for the transaction service.
+ *
+ * @beta
+ */
+export type OnchainInteraction = DeployContractInteraction;
```
alcuadrado, line 37: I think it would be easier to understand if all these have a shared name pattern, like all of them ending in `Message` or `JournalMessage`.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -0,0 +1,219 @@
...
+
+      const batchResult = await this._executeBatch(executionBatch, state);
+
+      if (batchResult.some((b) => b.type === "execution-failure")) {
+        return { status: "failed" };
+      }
+
+      if (batchResult.some((b) => b.type === "execution-hold")) {
+        return { status: "hold" };
+      }
```
alcuadrado, line 49: These returns loose data, you don't know what failed or hold later.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -0,0 +1,219 @@
...
+    future: Future,
+    state: ExecutionEngineState
+  ): FutureStart | FutureRestart {
+    const executionState = state.executionStateMap[future.id];
+
+    if (executionState === undefined) {
+      return this._initCommandFor(future, state);
+    }
+
+    return { type: "execution-restart", futureId: future.id };
```
alcuadrado, line 111: What does restart mean here?

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -0,0 +1,219 @@
...
+      executionState: state.executionStateMap[future.id],
+    };
+
+    const exectionStrategy = state.strategy.executeStrategy(context);
+
+    while (!isExecutionResult(current)) {
+      context.executionState = state.executionStateMap[future.id];
+
+      if (isOnChainAction(current)) {
+        current = await state.transactionService.onchain(current);
```
alcuadrado, line 88: I think this is missing the fact that the transaction service will need to do things that must be journaled internally (e.g. gas bump). Have you thought about it? Should it receive the entire state to be able to mutate it?

alcuadrado: Or maybe I'm missing how the transaction service should work

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -0,0 +1,219 @@
...
+    message: JournalableMessage
+  ): Promise<void> {
+    await state.journal.record(message);
+    state.executionStateMap = executionStateReducer(
+      state.executionStateMap,
+      message
+    );
+  }
+
+  private _resolveAddress(
```
alcuadrado, line 125: Thinking out loud: this feels a bit duplicated wrt the reconciliation pr, maybe we can avoid the repetition (not convinced it's worth it though)

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -0,0 +1,219 @@
...
+        const state: FutureStart = {
+          type: "execution-start",
+          futureId: future.id,
+          futureType: future.type,
+          strategy,
+          // status: ExecutionStatus.STARTED,
+          dependencies: [...future.dependencies].map((f) => f.id),
+          // history: [],
+          storedArtifactPath: "./artifact.json",
+          storedBuildInfoPath: "./build-info.json",
```
alcuadrado, line 167: These should be created here

packages/core/src/new-api/internal/execution/transactions/transaction-service.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    return this._dispatchDeployContract(interaction);
+  }
+
+  private async _dispatchDeployContract(
+    deployContractInteraction: DeployContractInteraction
+  ): Promise<DeployContractResult> {
+    // TODO: consider replacing this with a registry of artifacts
+    const artifact = await this._artifactLoader.load(
```
alcuadrado, line 41: I think this PR misses the fact that we need two artifact adapters. One for the project's, one for the deployment's artifacts.

alcuadrado: Also, we can't fit a registry of artifacts in memory, as mentioned in the design doc.

packages/core/src/new-api/internal/execution/transactions/transaction-service.ts
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @beta
+ */
+export class TransactionServiceImplementation implements TransactionService {
+  constructor(
+    private _artifactLoader: ArtifactResolver,
+    private _chainDispatcher: ChainDispatcher
+  ) {}
+
+  public async onchain(
```
alcuadrado, line 25: I think the missing part I don't fully grasp is how this would work with things like re-sending a transaction because it was dropped, bumping gas, etc.

packages/core/src/new-api/journal.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @beta
+ */
+export class MemoryJournal implements Journal {
+  private messages: string[] = [];
+
+  public async record(message: JournalableMessage): Promise<void> {
+    try {
+      this.messages.push(JSON.stringify(message));
+    } catch (err) {
+      console.error(err, message);
```
alcuadrado, line 17: I guess this was here for debugging purposes, but it's a (semi)silent error.

packages/core/src/new-api/journal.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  public async *read(): AsyncGenerator<JournalableMessage> {
```
alcuadrado, line 21: Using an async generator :o I never thought about this possibility.

packages/core/src/new-api/types/artifact.ts
```diff
@@ -12,3 +12,12 @@ export interface Artifact {
...
   >;
 }
+
+/**
+ * Retrieve artifacts based on contract name.
+ *
+ * @beta
+ */
+export interface ArtifactResolver {
+  load(contractName: string): Promise<Artifact>;
```
alcuadrado, line 22: I mentioned this somewhere else, but we need two different objects to resolve artifacts.

One is public, which does this, plus resolving `build-info`s. I would copy these signatures from HH:
* https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/types/artifacts.ts#L13
* https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/types/artifacts.ts#L42

The other one is internal, and reads/writes artifacts from the deployment directory.

packages/core/src/new-api/types/deployer.ts
```diff
@@ -0,0 +1,16 @@
...
+export type DeploymentResult =
+  | {
+      status: "success";
+      contracts: Record<
+        string,
+        { contractName: string; contractAddress: string }
+      >;
+    }
+  | {
+      status: "failed" | "hold";
```
alcuadrado, line 15: This needs more info.

packages/core/src/index.ts
kanej: I am going to push on these in this PR. I have added this issue to track:

https://github.com/NomicFoundation/ignition/issues/273

packages/core/src/index.ts
kanej: Tracked by: https://github.com/NomicFoundation/ignition/issues/273

packages/core/src/index.ts
kanej: Tracked by: https://github.com/NomicFoundation/ignition/issues/273

packages/core/src/index.ts
kanej: Tracked by: https://github.com/NomicFoundation/ignition/issues/273

packages/core/src/new-api/internal/reconciliation/utils.ts
kanej: Agreed. I have added `accountRuntimeValueToErrorString`, `moduleParameterToErrorString` and `addressToErrorString` as a replacement.

packages/core/src/new-api/deployer.ts
kanej: Tracked by: https://github.com/NomicFoundation/ignition/issues/273

packages/core/src/new-api/deployer.ts
kanej: This has been pulled up, I think in the validation branch.

packages/core/src/new-api/internal/execution/executionStateReducer.ts
kanej: Yes, I want to tackle this while doing restartability: https://github.com/NomicFoundation/ignition/issues/258

packages/core/src/new-api/types/journal.ts
kanej: Yes this will come in with restartability: https://github.com/NomicFoundation/ignition/issues/258

packages/core/src/new-api/types/journal.ts
kanej: I have added this as a todo in the restartability issue: https://github.com/NomicFoundation/ignition/pull/255#discussion_r1228729914

packages/core/src/new-api/internal/execution/executionStateReducer.ts
kanej: I have added an invariant that checks for deployment execution state.

For the moment static call, call and read event execution state are not covered, but there are issues to cover those todos.

packages/core/src/new-api/types/journal.ts
kanej: I have appended `Message` to each messages name.

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: This might not make sense.

I want us to get to restarting a future in the transaction service. I am not sure as whether the transaction service detects the restart or we do it based on the previously recorded execution state.

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: Agreed. I have added todo items to deal with failed and hold in the restartability issue: #258

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: This will be done as part of https://github.com/NomicFoundation/ignition/issues/109

packages/core/src/new-api/internal/execution/transactions/transaction-service.ts
kanej: After our discussions, a deployment loader that contains and exposes the journal will be added as part of https://github.com/NomicFoundation/ignition/issues/109

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: I just attempted to pull it out. But the resolveAddress logic is subtely different from reconciliation.

Specifically in reconciliation, we treat `from` set to undefined as a valid value. Here we replace undefined with `Accounts[0]`.

Because accounts can change between runs, should reconciliation keep the current behaviour of allowing undefined, or should it substitute in `accounts[0]`?

packages/core/src/new-api/types/artifact.ts
kanej: As mentioned above this is coming in as part of #109

packages/core/src/new-api/types/deployer.ts
kanej: Tracked in https://github.com/NomicFoundation/ignition/issues/258

packages/core/src/new-api/internal/execution/execution-engine.ts
alcuadrado: I think once we started executing a future with a certain from that should be considered fixed. If during a restart the future doesn't specify a `from`, and resolving `accounts[0]` leads to a different account, that should be a reconciliation error.

alcuadrado (approved):
Reviewed this again. I think it looks good as most of my comments were about missing things that are now tracked in issues.

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: Cool. That will allow us to share the logic. I will add a todo to pull that out.

Issue #258, opened 2023-06-02
Monitoring and tx resumption in execution engine


kanej:
Bring back the capability at the cli to restart a previously run deployment. Restarting should cover:

* Failed deployments
* Modified deployments that pass reconciliation

The `force` flag should also be brought back to support ignoring previous run state.

## TODO

- [x] #268
- [x] #269
- [x] #274 
- [x] #260
- [x] #293
- [x] Review journal signature - write should be sync https://github.com/NomicFoundation/ignition/pull/255#discussion_r1228729914
- [x] Consider polling interval config - https://github.com/NomicFoundation/ignition/pull/261#discussion_r1235196013

PR #261, opened 2023-06-02
feat/running-from-tests (PR 4)

### LINKED ISSUE #257: Support usage of deployer in tests

kanej:
Add a second ignition object to hre that is a test helper with one method deploy. The return from deploy are transformed into ethers contracts for use in tests.

Resolves #257.

## Details

The definition of done for this PR is that running tests within the `./examples/sample` project via `npx hardhat test` should pass.

This involved some more work that just adding the new ignition object, specifically:

* wiring through deployment parameters into the execution engine
* recursive resolution of args in execution (to support the HH sample project use of getParameter)
* refactor to move execution state type guards into `new-api/internal`

There is also an additional method on the artifactResolver to load the path. This is only used in tests for the moment. The interaction of artifactResolver and deployment file writing I will clarify in https://github.com/NomicFoundation/ignition/issues/109.


examples/sample/ignition/LockModule.js
```diff
@@ -1,19 +1,19 @@
...
 const TEN_YEARS_IN_FUTURE = currentTimestampInSeconds + TEN_YEAR_IN_SECS;
 
 const ONE_GWEI = hre.ethers.utils.parseUnits("1", "gwei").toString();
 
 module.exports = defineModule("LockModule", (m) => {
-  // const unlockTime = m.getOptionalParam("unlockTime", TEN_YEARS_IN_FUTURE);
-  // const lockedAmount = m.getOptionalParam("lockedAmount", ONE_GWEI);
+  const unlockTime = m.getParameter("unlockTime", TEN_YEARS_IN_FUTURE);
+  // TODO: do we intend to support runtime values for value?
+  // const lockedAmount = m.getParameter("lockedAmount", ONE_GWEI);
```
kanej, line 13: Can I confirm we intend to expand `value` to cover runtime values?

packages/core/src/internal/utils/guards.ts
```diff
@@ -155,23 +149,6 @@ export function isLibrary(
   return future.type === "library";
 }
 
-export function isDeploymentExecutionState(
```
kanej, line 158: These were wrongly outside of our `new-api` folder, and have been moved to `new-api/internal/type-guards.ts`

packages/core/src/new-api/deployer.ts
```diff
@@ -31,11 +31,11 @@ export class Deployer {
   private _moduleConstructor: ModuleConstructor;
   private _executionEngine: ExecutionEngine;
   private _transactionService: TransactionService;
-  private _artifactLoader: ArtifactResolver;
   private _strategy: ExecutionStrategy;
+  private _artifactResolver: ArtifactResolver;
 
   constructor(
-    options: { journal: Journal; artifactLoader: ArtifactResolver } & (
+    options: { journal: Journal; artifactResolver: ArtifactResolver } & (
```
kanej, line 38: A refactor to make the name and property match. This constructor has a todo for a complete rewrite.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -175,7 +198,9 @@ export class ExecutionEngine {
           // status: ExecutionStatus.STARTED,
           dependencies: [...future.dependencies].map((f) => f.id),
           // history: [],
-          storedArtifactPath: "./artifact.json",
+          storedArtifactPath: await artifactResolver.resolvePath(
```
kanej, line 201: This will change drastically with the deployment loader introduction in https://github.com/NomicFoundation/ignition/issues/109. But to get different values in tests I needed a minimal path resolution.

examples/sample/ignition/LockModule.js
```diff
@@ -1,19 +1,19 @@
...
 
-  const lock = m.contract("Lock", [TEN_YEARS_IN_FUTURE], {
-    // args: [unlockTime],
-    // value: lockedAmount
+  const lock = m.contract("Lock", [unlockTime], {
+    value: BigInt(ONE_GWEI),
```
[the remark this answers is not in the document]
zoeyTM: I believe this is intended. I think we decided to swap to native js bignumber instead

examples/sample/ignition/LockModule.js
alcuadrado: yes, it should support them

examples/sample/ignition/LockModule.js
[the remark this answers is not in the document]
alcuadrado: Yes, and ethers v6 uses native bigint

packages/core/src/new-api/internal/utils/resolve-module-parameter.ts
```diff
@@ -0,0 +1,39 @@
...
+      `No default value provided for module parameter ${moduleParamRuntimeValue.moduleId}/${moduleParamRuntimeValue.name}`
+    );
+
+    return moduleParamRuntimeValue.defaultValue;
+  }
+
+  return moduleParamValue;
+}
+
+export function resolveFuture() {}
```
alcuadrado, line 39:
```suggestion
```

I guess this was left here by accident

packages/core/src/new-api/types/artifact.ts
```diff
@@ -20,4 +20,5 @@ export interface Artifact {
  */
 export interface ArtifactResolver {
   load(contractName: string): Promise<Artifact>;
+  resolvePath(contractName: string): Promise<string>;
```
alcuadrado, line 23: Ideally we read the values instead of paths, as that offers more flexibility to implementers. 

I know this will change in another PR though.

packages/core/test/new-api/execution/execution-engine.ts
```diff
@@ -237,6 +266,144 @@ describe("execution engine", () => {
...
+          storedArtifactPath: "/user/path/Library1.json",
+          storedBuildInfoPath: "./build-info.json",
+          contractName: "Library1",
+          value: BigInt(0).toString(),
+          constructorArgs: [],
+          libraries: {},
+          from: accounts[0],
+        },
+        {
+          type: "onchain-action",
```
alcuadrado, line 337: I think I caught this quiet late, but shouldn't these messages have the futureId too?

packages/hardhat-plugin/src/ignition-test-helper.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+    const result = await deployer.deploy(
+      ignitionModuleDefinition,
+      parameters,
+      accounts
+    );
+
+    if (result.status !== "success") {
+      throw new IgnitionError("Failed deployment");
```
alcuadrado, line 45:
```suggestion
      // TODO: Show more information about why it failed
      throw new IgnitionError("Failed deployment");
```

packages/hardhat-plugin/src/ignition-test-helper.ts
```diff
@@ -0,0 +1,89 @@
...
+  ModuleParameters,
+} from "@ignored/ignition-core";
+import { Contract } from "ethers";
+import fs from "fs-extra";
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+
+import { buildAdaptersFrom } from "./buildAdaptersFrom";
+import { buildArtifactResolverFrom } from "./buildArtifactResolverFrom";
+
+export class IgnitionTestHelper {
```
alcuadrado, line 17: Just a clarification, in Hardhat you can't expose a function exclusively to tests, so this should also work for production deployments

packages/hardhat-plugin/src/index.ts
```diff
@@ -235,7 +245,8 @@ task("deploy2")
         parameters = resolveParametersString(parametersInput);
       }
 
-      const isHardhatNetwork = hre.network.name === "hardhat";
+      // TODO: bring back check with file journaling proper
+      const isHardhatNetwork = true; // hre.network.name === "hardhat";
```
alcuadrado, line 249: This check should be done with a request to`web3_clientVersion`

alcuadrado (approved):
I left a few minor comments

examples/sample/ignition/LockModule.js
kanej: I have added an issue for this enhancement here: https://github.com/NomicFoundation/ignition/issues/285

examples/sample/ignition/LockModule.js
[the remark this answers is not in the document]
kanej: Excellent.

packages/hardhat-plugin/src/index.ts
```diff
@@ -89,6 +90,15 @@ extendEnvironment((hre) => {
       networkName: hre.network.name,
     });
   });
+
+  hre.ignition2 = lazyObject(() => {
+    // const isHardhatNetwork = hre.network.name === "hardhat";
+
+    // TODO: rewire txPollingInterval
+    // const txPollingInterval = isHardhatNetwork ? 100 : 5000;
```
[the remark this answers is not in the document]
kanej: Captured here: https://github.com/NomicFoundation/ignition/issues/258

packages/core/test/new-api/execution/execution-engine.ts
kanej: Yes, in fact we need to identify the future and the particular on-chain action.

They are added in the restart PR: https://github.com/NomicFoundation/ignition/blob/b61f9ac8a933db91d3dd937cec3bd25a3bdfe2c3/packages/core/src/new-api/types/journal.ts#L55

TransactionId is a bad name on reflection, maybe chainActionId would be better but we can discuss there.

packages/core/src/new-api/types/artifact.ts
kanej: I have read the comments on this in the later PR. I think we need to walk through the memory and file implementations and then figure our how abstract vs how much delayed file resolution (reading the artifact on demand) we want.

I am going to close this comment in favour of that later discussion.

packages/hardhat-plugin/src/ignition-test-helper.ts
kanej: I think the current implementation makes sense in scripts as well. I will rename the class to `IgnitionHelper` to not confuse by emphasizing one usage.

packages/hardhat-plugin/src/index.ts
kanej: Cool. I will do that change in the store to file/deployment loader PR as that is where the check has moved.

kanej: After discussion we opted to leave the check. We want ignition in tests to be ephemeral but ignition against a localhost node to record a file deployment.

PR #279, opened 2023-06-20
feat/deployment-to-file (PR 5)

kanej:
Save a deployment to disk as it progresses.

Running `npx hardhat deploy2` with a network that isn't hardhat (localhost node but not tests or ephemeral scripts) will write out the deployment to `./ignition/deployments/<deployment-id>`.

The artifacts and buildinfos used during execution are copied here as are the journal and deployed addresses.

A quick first version of `ignition-info` task has been added as well to visualize the `deployed_addresses.json`. It accepts an optional `--json` flag.

Some minimal refactoring has been done to reduce the api exposed from core. And as such Deployer is now internal (but used from core tests) and a `deploy` function is now exported from core (I can't help myself).

## Preview

Jun-22-2023 20-28-49

## Testing

To run locally:

```sh
cd examples/sample
npx hardhat node
# in another tab
cd examples/sample
npx hardhat deploy2 ./ignition/LockModule.js --network localhost
npx hardhat ignition-info2 --deployment-id network-31337 --json
```

## TODO

- [x] Create id identified deployment folder on full run
- [x] Move journal file within folder
- [x] Copy artifacts json and buildinfo per node
- [x] write out deployed_addresses.json
- [x] Add cli command to read deployed_addresses.json (maybe with --json) flag
- [x] Refactor to reduce journal/transaction service api exposure from core


packages/core/package.json
```diff
@@ -61,6 +61,7 @@
     "@types/debug": "^4.1.7",
     "@types/fs-extra": "^9.0.13",
     "@types/mocha": "9.1.1",
+    "@types/ndjson": "2.0.1",
```
kanej, line 64: This is needed for the file journal

packages/core/package.json
```diff
@@ -91,8 +92,10 @@
     "@ethersproject/address": "5.6.1",
     "debug": "^4.3.2",
     "ethers": "^5.4.7",
+    "fs-extra": "^10.0.0",
```
kanej, line 95: The file access for the file deployment loader needs this. We could look at simplifying to just `fs` but I am not sure it is worth the trouble.

packages/core/src/new-api/deploy.ts
```diff
@@ -0,0 +1,57 @@
...
+import { DeploymentResult } from "./types/deployer";
+import { IgnitionModuleResult, ModuleParameters } from "./types/module";
+import { IgnitionModuleDefinition } from "./types/module-builder";
+
+/**
+ * Deploy an IgnitionModule to the chain
+ *
+ * @beta
+ */
+export function deploy({
```
kanej, line 17: This is the new entry point from plugin. We can use deployer from here and core tests, while not exposing more than the signature of this method.

Happy to split/refactor/classify this function, but I wanted remove journal/deployer/transactionservice from being exported.

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -0,0 +1,126 @@
...
+    const json = await fs.readFile(
+      path.join(this._paths?.deploymentDir, storedArtifactPath)
+    );
+
+    const artifact = JSON.parse(json.toString());
+
+    return artifact;
+  }
+
+  public async recordDeployedAddress(
```
kanej, line 102: This is what populated `deployed_addresses.json`

packages/core/src/new-api/internal/deployment-loader/memory-deployment-loader.ts
```diff
@@ -0,0 +1,55 @@
...
+    contractAddress: string
+  ): Promise<void> {
+    this._deployedAddresses[futureId] = contractAddress;
+  }
+
+  public async storeArtifact(
+    _futureId: string,
+    artifact: Artifact
+  ): Promise<string> {
+    const artifactPath = await this._artifactResolver.resolvePath(
```
kanej, line 39: This is a quirk of wanting to return the file path of the original contract artifact under `./artifacts` but starting with the `artifact` object.

An alternative that removes this dep on the artifactResolver would be to pass in both artifact and its original path.

kanej: The resolvePath function on artifactResolver exists solely for this use case.

packages/hardhat-plugin/src/index.ts
```diff
@@ -382,6 +386,56 @@ task("ignition-info")
     }
   );
 
+task("ignition-info2")
+  .addParam("deploymentId")
```
kanej, line 390: For the moment assume that the id is passed id.

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  public async storeArtifact(
+    futureId: string,
+    artifact: Artifact
+  ): Promise<string> {
+    assertIgnitionInvariant(
+      this._paths !== null,
+      "Cannot record deploy address until initialized"
+    );
```
kanej, line 59: I was tempted to have a deployment loader and a deployment loader factory, with the factory doing the initialization and returning a fully setup deployment loader; but I think I can live with these checks over the added complexity.

packages/core/src/new-api/internal/deployer.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    const reconciliationResult = Reconciler.reconcile(
+      module,
+      previousStateMap,
+      deploymentParameters,
+      accounts
+    );
+
+    if (reconciliationResult.reconciliationFailures.length > 0) {
+      throw new Error("Reconciliation failed");
```
alcuadrado, line 75:
```suggestion
      // TODO: Provide more information
      throw new Error("Reconciliation failed");
```

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -0,0 +1,131 @@
...
+    buildInfoDir: string;
+    journalPath: string;
+    deployedAddressesPath: string;
+  } | null = null;
+
+  constructor(private readonly _deploymentDirPath: string) {
+    this.journal = new MemoryJournal();
+  }
+
+  public async initialize(): Promise<void> {
```
alcuadrado, line 26: I think it would be better to make this private and call `await initialize()` at the start of every public method.

This requires turning `initialize()` into a no-op if run twice.

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
alcuadrado: oh, i just proposed an alternative in a comment above

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -0,0 +1,131 @@
...
+    );
+
+    // TODO: should this be made async to be closer to a single fs transaction?
+    const json = (
+      await fs.readFile(this._paths.deployedAddressesPath)
+    ).toString();
+
+    const deployedAddresses: { [key: string]: string } = JSON.parse(
+      json === "" ? "{}" : json
+    );
```
alcuadrado, line 122: TBH I found the `ensureFile` calls above a bit odd because I expected that to bring problems when reading the empty file

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -0,0 +1,131 @@
...
+  public async recordDeployedAddress(
+    futureId: string,
+    contractAddress: string
+  ): Promise<void> {
+    assertIgnitionInvariant(
+      this._paths !== null,
+      "Cannot record deploy address until initialized"
+    );
+
+    // TODO: should this be made async to be closer to a single fs transaction?
```
alcuadrado, line 115: I don't think that's necessary. We can always recreate this file based on the journal too.

packages/core/src/new-api/internal/deployment-loader/memory-deployment-loader.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  private _deployedAddresses: { [key: string]: string };
+
+  constructor(private _artifactResolver: ArtifactResolver) {
+    this.journal = new MemoryJournal();
+    this._deployedAddresses = {};
+  }
+
+  public async loadArtifact(storedArtifactPath: string): Promise<Artifact> {
+    const json = await fs.promises.readFile(storedArtifactPath);
```
alcuadrado, line 19: I don't understand this. How does the in memory one read from the fs? I must be missing something.

packages/core/src/new-api/internal/journal/file-journal.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A file-based journal.
+ *
+ * @beta
+ */
+export class FileJournal implements Journal {
+  constructor(private _filePath: string) {}
+
+  public async record(message: JournalableMessage): Promise<void> {
+    return fs.promises.appendFile(
```
alcuadrado, line 18: I think we should do something like this to write:

```ts
function appendJsonLine(path: string, value: unknown) {
  const fd = openSync(
    path,
    constants.O_CREAT | // Create it if it doesn't exist
      constants.O_WRONLY | // Write only
      constants.O_APPEND | // Append
      constants.O_DSYNC | // Synchronous I/O waiting for writes of content and metadata
      constants.O_DIRECT // Minimize caching
  );

  writeFileSync(fd, `\n${JSON.stringify(value)}`, "utf-8");
  closeSync(fd);
}
```

alcuadrado: Note that it would always leave an empty first line, which i guess should be ok

packages/core/src/new-api/internal/journal/utils/serialize-replacer.ts
```diff
@@ -0,0 +1,34 @@
...
+    return Object.fromEntries(value);
+  }
+
+  if (typeof value === "bigint") {
+    return `${value.toString(10)}n`;
+  }
+
+  if (value instanceof Error) {
+    return serializeError(new Error(value.message));
+  }
```
alcuadrado, line 21: I think this is unsafe and we shouldn't try to serialize errors altogether. We are tricking ourselves into thinking that we are storing them, but we are loosing info (e.g. the stack trace).

packages/core/src/new-api/internal/journal/utils/serialize-replacer.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (value instanceof Set) {
+    return Array.from(value).sort();
+  }
+
+  if (value instanceof Map) {
+    return Object.fromEntries(value);
+  }
+
+  if (typeof value === "bigint") {
+    return `${value.toString(10)}n`;
```
alcuadrado, line 16: This feels a bit risky, not too risky, but we could do something with less changes of clashing, like `{$bigint: "123"}`.

packages/core/src/new-api/types/artifact.ts
```diff
@@ -19,6 +19,126 @@ export interface Artifact {
  * @beta
  */
 export interface ArtifactResolver {
-  load(contractName: string): Promise<Artifact>;
-  resolvePath(contractName: string): Promise<string>;
+  loadArtifact(contractName: string): Promise<Artifact>;
+  getBuildInfo(contractName: string): Promise<BuildInfo | undefined>;
+  resolvePath(contractName: string): Promise<string | undefined>;
```
alcuadrado, line 24: Already mentioned this in a previous PR, but ideally we shouldn't use these paths. Not using them adds flexibility to the implementer of this interface (i.e. integrators of ignition)

packages/core/src/new-api/types/deployment-loader.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Read and write to the deployment storage.
+ *
+ * @beta
+ */
+export interface DeploymentLoader {
+  journal: Journal;
+  initialize(): Promise<void>;
+  loadArtifact(storedArtifactPath: string): Promise<Artifact>;
```
alcuadrado, line 12: I don't understand which path this receives, the internal to the deployment or the one that the Resolver uses?

packages/core/src/new-api/types/journal.ts
```diff
@@ -109,7 +109,7 @@ export interface FutureStartMessage {
   strategy: string;
   dependencies: string[];
   storedArtifactPath: string;
-  storedBuildInfoPath: string;
+  storedBuildInfoPath: string | undefined;
```
alcuadrado, line 112: good catch

alcuadrado (approved):
I left a few comments, but I really like this PR. Everything starts to take more shape now :)

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
kanej: I have removed the ensure file and swapped in a file exists check.

packages/core/src/new-api/internal/journal/file-journal.ts
kanej: I have switched to this implementation and made the `record` function sync.

packages/core/src/new-api/internal/deployment-loader/file-deployment-loader.ts
kanej: I made `initialize` private and idempotent.

packages/core/src/new-api/internal/journal/utils/serialize-replacer.ts
kanej: I switched to use the same serialization as in the `stored-deployment-serializer` so `{_kind: "bigint", value: n.toString(10)}`.

packages/core/src/new-api/internal/deployment-loader/memory-deployment-loader.ts
kanej: `memory-deployment-loader` is a misnomer. Maybe `ephemeral-deployment-loader`, this effectively a deferral of the artifact loading to within the execution engine; but that loading is from the hardhat artifacts path.

packages/core/src/new-api/internal/journal/utils/serialize-replacer.ts
kanej: I have pulled this out as its own issue: https://github.com/NomicFoundation/ignition/issues/286

packages/core/src/new-api/types/deployment-loader.ts
kanej: It uses whatever path is returned from `deploymentLoader.storeArtifact`.

packages/core/src/new-api/internal/deployment-loader/memory-deployment-loader.ts
kanej: I renamed to `ephemeral-deployment-loader`.

PR #282, opened 2023-06-23
fix: execution state from should always be set (PR 6)

### LINKED ISSUE #281: Fix undefined in from for executionstates
Currently execution states can have `undefined` from values. This is accepted in reconciliation.

Instead the `from` value should always be populated. A future with an `undefined` from should be resolved to `accounts[0]` for both reconciliation and execution.

kanej:
In execution state the `from` address should always be set. If the future specifies undefined, this should be reconcilied as `account[0]` the same way it is during execution.

Fixes #281

### DIFF
--- a/packages/core/src/new-api/internal/execution/execution-engine.ts
+++ b/packages/core/src/new-api/internal/execution/execution-engine.ts
@@ -20,7 +20,10 @@ import {
   NamedLibraryDeploymentFuture,
 } from "../../types/module";
 import { isDeploymentExecutionState } from "../type-guards";
-import { ExecutionEngineState } from "../types/execution-engine";
+import {
+  ExecutionEngineState,
+  ExecutionStrategyContext,
+} from "../types/execution-engine";
 import { ExecutionStateMap, ExecutionStatus } from "../types/execution-state";
 import { getFuturesFromModule } from "../utils/get-futures-from-module";
 import { replaceWithinArg } from "../utils/replace-within-arg";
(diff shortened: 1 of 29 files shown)


alcuadrado:
I think this looses information about the user intention that would be valuable for the execution strategy. 

If the user provided a `from` and wants to use an execution strategy that deploys from a contract (e.g. create2) the execution strategy should fail. Hence, it's important to keep the `undefined` up to that point.

I think the `from` should be resolved right before sending the first transaction (or static call) of a future.

Now, how can be reconcile them then? I guess we can do:
* If the future's from is `undefined`, validate that `accounts[0]` matches the from of any tx/static-call
* If the future's from is not `undefined`, validate that the future's `from` matches the from of any tx/static-call

kanej:
> I think this looses information about the user intention that would be valuable for the execution strategy.
> 
> If the user provided a `from` and wants to use an execution strategy that deploys from a contract (e.g. create2) the execution strategy should fail. Hence, it's important to keep the `undefined` up to that point.
> 
> I think the `from` should be resolved right before sending the first transaction (or static call) of a future.
> 
> Now, how can be reconcile them then? I guess we can do:
> 
>     * If the future's from is `undefined`, validate that `accounts[0]` matches the from of any tx/static-call
> 
>     * If the future's from is not `undefined`, validate that the future's `from` matches the from of any tx/static-call

I have made the following alterations. Execution state `from` can again be `undefined`. The execution strategy now has access to the `accounts` and can decide how to deal with an `undefined` from.

In reconciliation, the execution states from is used, or if it is `undefined` we use the `history` to find the first on-chain request with a `from` and take that as the previous from.

I have included an edge case for if the `from` is undefined but there are no usable on-chain messages. In that case _any_ from value for the future reconciles.

One aspect to this design I don't like is we assume in reconciliation that `undefined` future froms resolve to `account[0]` but don't enforce that in the execution strategy. Maybe swap giving access to `accounts` in the execution strategy with a `defaultFrom`?

alcuadrado (approved):

PR #284, opened 2023-06-23
feat/restart-on-error (PR 8)

kanej:
Support restarting a run after a failed future (i.e. a future that has a transaction that reverts).

A `wipe` cli command has been added to allow reruns on error by clearing
the state against a future:

```sh
npx hardhat wipe --deployment network-31337 --future LockModule:Lock
# LockModule:Lock state has been cleared
```

## TODO

- [x] Support error in journal messages
- [x] Display error messages at cli
- [x] Add cli task to remove future extension state
- [x] Add tests to cover error case

### DIFF
--- a/examples/sample/contracts/Lock.sol
+++ b/examples/sample/contracts/Lock.sol
@@ -5,30 +5,30 @@ pragma solidity ^0.8.9;
 // import "hardhat/console.sol";
 
 contract Lock {
-    uint public unlockTime;
-    address payable public owner;
+  uint public unlockTime;
+  address payable public owner;
 
-    event Withdrawal(uint amount, uint when);
+  event Withdrawal(uint amount, uint when);
 
-    constructor(uint _unlockTime) payable {
-        require(
-            block.timestamp < _unlockTime,
-            "Unlock time should be in the future"
-        );
+  constructor(uint _unlockTime) payable {
+    require(
+      block.timestamp < _unlockTime,
+      "Unlock time should be in the future"
+    );
 
-        unlockTime = _unlockTime;
-        owner = payable(msg.sender);
-    }
+    unlockTime = _unlockTime;
+    owner = payable(msg.sender);
+  }
 
-    function withdraw() public {
-        // Uncomment this line, and the import of "hardhat/console.sol", to print a log in your terminal
-        // console.log("Unlock time is %o and block timestamp is %o", unlockTime, block.timestamp);
+  function withdraw() public {
+    // Uncomment this line, and the import of "hardhat/console.sol", to print a log in your terminal
+    // console.log("Unlock time is %o and block timestamp is %o", unlockTime, block.timestamp);
 
-        require(block.timestamp >= unlockTime, "You can't withdraw yet");
-        require(msg.sender == owner, "You aren't the owner");
+    require(block.timestamp >= unlockTime, "You can't withdraw yet");
+    require(msg.sender == owner, "You aren't the owner");
 
-        emit Withdrawal(address(this).balance, block.timestamp);
+    emit Withdrawal(address(this).balance, block.timestamp);
 
-        owner.transfer(address(this).balance);
-    }
+    owner.transfer(address(this).balance);
+  }
 }
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -3,6 +3,7 @@ export * from "./errors";
 export * from "./initialization";
 export { defineModule } from "./new-api/define-module";
 export { deploy } from "./new-api/deploy";
+export { wipe } from "./new-api/wipe";
 /* TODO: how is module constructor getting exposed? */
 export { ModuleConstructor } from "./new-api/internal/module-builder";
 export { StoredDeploymentSerializer } from "./new-api/stored-deployment-serializer";
--- a/packages/core/src/new-api/internal/execution/execution-engine.ts
+++ b/packages/core/src/new-api/internal/execution/execution-engine.ts
@@ -5,10 +5,12 @@ import { ArtifactResolver } from "../../types/artifact";
 import { DeploymentResult } from "../../types/deployer";
 import { DeploymentLoader } from "../../types/deployment-loader";
 import {
+  ExecutionFailure,
   ExecutionResultMessage,
   FutureRestartMessage,
   FutureStartMessage,
   JournalableMessage,
+  OnchainResultFailureMessage,
 } from "../../types/journal";
 import {
   AccountRuntimeValue,
@@ -19,6 +21,10 @@ import {
   NamedContractDeploymentFuture,
   NamedLibraryDeploymentFuture,
 } from "../../types/module";
+import {
+  isOnChainFailureMessage,
+  isOnChainSuccessMessage,
+} from "../journal/type-guards";
 import {
   isCallExecutionState,
   isDeploymentExecutionState,
@@ -43,9 +49,9 @@ import { resolveModuleParameter } from "../utils/resolve-module-parameter";
 import { executionStateReducer } from "./executionStateReducer";
 import {
   isDeployedContractExecutionSuccess,
-  isExecutionResult,
+  isExecutionFailure,
+  isExecutionResultMessage,
   isOnChainAction,
-  isOnchainResult,
 } from "./guards";
 
 type ExecutionBatch = Future[];
@@ -64,8 +70,15 @@ export class ExecutionEngine {
 
       const batchResult = await this._executeBatch(executionBatch, state);
 
-      if (batchResult.some((b) => b.type === "execution-failure")) {
-        return { status: "failed" };
+      if (batchResult.some(isExecutionFailure)) {
+        return {
+          status: "failure",
+          errors: Object.fromEntries(
+            batchResult
+              .filter(isExecutionFailure)
+              .map((r) => [r.futureId, r.error])
+          ),
+        };
       }
 
       if (batchResult.some((b) => b.type === "execution-hold")) {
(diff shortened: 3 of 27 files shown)


examples/sample/contracts/Lock.sol
```diff
@@ -5,30 +5,30 @@ pragma solidity ^0.8.9;
 // import "hardhat/console.sol";
 
 contract Lock {
-    uint public unlockTime;
-    address payable public owner;
+  uint public unlockTime;
```
kanej, line 8: I suspect I should add prettier format to the examples packages.

packages/core/src/index.ts
```diff
@@ -3,6 +3,7 @@ export * from "./errors";
 export * from "./initialization";
 export { defineModule } from "./new-api/define-module";
 export { deploy } from "./new-api/deploy";
+export { wipe } from "./new-api/wipe";
```
kanej, line 6: This is the new wipe future extension state.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -107,25 +120,39 @@ export class ExecutionEngine {
...
 
       if (isOnChainAction(current)) {
         current = await state.transactionService.onchain(current, {
           libraries,
         });
-      } else if (isOnchainResult(current)) {
+      } else if (isOnChainSuccessMessage(current)) {
         current = (await exectionStrategy.next(current)).value;
+      } else if (isOnChainFailureMessage(current)) {
+        current = this._convertToExecutionFailure(current);
```
kanej, line 133: Failures do not lead to the execution strategy, instead the execution state is updated immediately as a failure.

packages/core/src/new-api/internal/execution/execution-strategy.ts
```diff
@@ -43,17 +44,19 @@ export class BasicExecutionStrategy
...
       subtype: "deploy-contract",
+      futureId: deploymentExecutionState.id,
+      transactionId: 1,
       contractName: deploymentExecutionState.contractName,
       value: deploymentExecutionState.value.toString(),
       args: deploymentExecutionState.constructorArgs,
       // TODO: the possibily of undefined for `from` needs
       // a working resolution
-      from: deploymentExecutionState.from ?? "n/a",
+      from: deploymentExecutionState.from,
```
kanej, line 59: `from` state can no longer be undefined on `deploymentExecutionState`. This is cleanup in passing that was missed in an earlier PR.

packages/core/src/new-api/internal/wiper.ts
```diff
@@ -0,0 +1,52 @@
+import { IgnitionError } from "../../errors";
+import { Journal, WipeMessage } from "../types/journal";
+
+import { executionStateReducer } from "./execution/executionStateReducer";
+import { ExecutionStateMap } from "./types/execution-state";
+
+export class Wiper {
```
kanej, line 7: This is matching the pattern of an internal Deployer and a public `deploy` function.

packages/core/test/new-api/call.ts
```diff
@@ -558,7 +558,7 @@ describe("call", () => {
       await assert.isRejected(
         validateNamedContractCall(
           future as any,
-          setupMockArtifactResolver({} as any)
+          setupMockArtifactResolver({ Another: {} as any })
```
kanej, line 561: This is a side-effect of making `setupMockArtifactResolver` more flexible.

It now will return a generic artifact, unless your set it up with a map of contract names to artifacts.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -111,25 +124,39 @@ export class ExecutionEngine {
...
       } else {
         current = (await exectionStrategy.next(null)).value;
       }
 
+      // TODO: add never type check?
```
alcuadrado, line 142: I guess there's no other way to do this, because a switch wouldn't work.

alcuadrado: I do think that we need some sort of check, because in the latest else `current` can have quiet a few types, some of them unrelated, like `WipeMessage`.

I guess the least we can do is avoid using plain `else` (without if) in these chains of `if`s

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
```diff
@@ -2,11 +2,19 @@ import { ethers } from "ethers";
...
-interface TransactionReceipt {
+interface TransactionSuccess {
+  type: "transaction-success";
   contractAddress?: string;
   txId?: string;
 }
 
+interface TransactionFailure {
+  type: "transaction-failure";
+  error: Error;
```
alcuadrado, line 13: I don't think we should record `Error`s like this, but rather the most low-level representation of the failure that we can get.

In the ase of a transaction we don't get much, as you don't get access to the `returndata` of a failed transaction. We may want to simulate transactions before sending them, and record the return values of those simulations instead.

For static calls this does mean recording the raw data returned by the node we are connected to. It may be simpler to access it by doing the `eth_call` ourselves, and only using ethers for ABI encoding.

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
```diff
@@ -2,11 +2,19 @@ import { ethers } from "ethers";
...
+}
+
+type TransactionReceipt = TransactionSuccess | TransactionFailure;
```
alcuadrado, line 16: I think this type is really confusing. `TransactionReceipt` as receipt is already used in the protocol.

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
```diff
@@ -48,16 +56,25 @@ export class EthersChainDispatcher implements ChainDispatcher {
...
       tx.gasPrice = gasPrice;
     }
+    try {
+      const response = await signer.sendTransaction(tx);
 
-    const response = await signer.sendTransaction(tx);
+      const txHash = response.hash;
 
-    const txHash = response.hash;
+      const receipt = await this._transactionProvider.wait(txHash);
```
alcuadrado, line 64: Just to confirm: we won't use this behavior (i.e. `.wait`), right?

packages/core/src/new-api/types/deployer.ts
```diff
@@ -22,6 +23,17 @@ export interface DeploymentResultSuccess {
...
 
+/**
+ * The result of a failed deployment run (at least one future had an
+ * failed transaction).
+ *
+ * @beta
+ */
+export interface DeploymentResultFailure {
+  status: "failure";
+  errors: { [key: string]: Error };
```
alcuadrado, line 34: In a similar fashion than https://github.com/NomicFoundation/ignition/pull/284/files#r1244391873, I don't think Error is the right abstraction here. We should have a more precise type, possibly a union of different possible errors.

packages/core/src/new-api/types/journal.ts
```diff
@@ -324,6 +358,8 @@ export interface StaticCallExecutionSuccess {
  */
 export interface ExecutionFailure {
   type: "execution-failure";
+  futureId: string;
+  error: Error;
```
alcuadrado, line 362: Same re errors

packages/core/src/new-api/types/journal.ts
```diff
@@ -49,6 +52,8 @@ export type OnchainInteractionMessage =
 export interface DeployContractInteractionMessage {
   type: "onchain-action";
   subtype: "deploy-contract";
+  futureId: string;
+  transactionId: number;
```
zoeyTM, line 56: What is this `transactionId`? If this is meant to be like an internal nonce, I really think we should come up with another name that doesn't conflict with existing Ethereum concepts.

kanej: Agreed it is a bad name.

It is trying to give an id to a desired on-chain interaction for a future. So the first message yielded in the execution strategy would be 1, and the second 2. This lets us know when we are rerunning the same on-chain interaction.

kanej: Maybe `OnchainId`? Do you have another suggestion?

zoeyTM: `executionId` or `messageId` maybe?

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: I am going to add an assertNever case to typecheck that all message types are covered, in a later PR.

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
kanej: This has been captured as an issue here: https://github.com/NomicFoundation/ignition/issues/286

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
kanej: This will all be restructured in the monitoring PR.

packages/core/src/new-api/internal/execution/transactions/chain-dispatcher.ts
kanej: This is a terrible name. But that code is getting folded (deleted) into the execution-engine.

packages/core/src/new-api/types/journal.ts
kanej: Alright, I will go with `executionId`, I will push it in a later PR to reduce rebase pain.

alcuadrado (approved):

PR #291, opened 2023-07-03
Enhance our reconciliation phase with reconciliation rules based on artifacts (PR 12)

### LINKED ISSUE #269: Add changes to artifacts reconciliation
Enhance our reconciliation phase with reconciliation rules based on artifacts.

Specifically we want to reconcile whether the bytecode of a contract has changed between the previous run and this coming run.

zoeyTM:
fixes #269


packages/core/src/new-api/internal/deployer.ts
```diff
@@ -62,11 +74,28 @@ export class Deployer {
...
+    );
+
+    // realistically this should be impossible to fail.
+    // just need it here for the type inference
+    assertIgnitionInvariant(
+      isContractExecutionStateArray(contractStates),
+      "Invalid state map"
+    );
+
+    const moduleArtifactMap = await this._loadModuleArtifactMap(contracts);
```
alcuadrado, line 89: Add a comment explaining that this was done like this because the reconciler is sync. If it were async we could pass it the loaders and it could load them JIT.

packages/core/src/new-api/types/artifact.ts
```diff
@@ -13,6 +13,16 @@ export interface Artifact {
   >;
 }
 
+/**
+ * A map of (contract) future id's to their respective artifact.
+ * Used in reconciliation.
+ *
+ * @beta
+ */
+export interface ArtifactMap {
```
alcuadrado, line 22: We can make this internal

alcuadrado:
I think this looks good. I would just add a few simple tests.

packages/core/src/new-api/internal/reconciliation/utils.ts
```diff
@@ -54,13 +56,33 @@ export function addressToErrorString(potential: string | undefined) {
...
   return bytecode.slice(-METADATA_LENGTH).readUInt16BE(0) + METADATA_LENGTH;
 }
 
+function isValidMetadata(data: Buffer): boolean {
+  const { decode } = require("cbor") as typeof CborT;
+  try {
+    decode(data);
+    return true;
+  } catch (e) {
+    console.log(e);
```
kanej, line 69:
```suggestion
```

Clean up a debug `console.log`

kanej (approved):

PR #292, opened 2023-07-03
execution: switch to cross batch based execution (PR A)

kanej:
Add monitoring of new transactions and allow the connection of monitoring to newly restarted futures.

## TODO

- [x] Restructure inner loop of execution engine
- [x] Bring back execution for each subtype
- [ ] ~Wire in new block tick~
- [ ] ~Nonce initialization check~
- [ ] ~Monitor check for in-flight transactions~


packages/core/src/new-api/internal/execution/chain-dispatcher.ts
```diff
@@ -0,0 +1,160 @@
...
+export class ChainDispatcherImpl implements ChainDispatcher {
+  private _accountsState: AccountsState;
+
+  constructor(private _adapters: Adapters) {
+    this._accountsState = {};
+  }
+
+  // TODO: should nonce management be separated out into its own
+  // class?
+  public async allocateNextNonceForAccount(address: string): Promise<number> {
```
kanej, line 19: This may be complecting nonce management with the ethers chain interactions, we should consider moving it out.

In general the nonce initialization check still needs done.

packages/core/src/new-api/internal/execution/chain-dispatcher.ts
```diff
@@ -0,0 +1,160 @@
...
+    const eventNameTopic = filter.topics?.[0];
+
+    assertIgnitionInvariant(eventNameTopic !== undefined, "Unknown event name");
+
+    const receipt = await this._adapters.transactions.getTransactionReceipt(
+      txToReadFrom
+    );
+
+    // TODO: should this really return an error result
+    assertIgnitionInvariant(
```
kanej, line 101: Is this one instance of the transaction dropped case?

packages/core/src/new-api/internal/execution/chain-dispatcher.ts
```diff
@@ -0,0 +1,160 @@
...
+    functionName: string,
+    args: ArgumentType[],
+    from: string
+  ): Promise<any> {
+    const signer: ethers.Signer = await this._adapters.signer.getSigner(from);
+
+    const contractInstance = new Contract(contractAddress, abi, signer);
+
+    const result = await contractInstance[functionName](...args, {
+      from: await signer.getAddress(),
```
kanej, line 145: Actually can this just be the previous `from`?

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -105,73 +131,487 @@ export class ExecutionEngine {
...
+
+      return {
+        type: "onchain-result",
+        subtype: "read-event-arg-success",
+        futureId: request.futureId,
+        executionId: request.executionId,
+        result,
+      };
+    } catch (error) {
+      return {
```
kanej, line 561: I am not sure about having the transaction handling here. We are letting ethers knowledge escape from the chain dispatcher.

Similarly we need to refine the error capture to be a more specific error value and code that Ignition know.

packages/core/src/new-api/internal/execution/executionStateReducer.ts
```diff
@@ -120,6 +122,18 @@ export function executionStateReducer(
     );
   }
 
+  if (isExecutionFailure(action)) {
```
kanej, line 125: Actually update the execution state on failure. The change to check whether a batch is complete by looking at execution state exposed that this was missing.

packages/core/src/new-api/internal/execution/chain-dispatcher.ts
kanej: We need to resolve nonces from history when we are restarting.

This might be the place to detect user interference with the nonce.

kanej: It needs to get the pending transaction count.

packages/core/src/new-api/internal/execution/chain-dispatcher.ts
kanej: We are happy with assertion here, this should never be called on an unconfirmed transaction.

packages/core/src/new-api/types/journal.ts
```diff
@@ -150,6 +151,25 @@ export interface ContractAtInteractionMessage {
...
+ * @beta
+ */
+export interface OnchainTransaction {
+  type: "onchain-transaction";
+  futureId: string;
+  executionId: number;
+  txHash: string;
+  from: string;
+  nonce: number;
+  tx: any;
```
kanej, line 168: This needs tightened up as it is a sketch.

We should define tx.

kanej: Pato: we definitely need the args.

We do want to record the raw tx.

packages/core/test/new-api/execution/execution-engine.ts
```diff
@@ -248,15 +247,13 @@ describe("execution engine", () => {
...
-            executionId: 1,
+        [accounts[2]]: {
+          0: {
+            blockNumber: 0,
+            confirmations: 1,
             contractAddress: exampleAddress,
-            txId,
-          },
+            transactionHash: txId,
+          } as any,
```
kanej, line 256: How do we get rid of any here?

alcuadrado (approved):
I think it'd make sense to do an in-depth review one or two prs later, so I'm approving this

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: This will cleaned up with error handling.

packages/core/src/new-api/internal/execution/chain-dispatcher.ts
kanej: We used from directly.

PR #295, opened 2023-07-05
Add validation for module params (PR 2)

### LINKED ISSUE #277: validation: add getParameter check
As part of the valiation phase we should report a validation error if there is no parameter provided for the `getParameter` call.

zoeyTM:
fixes #277

### DIFF
--- a/packages/core/src/new-api/internal/deployer.ts
+++ b/packages/core/src/new-api/internal/deployer.ts
@@ -56,7 +56,11 @@ export class Deployer {
   ): Promise<DeploymentResult> {
     const module = this._moduleConstructor.construct(moduleDefinition);
 
-    await validate(module, this._artifactResolver);
+    await validate(
+      module,
+      this._artifactResolver,
+      deploymentParameters[module.id]
+    );
 
     const previousStateMap = await this._loadExecutionStateFrom(
       this._deploymentLoader.journal
--- a/packages/core/src/new-api/internal/deployer.ts
+++ b/packages/core/src/new-api/internal/deployer.ts
@@ -56,11 +56,7 @@ export class Deployer {
   ): Promise<DeploymentResult> {
     const module = this._moduleConstructor.construct(moduleDefinition);
 
-    await validate(
-      module,
-      this._artifactResolver,
-      deploymentParameters[module.id]
-    );
+    await validate(module, this._artifactResolver, deploymentParameters);
 
     const previousStateMap = await this._loadExecutionStateFrom(
       this._deploymentLoader.journal
--- a/packages/core/src/new-api/internal/utils/retrieve-nested-runtime-values.ts
+++ b/packages/core/src/new-api/internal/utils/retrieve-nested-runtime-values.ts
@@ -1,0 +1,41 @@
+import {
+  isFuture,
+  isModuleParameterRuntimeValue,
+  isRuntimeValue,
+} from "../../type-guards";
+import {
+  ArgumentType,
+  ModuleParameterRuntimeValue,
+  RuntimeValueType,
+} from "../../types/module";
+
+export function retrieveNestedRuntimeValues(
+  args: ArgumentType[]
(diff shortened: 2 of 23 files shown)


packages/core/src/new-api/internal/deployer.ts
```diff
@@ -56,7 +56,11 @@ export class Deployer {
   ): Promise<DeploymentResult> {
     const module = this._moduleConstructor.construct(moduleDefinition);
 
-    await validate(module, this._artifactResolver);
+    await validate(
+      module,
+      this._artifactResolver,
+      deploymentParameters[module.id]
```
alcuadrado, line 62: Wouldn't passing just this module's params limit the ability to validate submodules?

packages/core/src/new-api/internal/validation/futures/validateArtifactContractDeployment.ts
```diff
@@ -1,13 +1,34 @@
...
+  ModuleParameters,
+} from "../../../types/module";
 
 export async function validateArtifactContractDeployment(
   future: ArtifactContractDeploymentFuture,
-  artifactLoader: ArtifactResolver
+  artifactLoader: ArtifactResolver,
+  moduleParameters: ModuleParameters
 ) {
+  const moduleParams = future.constructorArgs.filter(
```
alcuadrado, line 16: I think this wouldn't validate something like:

```
m.contract("A", artifact, [[m.getParam("p")]])
```

because the param is nested in an array.

I think we'd need to do something similar to what @kanej made when replacing complex args with other representations.

alcuadrado: This comment is also valid for the rest of the futures that take constructor or call arguments.

packages/core/src/new-api/internal/deployer.ts
alcuadrado: Oh I just noticed that we aren't validating submodules at all.

Should this be address in this PR or a new one? /cc @kanej

packages/core/src/new-api/internal/validation/futures/validateArtifactContractDeployment.ts
kanej: Here is the `replace-within-arg` code that has to do a recursive breakdown of the argument list @zoeyTM:

https://github.com/NomicFoundation/ignition/blob/1d20a732a93bcc03b2260b78fe76b613ea7bcd6e/packages/core/src/new-api/internal/utils/replace-within-arg.ts#L48

zoeyTM: fixed in bac1d82

packages/core/src/new-api/internal/deployer.ts
zoeyTM: fixed in 852f58f

kanej (approved):

packages/core/src/new-api/internal/utils/retrieve-nested-runtime-values.ts
```diff
@@ -0,0 +1,41 @@
...
+
+function checkForValues(
+  arg: ArgumentType
+):
+  | Array<ModuleParameterRuntimeValue<any> | null>
+  | ModuleParameterRuntimeValue<any>
+  | null {
+  if (isRuntimeValue(arg)) {
+    if (arg.type === RuntimeValueType.ACCOUNT) {
+      return null;
```
alcuadrado, line 26: Doesn't this mean that we aren't validating that the account's index exist? We should validate params and account runtime values, as both can be missing for different reasons.

zoeyTM: This is for validating module parameters, and account runtime values can't be passed in as a module parameter afaik. This specific line is just type narrowing.

PR #298, opened 2023-07-10
execution: monitoring inflight transactions (PR B)

kanej:
Support pending transactions, both between block loops and on restarts.

## TODO

- [x] Switch to state machine for tracking on-chain interactions
- [x] add tests for failed request cases
- [x] merge onchain state reducer into execution reducer
- [x] remove duplication in state machine for deploy/call-function/send-data

### DIFF
--- a/packages/core/src/new-api/internal/execution/execution-state-reducer.ts
+++ b/packages/core/src/new-api/internal/execution/execution-state-reducer.ts
@@ -7,6 +7,7 @@ import {
   isReadEventArgumentStartMessage,
   isSendDataStartMessage,
   isStaticCallStartMessage,
+  isTransactionMessage,
 } from "../journal/type-guards";
 import {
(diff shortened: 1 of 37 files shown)


packages/core/src/new-api/internal/execution/execution-state-reducer.ts
```diff
@@ -145,6 +149,7 @@ export function executionStateReducer(
     const updateWithOnchainAction: ExecutionState = {
       ...previousExState,
       history: [...previousExState.history, action],
+      onchain: onchainActionReducer(previousExState.onchain, action),
```
kanej, line 152: We build up per onchain action state under this property - open to new name suggestions!

alcuadrado (approved):
Discussed this on a call and decided to address any issue in future PRs.

PR #303, opened 2023-07-11
execution/new-block-tick (PR D)

kanej:


packages/core/src/new-api/internal/deployer.ts
```diff
@@ -113,6 +115,10 @@ export class Deployer {
     const batches = Batcher.batch(module, previousStateMap);
 
     return this._executionEngine.execute({
+      block: { number: -1, hash: "-" },
```
alcuadrado, line 118: I commented this on PR B, but I think everything that's passed here should be the internal state of the ExecutionEngine, and it should receive the necessary params instead.

packages/core/src/new-api/internal/deployer.ts
```diff
@@ -30,6 +30,8 @@ import { assertIgnitionInvariant } from "./utils/assertions";
 import { getFuturesFromModule } from "./utils/get-futures-from-module";
 import { validate } from "./validation/validate";
 
+const BLOCK_POLLING_INTERVAL = 200;
```
alcuadrado, line 33: Should be received by `deploy`, which means it'd be configurable by the plugin.

alcuadrado (approved):

packages/core/src/new-api/internal/deployer.ts
kanej: Agreed, I have recorded a todo for this: https://github.com/NomicFoundation/ignition/issues/311

packages/core/src/new-api/internal/deployer.ts
kanej: I have recorded a todo for this. It has been partially done in the timeouts PR: https://github.com/NomicFoundation/ignition/pull/306

But I have added an overall todo to check that config is being passed through the entire stack and comes from the plugin: https://github.com/NomicFoundation/ignition/issues/312

PR #304, opened 2023-07-11
execution/nonce-check (PR E)

### LINKED ISSUE #300: Nonce initialization check
With cross-batch processing nonce management is important. We want to ensure at the start of an execution run that the currently in-flight nonce usage matches up with the chain. During execution we want to ensure that whenever a transaction receipt is processed that it matches the expected transaction accorind to execution state.

## TODO

- [x] Add initialization nonce check on first batch
- [x] Add nonce check on all transaction receipts

kanej:
Any pending transactions on accounts that will be used should have no pending transactions.

## TODO

- [x] Fail if accounts have pending transactions (not through ignition)
- [x] Support pending transactions that have been recorded in Ignition
- [x] Support pending ignition transaction that has completed between runs
- [x] Support pending ignition transaction falling out of mempool
  - [x] Confirmed transaction from user with same nonce
  - [x] Pending transaction from user with same nonce
  - [x] transaction dropped without user interference
- [x] Nonce check during batch processing

Resolves #300


packages/core/src/new-api/type-guards.ts
```diff
@@ -179,6 +204,31 @@ export function isDeploymentFuture(
...
 
+/**
+ * Returns true if the future requires submitting a transaction on-chain
+ *
+ * @beta
+ */
+export function isFutureThatSubmitsOnchainTransaction(
+  f: Future
+): f is Exclude<
+  Exclude<
```
alcuadrado, line 215: 😱

packages/core/test/new-api/helpers.ts
```diff
@@ -399,9 +400,39 @@ export class MockChainDispatcher implements ChainDispatcher {
     return response as any;
   }
 
+  public async getTransaction(
+    txHash: string
+  ): Promise<ethers.providers.TransactionResponse | null | undefined> {
+    const [from, nonce] = txHash.split("--");
```
alcuadrado, line 406: What's this? I don't get it.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
```diff
@@ -619,3 +587,43 @@ async function _readEventArg(
...
+  successConstructor: (
+    receipt: ethers.providers.TransactionReceipt
+  ) => ExecutionSuccess | ExecutionFailure | TransactionMessage
+): Promise<OnchainStateTransitionContinue | OnchainStateTransitionPause> {
+  const currentTransaction = await state.chainDispatcher.getTransaction(
+    next.txHash
+  );
+
+  if (currentTransaction === null || currentTransaction === undefined) {
+    throw new IgnitionError(
```
alcuadrado, line 603: I don't think this is the intended behavior. We should be trying to discern if this is due to user interference or just dropped, and report different errors.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
```diff
@@ -619,3 +587,43 @@ async function _readEventArg(
...
+  }
+
+  const receipt = await state.chainDispatcher.getTransactionReceipt(
+    next.txHash
+  );
+
+  if (receipt === null || receipt === undefined) {
+    // No receipt means it hasn't been included in a block yet
+    return { status: "pause" };
+  }
```
alcuadrado, line 615: This information is already in the `currentTransaction` object, so the `getTransactionReceipt` call can be saved.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -105,6 +115,130 @@ export class ExecutionEngine {
...
+      const maxNonceTran = maxBy(inflight, "nonce");
+
+      assertIgnitionInvariant(
+        maxNonceTran !== undefined,
+        "Always at least one entry going into maxBy"
+      );
+
+      // The pending transactions do not match the expected list
+      // given the inflight transactions from the previous run
+      if (maxNonceTran.nonce + 1 < pending) {
```
alcuadrado, line 183: I don't understand this condition. Can we name them? maybe

`const morePendingTransactionsThanWeExpected = maxNonceTran.nonce < pending`

Is that what we are trying to check?

I think we can do better in this case, but let's leave it as it is for now.

alcuadrado:
I haven't looked at the tests, but LGTM

packages/core/test/new-api/helpers.ts
kanej: The `sendTx` above is generating fake hashes based on:
```
const hash = `${from}--${tx.nonce?.toString() ?? "no-nonce"}`
```

This is pulling the from/nonce from this as a way of smuggly information within the mock.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
kanej: Is it possible to get the contractAddress from a deploy transaction object? It is only needed for deploys rather than calls/sends, so we could split out two different versions even if we can't remove it for deploy.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
kanej: I have added an issue for this one here: https://github.com/NomicFoundation/ignition/issues/313

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: I have pulled out the condition as a private function so we can name it: `_isMorePendingThanPreviousRunIndicates`

PR #305, opened 2023-07-12
Support `ModuleParameterRuntimeValue` for `value` options (PR 2)

### LINKED ISSUE #285: Support `getParameter` runtime values for `value`
Raised here: https://github.com/NomicFoundation/ignition/pull/261#discussion_r1244220266

When passing a value into `m.contract` (and any other action that accepts value) the value should be specifiable as a module parameter runtime value.

zoeyTM:
fixes #285

### DIFF
--- a/packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractDeployment.ts
+++ b/packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractDeployment.ts
@@ -3,6 +3,7 @@ import { isEqual } from "lodash";
 import { ArtifactContractDeploymentFuture } from "../../../types/module";
 import { DeploymentExecutionState } from "../../types/execution-state";
 import { resolveFromAddress } from "../../utils/resolve-from-address";
+import { resolveModuleParameter } from "../../utils/resolve-module-parameter";
 import { ExecutionStateResolver } from "../execution-state-resolver";
 import { ReconciliationContext, ReconciliationFutureResult } from "../types";
 import {
@@ -54,12 +55,15 @@ export function reconcileArtifactContractDeployment(
     return fail(future, "Libraries have been changed");
   }
 
-  if (!isEqual(future.value, executionState.value)) {
+  const resolvedValue =
+    typeof future.value === "bigint"
+      ? future.value
+      : (resolveModuleParameter(future.value, context) as bigint);
+
+  if (!isEqual(resolvedValue, executionState.value)) {
     return fail(
       future,
-      `Value has been changed from ${executionState.value} to ${
-        typeof future.value === "bigint" ? future.value : "a module parameter"
-      }`
+      `Value has been changed from ${executionState.value} to ${resolvedValue}`
     );
   }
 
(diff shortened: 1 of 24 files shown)


packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractDeployment.ts
```diff
@@ -39,7 +39,9 @@ export function reconcileArtifactContractDeployment(
   if (!isEqual(future.value, executionState.value)) {
     return fail(
       future,
-      `Value has been changed from ${executionState.value} to ${future.value}`
+      `Value has been changed from ${executionState.value} to ${
```
kanej, line 42: Should this compare the previous value from execution state with the proposed resolve parameter value?

packages/core/src/new-api/types/serialized-deployment.ts
```diff
@@ -79,7 +79,7 @@ export interface SerializedNamedContractDeploymentFuture
   contractName: string;
   constructorArgs: SerializedArgumentType[];
   libraries: SerializedLibraries;
-  value: SerializedBigInt;
+  value: SerializedBigInt | SerializedModuleParameterRuntimeValue;
```
kanej, line 82: Does the change in types imply a change in the serializer?

If we added a test for serialization of modules with Runtime values for `value` would it serialize/deserialize?

kanej (commented):
This is looking good. I left two comments about reconciliation and serialization.

packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractDeployment.ts
zoeyTM: fixed

packages/core/src/new-api/types/serialized-deployment.ts
zoeyTM: test added and it does appear to serialize/deserialize properly

packages/core/src/new-api/internal/reconciliation/futures/reconcileArtifactContractDeployment.ts
```diff
@@ -36,12 +37,15 @@ export function reconcileArtifactContractDeployment(
     return fail(future, "Libraries have been changed");
   }
 
-  if (!isEqual(future.value, executionState.value)) {
+  const resolvedValue =
+    typeof future.value === "bigint"
+      ? future.value
+      : (resolveModuleParameter(future.value, context) as bigint);
```
kanej, line 43: Can we pull out this entire section as a function and reuse it for each.

That is so that we can add a check on the return result if it is a module parameter. We want to check that the value is a bigint and fail there.

alcuadrado:
Looks good, but it's missing validations. It should validate those values like #295

packages/core/src/new-api/internal/validation/futures/validateSendData.ts
```diff
@@ -9,13 +9,33 @@ export async function validateSendData(
...
+      moduleParameters[future.value.name] ?? future.value.defaultValue;
+    if (param === undefined) {
+      throw new IgnitionValidationError(
+        `Module parameter '${future.value.name}' requires a value but was given none`
+      );
+    } else if (typeof param !== "bigint") {
+      throw new IgnitionValidationError(
+        `Module parameter '${
+          future.value.name
+        }' must be of type 'bigint' but is '${typeof param}'`
```
kanej, line 37: I like the typeof here.

PR #306, opened 2023-07-17
execution/timeouts (PR F)

### LINKED ISSUE #293: Allow restarting on timeouts
Support timeouts in the execution engine. And allow restarting in cases where there has been a timeout.

## TODO

- [x] Set a future as timed out on one of its transactions not completing in time
- [x] Ensure correct response in test/cli
- [x] Support restart of timed out deployment

kanej:
The timeout is checked on lookup of a pending transaction. If the configured timeout is passed then the future is moved to the `timeout` state via a message.

Resolves #293.

### DIFF
--- a/packages/core/src/new-api/deploy.ts
+++ b/packages/core/src/new-api/deploy.ts
@@ -4,7 +4,7 @@ import { FileDeploymentLoader } from "./internal/deployment-loader/file-deployme
 import { ChainDispatcherImpl } from "./internal/execution/chain-dispatcher";
 import { Adapters } from "./types/adapters";
 import { ArtifactResolver } from "./types/artifact";
-import { DeploymentResult } from "./types/deployer";
+import { DeployConfig, DeploymentResult } from "./types/deployer";
 import { IgnitionModuleResult, ModuleParameters } from "./types/module";
 import { IgnitionModuleDefinition } from "./types/module-builder";
 
@@ -14,6 +14,7 @@ import { IgnitionModuleDefinition } from "./types/module-builder";
  * @beta
  */
 export async function deploy({
+  config,
   artifactResolver,
   adapters,
   deploymentDir,
@@ -22,6 +23,7 @@ export async function deploy({
   accounts,
   verbose,
 }: {
+  config?: Partial<DeployConfig>;
   artifactResolver: ArtifactResolver;
   adapters: Adapters;
   deploymentDir?: string;
@@ -42,6 +44,7 @@ export async function deploy({
   const chainDispatcher = new ChainDispatcherImpl(adapters);
 
   const deployer = new Deployer({
+    config: config ?? {},
     artifactResolver,
     deploymentLoader,
     chainDispatcher,
--- a/packages/core/src/new-api/internal/deployer.ts
+++ b/packages/core/src/new-api/internal/deployer.ts
@@ -7,7 +7,7 @@ import type { IgnitionModuleDefinition } from "../types/module-builder";
 
 import { isContractFuture } from "../type-guards";
 import { Artifact, ArtifactResolver } from "../types/artifact";
-import { DeploymentResult } from "../types/deployer";
+import { DeployConfig, DeploymentResult } from "../types/deployer";
 import { DeploymentLoader } from "../types/deployment-loader";
 import { Journal } from "../types/journal";
 
@@ -15,6 +15,7 @@ import { Batcher } from "./batcher";
 import { ExecutionEngine } from "./execution/execution-engine";
 import { executionStateReducer } from "./execution/execution-state-reducer";
 import { BasicExecutionStrategy } from "./execution/execution-strategy";
+import { TranactionLookupTimerImpl } from "./execution/transaction-lookup-timer";
 import { ModuleConstructor } from "./module-builder";
 import { Reconciler } from "./reconciliation/reconciler";
 import { ArtifactMap } from "./reconciliation/types";
@@ -26,11 +27,17 @@ import {
   DeploymentExecutionState,
   ExecutionStateMap,
 } from "./types/execution-state";
+import { TransactionLookupTimer } from "./types/transaction-timer";
 import { assertIgnitionInvariant } from "./utils/assertions";
 import { getFuturesFromModule } from "./utils/get-futures-from-module";
 import { validate } from "./validation/validate";
 
-const BLOCK_POLLING_INTERVAL = 200;
+// The interval between checks to see if a new block has been created
+const DEFAULT_BLOCK_POLLING_INTERVAL = 200;
+
+// In milliseconds the amount of time to wait on a transaction to
+// confirm before timing out
+const DEFAULT_TRANSACTION_TIMEOUT_INTERVAL = 3 * 60 * 1000;
 
 /**
  * Run an Igntition deployment.
@@ -38,18 +45,27 @@ const BLOCK_POLLING_INTERVAL = 200;
  * @beta
  */
 export class Deployer {
+  private _config: DeployConfig;
   private _moduleConstructor: ModuleConstructor;
   private _executionEngine: ExecutionEngine;
   private _strategy: ExecutionStrategy;
   private _artifactResolver: ArtifactResolver;
   private _deploymentLoader: DeploymentLoader;
   private _chainDispatcher: ChainDispatcher;
+  private _transactionLookupTimer: TransactionLookupTimer;
 
   constructor(options: {
+    config?: Partial<DeployConfig>;
     artifactResolver: ArtifactResolver;
     deploymentLoader: DeploymentLoader;
     chainDispatcher: ChainDispatcher;
   }) {
+    this._config = {
+      blockPollingInterval: DEFAULT_BLOCK_POLLING_INTERVAL,
+      transactionTimeoutInterval: DEFAULT_TRANSACTION_TIMEOUT_INTERVAL,
+      ...options.config,
+    };
+
     this._strategy = new BasicExecutionStrategy();
     this._artifactResolver = options.artifactResolver;
     this._deploymentLoader = options.deploymentLoader;
@@ -58,6 +74,9 @@ export class Deployer {
 
     this._moduleConstructor = new ModuleConstructor();
     this._executionEngine = new ExecutionEngine();
+    this._transactionLookupTimer = new TranactionLookupTimerImpl(
+      this._config.transactionTimeoutInterval
+    );
   }
 
   public async deploy(
@@ -115,10 +134,8 @@ export class Deployer {
     const batches = Batcher.batch(module, previousStateMap);
 
     return this._executionEngine.execute({
+      config: this._config,
       block: { number: -1, hash: "-" },
-      config: {
-        blockPollingInterval: BLOCK_POLLING_INTERVAL,
-      },
       strategy: this._strategy,
       artifactResolver: this._artifactResolver,
       batches,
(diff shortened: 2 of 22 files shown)


packages/core/src/new-api/deploy.ts
```diff
@@ -14,13 +14,15 @@ import { IgnitionModuleDefinition } from "./types/module-builder";
...
 export async function deploy({
+  config,
   artifactResolver,
   adapters,
   deploymentDir,
   moduleDefinition,
   deploymentParameters,
   accounts,
 }: {
+  config?: Partial<DeployConfig>;
```
kanej, line 25: This is the test helper function, I wanted the option to change config setting (specifically the timeout for transactions)

packages/core/src/new-api/internal/deployer.ts
```diff
@@ -26,30 +27,45 @@ import {
...
 import { getFuturesFromModule } from "./utils/get-futures-from-module";
 import { validate } from "./validation/validate";
 
-const BLOCK_POLLING_INTERVAL = 200;
+// The interval between checks to see if a new block has been created
+const DEFAULT_BLOCK_POLLING_INTERVAL = 200;
+
+// In milliseconds the amount of time to wait on a transaction to
+// confirm before timing out
+const DEFAULT_TRANSACTION_TIMEOUT_INTERVAL = 3 * 60 * 1000;
```
kanej, line 40: Switch from these being the hardcoded values to being the hard coded defaults. I went with 3 mins for default transaction timeout.

packages/core/src/new-api/internal/deployer.ts
```diff
@@ -26,30 +27,45 @@ import {
...
  */
 export class Deployer {
+  private _config: DeployConfig;
   private _moduleConstructor: ModuleConstructor;
   private _executionEngine: ExecutionEngine;
   private _strategy: ExecutionStrategy;
   private _artifactResolver: ArtifactResolver;
   private _deploymentLoader: DeploymentLoader;
   private _chainDispatcher: ChainDispatcher;
+  private _transactionLookupTimer: TransactionLookupTimer;
```
kanej, line 55: This _class_ (God forgive me) encapsulates setting times against transaction lookup starts so that timeouts can be determined.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -80,6 +85,13 @@ export class ExecutionEngine {
 
     await this._checkNonces(state, futures);
 
+    // Record start of the run (which resets timeout to started)
+    const startRun: StartRunMessage = {
+      type: "run-start",
+    };
+
+    await this._apply(state, startRun);
```
kanej, line 93: After the nonces check we should be good to start an execution. At that point we record a "run-start" message, which resets `TIMEOUT` futures to `STARTED`. The execution engine will jump those futures to same checks as before the timeout on the previous run.

NOTE: we should probably enhance the start run message to include chainId, to better support a chainId check in reconciliation.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -99,11 +116,22 @@ export class ExecutionEngine {
         };
       }
 
-      if (batchResult.some((b) => b.type === "execution-hold")) {
+      if (batchResult.some(isExecutionTimeout)) {
+        return {
+          status: "timeout",
```
kanej, line 121: We need a new `timeout` execution result as that is a valid end result of a deploy now.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -99,11 +116,22 @@ export class ExecutionEngine {
...
+          timeouts: batchResult.filter(isExecutionTimeout).map((r) => ({
+            futureId: r.futureId,
+            executionId: r.executionId,
+            txHash: r.txHash,
```
kanej, line 125: `executionid` and `txHash` are added for display in the test exception/cli.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -262,18 +290,23 @@ export class ExecutionEngine {
...
 
       batchResults = [...batchResults, ...results];
 
-      if (this._isBatchComplete(sortedFutures, state)) {
+      if (this._isBatchComplete(state, sortedFutures)) {
         break;
       }
 
-      await this._newBlock(state);
+      const timeoutResults = await this._newBlockOrTimeout(
```
kanej, line 304: This needed enhanced. If the timeout comes between blocks, we need check for timeouts in this wait period.

This also implies it must return timeout result messages so they can be gathered together for the batch.

kanej: A refactor we should consider is not passing result messages around here, and calculate the final result based on the execution state. Actually let me add a todo.

packages/core/src/new-api/internal/execution/execution-state-reducer.ts
```diff
@@ -23,21 +26,30 @@ import {
...
+  isExecutionTimeout,
+} from "./guards";
 import { onchainActionReducer } from "./onchain-action-reducer";
 
 export function executionStateReducer(
   executionStateMap: ExecutionStateMap,
   action: JournalableMessage
 ): ExecutionStateMap {
+  if (isStartRunMessage(action)) {
+    return setTimeoutFuturesToStarted(executionStateMap);
```
kanej, line 42: On a start run message we update the state of all timed out futures to be `STARTED` instead. Let us process them in the batch.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
```diff
@@ -595,6 +600,12 @@ async function checkTransactionComplete(
     receipt: ethers.providers.TransactionReceipt
   ) => ExecutionSuccess | ExecutionFailure | TransactionMessage
 ): Promise<OnchainStateTransitionContinue | OnchainStateTransitionPause> {
+  state.transactionLookupTimer.registerStartTime({
```
kanej, line 603: Every time we check for a transaction by receipt we register a start time (this is idempotent, on the second time we ignore) with the `transactionLookupTimer` which encapsulates this ephemeral state.

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
```diff
@@ -609,8 +620,19 @@ async function checkTransactionComplete(
     next.txHash
   );
 
+  // No receipt means it hasn't been included in a block yet
   if (receipt === null || receipt === undefined) {
-    // No receipt means it hasn't been included in a block yet
+    if (state.transactionLookupTimer.isTimedOut(next.txHash)) {
```
kanej, line 625: If we get a receipt, we never timeout - even if it hasn't reached the number of confirmations yet.

If the timeout is passed, apply a timeout.

packages/hardhat-plugin/test/execution/timeouts/deploy-run-times-out.ts
```diff
@@ -0,0 +1,38 @@
...
+  useDeploymentDirectory,
+} from "../nonce-checks/useDeploymentDirectory";
+
+/**
+ * A run that deploys a contract times out
+ */
+describe("execution - deploy run times out", () => {
+  useDeploymentDirectory("minimal-new-api", "deploy-run-times-out");
+
+  it("should error naming timed out transactions", async function () {
```
kanej, line 16: This timeout test leverages not processing a block.

Ideally I would like to add a further test where blocks are continuously processed but the transactions times out because the transaction is not picked up individually.

@alcuadrado is gas limit the best way of doing this?

packages/core/src/new-api/internal/execution/execution-engine.ts
alcuadrado: Do we need a message for this? Shouldn't the normal reducer flow reset TIMEOUT to STARTED?

Not opposed to having it, just trying to understand the reasoning. I do see one advantage: doing it this way the normal reducer creates the state as seen in the journal, and then we decide what to do with timeout as part of the execution. This is handy for things like displaying the current status in plan, whereas switching the TIMEOUTs there wouldn't allow us to do it.

packages/core/src/new-api/internal/execution/execution-engine.ts
alcuadrado: Yes, I think that would be good.

packages/core/src/new-api/internal/execution/execution-engine.ts
```diff
@@ -326,9 +360,16 @@ export class ExecutionEngine {
   }
 
   /**
-   * Wait for the next block to  be processed on-chain
+   * Wait for the next block to  be processed on-chain, monitoring transactions
+   * for timeout at the same time. If timing out transactions complete
+   * the batch we return before the next block.
    */
-  private async _newBlock(state: ExecutionEngineState): Promise<void> {
+  private async _newBlockOrTimeout(
```
alcuadrado, line 367: I think we can wait for a new block, and run the timeout logic after it gets mined. If a future timeouts between two blocks it'd wait at most the block interval time of the chain, which is in the order of seconds.

alcuadrado: This will move timeout handling to process future tick, making things simpler / more consistent.

packages/core/src/new-api/internal/execution/execution-state-reducer.ts
```diff
@@ -120,9 +132,31 @@ export function executionStateReducer(
...
+    return {
+      ...executionStateMap,
+      [action.futureId]: {
+        ...failedExState,
+        status: ExecutionStatus.TIMEOUT,
+      },
+    };
+  }
+
+  if (isExecutionHold(action)) {
```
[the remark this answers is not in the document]
alcuadrado: Let's delay that decision until we implement the last few features

packages/core/src/new-api/internal/execution/onchain-state-transitions.ts
alcuadrado: maybe `registerStartTimeIfNeeded`?

packages/hardhat-plugin/test/execution/timeouts/deploy-run-times-out.ts
alcuadrado: That's an easy way to achieve it, yes.

packages/core/src/new-api/internal/execution/execution-engine.ts
alcuadrado: Well, I think I answered myself.

alcuadrado:
This looks good. I left a few comments, of which the only one I consider somewhat important is [this](https://github.com/NomicFoundation/ignition/pull/306#discussion_r1271466691). We can create an issue for that and merge this.

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: I want to make that message richer as well: chainId and a count of runs. chainId for reconciliation purposes (you can't run on a different chain from previous run) and count of runs to improve both UI and UX of understanding the journal as a log.

packages/core/src/new-api/internal/execution/execution-engine.ts
kanej: I have captured a cleanup task here: https://github.com/NomicFoundation/ignition/issues/320

PR #309, opened 2023-07-18
cleanup/port-integration-tests (PR G)

### LINKED ISSUE #308: Double check integration tests for any missing value
The integration tests are mainly skipped. Port them over checking to see if there is anything important we have missed.

kanej:
Port the skipped integration tests to the new api.

Resolves #308.


packages/core/src/new-api/internal/module-builder.ts
```diff
@@ -120,6 +120,12 @@ export class ModuleConstructor {
       )
     );
 
+    if ((mod as any).results instanceof Promise) {
+      throw new IgnitionError(
```
kanej, line 124: This was feedback from Argentina's hackathon

packages/hardhat-plugin/test/calls.ts
```diff
@@ -1,27 +1,26 @@
...
 import { assert } from "chai";
 
-import { deployModule } from "./helpers";
-import { useEnvironment } from "./useEnvironment";
+import { useEphemeralIgnitionProject } from "./use-ignition-project";
 
-describe.skip("calls", () => {
-  useEnvironment("minimal");
+describe("calls", () => {
+  useEphemeralIgnitionProject("minimal-new-api");
```
kanej, line 8: Open to better names! This adds a `this.deploy` in the same way that `useFileIgnitionProject` does for the nonce-checks.

This deploy is scoped to a fixture project but doesn't write to a deployments folder.

packages/hardhat-plugin/test/use-module.ts
```diff
@@ -1,157 +1,82 @@
...
 
-      assert.equal(x, Number(2));
+      assert.equal(await result.foo.x(), Number(2));
     });
   });
 
-  describe("passing futures into submodules", () => {
-    it("using useModule", async function () {
+  describe("modules depending on modules", () => {
+    // TODO: confirm we are happy for this to go
```
kanej, line 35: Ordering of `useModule` is now not expressable. Just wanting to double check we have remembered that before I delete this test. @alcuadrado

alcuadrado: Yes. Expected behavior.

alcuadrado: If we want to add it later, it can be added as a batching rule. Similar to the malaga rule.

packages/core/src/new-api/internal/module-builder.ts
alcuadrado: nice!

alcuadrado (approved):
Really minimal review, but let's merge it

PR #317, opened 2023-07-21
fix/from-artifact-in-ephemeral-loader

kanej:
Our ephemeral artifact loader fails for actions that pass in the artifact. We need to enhance the ephemeral loader to hold passed in artifacts in memory, but continue to rely on hardhat loading via the `ArtifactResolver` for named contracts/libraries/etc.

## TODO

- [x] Add `storeNamedArtifact` to deployment loader api - to enable load by contract name in ephemeral
- [x] Pass back the artifact against each successful contract deploy in the results
- [x] Update both ephemeral and file deployment loader to work in terms of `futureId` rather than encoding the path to the artifact
- [x] Remove `resolvePath` from Artifact loader


packages/core/src/new-api/internal/types/execution-state.ts
```diff
@@ -125,8 +125,7 @@ export interface DeploymentExecutionState
     | FutureType.NAMED_LIBRARY_DEPLOYMENT
     | FutureType.ARTIFACT_LIBRARY_DEPLOYMENT
   > {
-  storedArtifactPath: string; // As stored in the deployment directory.
-  storedBuildInfoPath?: string; // As stored in the deployment directory. Optional as it's not always present
+  artifactFutureId: string; // As stored in the deployment directory.
```
alcuadrado, line 128: We may want to revisit this later, minimizing the state that we store, and relying more on "view" functions. This would keep the model and reducers simpler. Fine for now though.

packages/core/src/new-api/types/deployment-loader.ts
```diff
@@ -8,9 +8,14 @@ import { Journal } from "./journal";
  */
 export interface DeploymentLoader {
   journal: Journal;
-  loadArtifact(storedArtifactPath: string): Promise<Artifact>;
-  storeArtifact(futureId: string, artifact: Artifact): Promise<string>;
-  storeBuildInfo(buildInfo: BuildInfo): Promise<string>;
+  loadArtifact(artifactFutureId: string): Promise<Artifact>;
+  storeArtifact(futureId: string, artifact: Artifact): Promise<void>;
```
alcuadrado, line 12: Maybe `storeUserProvidedArtifact` or something like that, to make it evident why it's different from the other one.

alcuadrado (approved):

packages/core/src/new-api/internal/types/execution-state.ts
kanej: Recorded as a cleanup task here: https://github.com/NomicFoundation/ignition/issues/322

packages/core/src/new-api/types/deployment-loader.ts
kanej: Done

PR #324, opened 2023-07-24
fix: artifact module loading (PR 3)

kanej:
Pre-artifact loading for the passed Ignition module should use the artifact resolved for named contracts, but read the artifacts from the passed future for user provided cases.

Two integration tests have been added that cover the bug.

### DIFF
--- a/packages/core/src/new-api/internal/deployer.ts
+++ b/packages/core/src/new-api/internal/deployer.ts
@@ -5,7 +5,16 @@ import type {
 } from "../types/module";
 import type { IgnitionModuleDefinition } from "../types/module-builder";
 
-import { isContractFuture } from "../type-guards";
+import { IgnitionError } from "../../errors";
+import {
+  isArtifactContractAtFuture,
+  isArtifactContractDeploymentFuture,
+  isArtifactLibraryDeploymentFuture,
+  isContractFuture,
+  isNamedContractAtFuture,
+  isNamedContractDeploymentFuture,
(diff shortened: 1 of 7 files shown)


packages/core/src/new-api/internal/deployer.ts
```diff
@@ -123,8 +132,11 @@ export class Deployer {
     );
 
     if (reconciliationResult.reconciliationFailures.length > 0) {
-      // TODO: Provide more information
-      throw new Error("Reconciliation failed");
+      const failures = reconciliationResult.reconciliationFailures
+        .map((rf) => `  ${rf.futureId} - ${rf.failure}`)
+        .join("\n");
+
+      throw new IgnitionError(`Reconciliation failed\n\n${failures}`);
```
kanej, line 139: This got included as some cleanup while working through the bug.

packages/core/src/new-api/internal/deployer.ts
```diff
@@ -167,11 +179,33 @@ export class Deployer {
...
+      }
+
+      if (
+        isArtifactContractDeploymentFuture(contract) ||
+        isArtifactContractAtFuture(contract) ||
+        isArtifactLibraryDeploymentFuture(contract)
+      ) {
+        const artifact = contract.artifact;
+
+        entries.push([contract.id, artifact]);
```
kanej, line 203: For user provided artifacts we load from the future.

alcuadrado (approved):

PR #326, opened 2023-07-24
cleanup/remove-old-code (PR 6)

### LINKED ISSUE #297: Remove old code versions
The previous version of the ignition system is still in the code base as incremented tests away from it.

This version should be removed (patching any code that was cross referenced).

## TODO

- [x] #308
- [x] #328
- [x] #329

kanej:
Remove old versions of the code, leaving only the new-api.

This PR does not move the new-api code into new locations - lets get all the PRs merged before we do that.

Fixes #297 

## TODO

- [x] Apply import plugin to eslint to allow for detection of unused modules/exports from modules
- [x] remove ui from plugin - we can bring this back later
- [x] remove old code in `./packages/core/src` and `./packages/core/test`
- [x] deal with any core changes under new-api needed to meet the tighter eslint rules

## Reviewer Guidance

The changes are lots of deletes plus a few minor refactors to keep eslint happy. To keep clear what is going on I would review one commit at a time.


packages/hardhat-plugin/src/index.ts
```diff
@@ -376,7 +371,7 @@ function resolveConfigPath(filepath: string): any {
   }
 }
 
-function resolveParametersString(paramString: string): ModuleParams {
```
kanej, line 380: This needs cleaned up by reworking the module loading with api refinement.

packages/hardhat-plugin/src/index.ts
```diff
@@ -125,7 +123,7 @@ task("deploy")
 
       await hre.run("compile", { quiet: true });
 
-      const userModule: Module<ModuleDict> | undefined = loadModule(
+      const userModule: any | undefined = loadModule(
```
alcuadrado, line 122: Why the `any`s in this file?

packages/core/test/helpers/exact-interface.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-unused-modules */
```
alcuadrado, line 1: I think we should try to remove all the eslint-disable directives later and see what happens.

packages/core/src/new-api/types/journal.ts
```diff
@@ -225,7 +225,7 @@ export type OnchainResultMessage =
   | OnchainResultSuccessMessage
   | OnchainResultFailureMessage;
 
-export type OnchainResultSuccessMessage =
+type OnchainResultSuccessMessage =
```
alcuadrado, line 228: Why remove the export? I think this contradicts the api-extractor recommendations: if a type is used in an exported type, it should also be exported.

alcuadrado: At least for public types

alcuadrado:
Left a few minor comments, but LGTM

packages/hardhat-plugin/src/index.ts
kanej: The module loading code needs cleaned up. It hasn't been touched, it just _happens_ to work.

I intend to do this along with the config refactor.

PR #335, opened 2023-07-25
Deployment Loader Cleanup (PR 7)

### LINKED ISSUE #315: Deployment loader cleanup

zoeyTM:
- [x] remove `isAdapters` type guard as it is not used
- [x] move `types/journals` to the internal types folder
- [x] `types/deployment-loader.ts#Deployer.journal` should be under the private types

fixes #315


kanej:
The text is ambigous but my interpretation was that the `deployment-loader.ts` file should move from `./types` to `./internal/types`.

It would of taken as long to explain it, so I have pushed that change as a commit.

kanej (approved):

alcuadrado (approved):

PR #336, opened 2023-07-25
Create type for DeploymentParams (PR 8)

### LINKED ISSUE #316: Create type for deploymentParameters
We don't have a type for `ExecutionEngineState.deploymentParameters`.

image

The key should be named moduleId. This is repeated 9 times in the codebase. We should find where we are defining it ad-hoc and use that type.

zoeyTM:
fixes #316

### DIFF
--- a/packages/core/src/new-api/internal/validation/validate.ts
+++ b/packages/core/src/new-api/internal/validation/validate.ts
@@ -1,8 +1,8 @@
 import { ArtifactResolver } from "../../types/artifact";
 import {
+  DeploymentParameters,
   FutureType,
   IgnitionModule,
-  ModuleParameters,
 } from "../../types/module";
 import { getFuturesFromModule } from "../utils/get-futures-from-module";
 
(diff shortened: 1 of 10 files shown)


packages/core/src/new-api/internal/validation/validate.ts
```diff
@@ -20,7 +20,7 @@ import { validateSendData } from "./futures/validateSendData";
 export async function validate(
   module: IgnitionModule,
   artifactLoader: ArtifactResolver,
-  moduleParameters: { [key: string]: ModuleParameters }
+  moduleParameters: DeploymentParameters
```
kanej, line 23: Lets rename `moduleParameters` to `deploymentParameters` here.

packages/core/src/new-api/types/module.ts
```diff
@@ -361,6 +361,15 @@ export interface ModuleParameters {
   [parameterName: string]: ModuleParameterType;
 }
 
+/**
+ * An object containing a map of module ID's to their respective ModuleParameters.
+ *
+ * @beta
+ */
+export interface DeploymentParameters {
```
kanej, line 369: I think we should move this to `./types/deployer.ts`, as that is where the user will encounter it first.

What do you think?

kanej (approved):
I left a question about the best place to put the types, but after whatever the decision is there it is good to merge.

packages/core/src/new-api/types/module.ts
alcuadrado: Makes sense to me

PR #346, opened 2023-07-26
Fix for validation bugs (PR 12)

zoeyTM:
Found a bug in the validation code that caused it to only validate the first future of a deployment, and fixing that revealed another bug in the code for validating nested module parameters. Both bugs are fixed in this PR and all affected tests are updated as well.

### DIFF
--- a/packages/core/src/new-api/internal/validation/validate.ts
+++ b/packages/core/src/new-api/internal/validation/validate.ts
@@ -30,38 +30,74 @@ export async function validate(
     await validate(submodule, artifactLoader, deploymentParameters);
   }
 
-  const params = deploymentParameters[module.id] ?? {};
-
   for (const future of futures) {
     switch (future.type) {
       case FutureType.ARTIFACT_CONTRACT_DEPLOYMENT:
-        return validateArtifactContractDeployment(
+        await validateArtifactContractDeployment(
           future,
           artifactLoader,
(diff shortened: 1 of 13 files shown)


packages/core/src/new-api/internal/validation/validate.ts
```diff
@@ -30,41 +30,77 @@ export async function validate(
...
+    await validate(submodule, artifactLoader, deploymentParameters);
   }
 
-  const params = moduleParameters[module.id] ?? {};
-
   for (const future of futures) {
     switch (future.type) {
       case FutureType.ARTIFACT_CONTRACT_DEPLOYMENT:
-        return validateArtifactContractDeployment(
+        await validateArtifactContractDeployment(
```
kanej, line 39: What drove the switch here to using an `await` rather than returning?

alcuadrado: I think it's because they return `Promise<void>`

alcuadrado (approved):

packages/core/src/new-api/internal/validation/validate.ts
zoeyTM: This was the bug causing us to only validate a single future from each deployment. The reason being that we are looping through the futures so returning ends the loop early.

PR #347, opened 2023-07-27
Add accounts index check (PR 1)

### LINKED ISSUE #332: validation/add-accounts-index-check
If a `m.getAccount(...)` call is used we should validate that the index value makes sense given the input accounts at deploy time.

We should also validate that the index is not negative.

## Acceptance Criteria

- [x] a component/unit test (i.e. a core test) that injects higher than expected index
- [x] a test that injects -1 as the index

zoeyTM:
## Acceptance Criteria

- [x] a component/unit test (i.e. a core test) that injects higher than expected index
- [x] a test that injects -1 as the index

fixes #332


kanej (commented):
At what point will the account index outside of bounds be checked? Has execution started?

zoeyTM:
> At what point will the account index outside of bounds be checked? Has execution started?

@kanej yes, the very first step of execution, actually (init execution state in the engine)

alcuadrado (approved):

Issue #351, opened 2023-07-28
ENS Plan mermaid bug


kanej:
Running plan on the ens example generates the report but the mermaid plan errors: 

![image](https://github.com/NomicFoundation/ignition/assets/24030/39450307-a484-4d1c-856f-8a5ab74a63ee)

PR #352, opened 2023-07-28
fix/ens-plan (PR 2)

kanej:
If a call includes the arg list (e.g. "setUniswap(boolean,address)") we need to escape the brackets and commas to form an allowed mermaid id.

I used a `window.diagram` local variable on the plan page to diagnose the issue and intend to retain it to help trouble shooting.

Fixes https://github.com/NomicFoundation/ignition/issues/351.

## Preview

image


alcuadrado:
Option 1, definitely.

packages/ui/examples/ComplexModule.js
```diff
@@ -31,7 +31,7 @@ const synthetixDefinition = defineModule("Synthetix", (m) => {
 
   const { router } = m.useModule(uniswap);
 
-  m.call(synthetixCore, "setUniswap", [router]);
+  m.call(synthetixCore, "setUniswap(boolean,address)", [false, router]);
```
alcuadrado, line 34: This should be just `setUniswap`. If this doesn't work, it's a bug.

packages/ui/test/to-mermaid.ts
```diff
@@ -0,0 +1,269 @@
...
+
+          Module:Contract1["Deploy Contract1"]
+        end
+
+      classDef startModule stroke-width:4px`;
+
+    assertDiagram(moduleDefinition, expectedResult);
+  });
+
+  it.only("should render a module with a space in the name", () => {
```
alcuadrado, line 36: You forgot a `.only` here.

packages/ui/examples/ComplexModule.js
kanej: You mean we should never have a user putting in the parameters in `setUniswap(boolean,address)`?

PR #353, opened 2023-07-28
feat/update-config

### LINKED ISSUE #312: Pass configuration down through deploy
Ensure that configuration passes from hardhat plugin to execution engine properly.

We should double check that we are meeting the HH plugin rules for config and pull up the configuration options used in the exeuction engine to the hardhat plugin config section.

kanej:
Connect the deploy config to the hardhat-ignition plugin configuration.

Resolves #312 

## TODO

- [x] Add number of confirmations to execution logic (currently hardcoded)
- [x] Expose via deploy the config, and wire it to the hardhat plugin config settings
- [x] Bring back the integration config tests
- [x] Bring back docs section

### DIFF
--- a/packages/hardhat-plugin/src/index.ts
+++ b/packages/hardhat-plugin/src/index.ts
@@ -5,7 +5,6 @@ import {
   wipe,
 } from "@ignored/ignition-core";
 import "@nomiclabs/hardhat-ethers";
-import { BigNumber } from "ethers";
 import { existsSync, readdirSync, readJSONSync } from "fs-extra";
 import { extendConfig, extendEnvironment, task } from "hardhat/config";
 import { lazyObject } from "hardhat/plugins";
@@ -24,20 +23,8 @@ import "./type-extensions";
 // eslint-disable-next-line import/no-unused-modules
 export { buildModule } from "@ignored/ignition-core";
 
-export interface IgnitionConfig {
-  maxRetries: number;
-  gasPriceIncrementPerRetry: BigNumber | null;
-  pollingInterval: number;
-  eventDuration: number;
-}
-
 /* ignition config defaults */
 const IGNITION_DIR = "ignition";
-const DEPLOYMENTS_DIR = "deployments";
-const MAX_RETRIES = 4;
-const GAS_INCREMENT_PER_RETRY = null;
-const POLLING_INTERVAL = 300;
-const AWAIT_EVENT_DURATION = 3000; // ms
 
 extendConfig((config, userConfig) => {
   /* setup path configs */
@@ -49,23 +36,12 @@ extendConfig((config, userConfig) => {
       config.paths.root,
       userPathsConfig.ignition ?? IGNITION_DIR
     ),
-    deployments: path.resolve(
-      config.paths.root,
-      userPathsConfig.deployments ?? DEPLOYMENTS_DIR
-    ),
   };
 
   /* setup core configs */
   const userIgnitionConfig = userConfig.ignition ?? {};
 
-  // TODO: this should wire into the new config
-  config.ignition = {
-    maxRetries: userIgnitionConfig.maxRetries ?? MAX_RETRIES,
-    gasPriceIncrementPerRetry:
-      userIgnitionConfig.gasPriceIncrementPerRetry ?? GAS_INCREMENT_PER_RETRY,
-    pollingInterval: userIgnitionConfig.pollingInterval ?? POLLING_INTERVAL,
-    eventDuration: userIgnitionConfig.eventDuration ?? AWAIT_EVENT_DURATION,
-  };
+  config.ignition = userIgnitionConfig;
 });
 
 /**
@@ -162,6 +138,7 @@ task("deploy")
         const adapters = buildAdaptersFrom(hre);
 
         const result = await deploy({
+          config: hre.config.ignition,
           adapters,
           artifactResolver,
           deploymentDir,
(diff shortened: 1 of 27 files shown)


packages/core/src/new-api/internal/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * The number of block confirmations to wait before considering
+ * a transaction to be confirmed during Ignition execution.
+ */
+const DEFAULT_BLOCK_CONFIRMATIONS = 0;
+
+/**
+ * Ignitions default deployment configuration values.
+ */
+export const defaultConfig: DeployConfig = {
```
kanej, line 24: The alternative is to name the file `defaultConfig.ts`.

packages/hardhat-plugin/src/index.ts
```diff
@@ -44,23 +31,12 @@ extendConfig((config, userConfig) => {
       config.paths.root,
       userPathsConfig.ignition ?? IGNITION_DIR
     ),
-    deployments: path.resolve(
```
kanej, line 47: I am not sure we should allow restructuring within the Ignition directory. So for the moment only allow users to set the top level Ignition folder.

packages/hardhat-plugin/test/config.ts
```diff
@@ -1,43 +1,39 @@
...
   it("should only have known config", () => {
-    assert.deepStrictEqual(Object.keys(loadedOptions).sort(), [
-      "eventDuration",
-      "gasPriceIncrementPerRetry",
-      "maxRetries",
-      "networkName",
-      "pollingInterval",
-      "txPollingInterval",
-    ]);
+    const configOptions: KeyListOf<DeployConfig> = [
```
kanej, line 31: It can be done!

packages/hardhat-plugin/test/type-helper.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Where ExpectedConfigOptions is the type level array:
+ * ```
+ * [
+ *    "blockConfirmations",
+ *    "blockPollingInterval",
+ *    "transactionTimeoutInterval",
+ * ]
+ * ```
+ */
+export type KeyListOf<T extends object> = ConvertObjectToTypeArray<
```
kanej, line 16: Lets see if we can crash the typescript compiler.

packages/core/src/new-api/internal/constants.ts
alcuadrado: sounds good

packages/core/src/new-api/internal/constants.ts
```diff
@@ -0,0 +1,28 @@
+import { DeployConfig } from "../types/deployer";
+
+/**
+ * The interval, in milliseconds, between checks to see if a new block
+ * has been created
+ */
+const DEFAULT_BLOCK_POLLING_INTERVAL = 200;
```
alcuadrado, line 7: I think we should do something like `1000` initially, as most chains will take seconds to mine a block.

packages/hardhat-plugin/src/index.ts
alcuadrado: Makes sense

packages/hardhat-plugin/src/type-extensions.ts
```diff
@@ -2,28 +2,25 @@
...
   }
 
   export interface ProjectPathsConfig {
     ignition: string;
-    deployments: string;
   }
 
   export interface HardhatUserConfig {
-    ignition?: Partial<IgnitionConfig>;
+    ignition?: Partial<DeployConfig>;
```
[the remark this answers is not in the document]
alcuadrado: `HardhatUserConfig`: the one written by the user, normally everything is optional.
`HardhatConfig`: The one used at runtime, with defaults resolved, and possibly simplified.

packages/hardhat-plugin/src/type-extensions.ts
```diff
@@ -2,28 +2,25 @@
...
   }
 
   export interface HardhatConfig {
-    ignition: IgnitionConfig;
+    ignition: Partial<DeployConfig>;
```
alcuadrado, line 23: Note that https://github.com/NomicFoundation/ignition/pull/353/files#r1279757963 may give the impression that we should resolve things here. It's not mandatory. This looks good.

alcuadrado (approved):
LGTM. We should just change the default values.

packages/core/src/new-api/internal/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+ * The amount of time, in milliseconds, to wait on a transaction to
+ * confirm before timing out
+ */
+const DEFAULT_TRANSACTION_TIMEOUT_INTERVAL = 3 * 60 * 1000;
+
+/**
+ * The number of block confirmations to wait before considering
+ * a transaction to be confirmed during Ignition execution.
+ */
+const DEFAULT_BLOCK_CONFIRMATIONS = 0;
```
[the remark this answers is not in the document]
kanej: I updated this to be six confirmations.

However, that requires overrides in out tests to set them back to 0 confirmations.

I suspect we shouldn't merge this change until we have auto-detection of automining, and setting of config defaults based on that. What do you think?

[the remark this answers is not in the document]
kanej: Lets add a `getDefaultNumberOfConfirmations` for chain.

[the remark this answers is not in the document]
alcuadrado: let's merge it and tackle the automining thing in a next pr

alcuadrado (approved):

PR #358, opened 2023-07-30
fix: allow staticCall as address in contractAt(FromArtifact)

### LINKED ISSUE #357: validation error when using the output of a staticCall as the address of a contractAt
Reported in the alpha channel:

```solidity
const apm = m.contractAt('APMRegistry', apmAddress);
const registrarAddress = m.staticCall(apm, 'registrar');
const registrar = m.contractAt('ENSSubdomainRegistrar', registrarAddress);
```

Ignition failed with the error:

> IgnitionError: Cannot resolve address of OpenAragonPM:ENSSubdomainRegistrar, not an allowed future type NAMED_STATIC_CALL

zoeyTM:
A check has been added to see if the address a staticCall, and that it resolves to a valid address string.

Also fixed a bug in the static call execution code.

Fixes #357.

### DIFF
--- a/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
+++ b/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
@@ -168,7 +168,11 @@ export class ChainDispatcherImpl implements ChainDispatcher {
 
     const contractInstance = new Contract(contractAddress, abi, signer);
 
-    const result = await contractInstance[functionName](...args, {
+    const validFunctionName = functionName.endsWith("()")
+      ? functionName
+      : `${functionName}()`;
+
+    const result = await contractInstance[validFunctionName](...args, {
       from,
     });
 
--- a/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
+++ b/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
@@ -168,9 +168,23 @@ export class ChainDispatcherImpl implements ChainDispatcher {
 
     const contractInstance = new Contract(contractAddress, abi, signer);
 
-    const validFunctionName = functionName.endsWith("()")
-      ? functionName
-      : `${functionName}()`;
+    // todo: temp fix
+    // if user provided a function name with parenthesis, assume they know what they're doing
+    // if not, add the open parenthesis for the regex test
+    const userFnOrNormalized = functionName.endsWith(")")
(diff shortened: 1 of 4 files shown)


packages/core/src/new-api/internal/execution/chain-dispatcher.ts
```diff
@@ -168,7 +168,11 @@ export class ChainDispatcherImpl implements ChainDispatcher {
 
     const contractInstance = new Contract(contractAddress, abi, signer);
 
-    const result = await contractInstance[functionName](...args, {
+    const validFunctionName = functionName.endsWith("()")
+      ? functionName
+      : `${functionName}()`;
```
kanej, line 173: Can you walk me through this? What case pushed that change.

zoeyTM: Soooo while finding the root of the original bug, I learned that this code didn't actually work before this specific change. The short answer is it's a combination of the way we're invoking the function on the contract object and the way ethers has their methods available on the contract object.

The contract members look like this:
```typescript
const contract = {
  "functionName()": functionName(...),
  functionName: functionName(...)
}
```
so because we're invoking the function through a dynamic string index (`contract[functionName](...)`), for Javascript Reasons :tm:, it only works if it has the parenthesis at the end.

zoeyTM: The reason it didn't fail before is because ethers doesn't throw an error if you try to index the function without the parenthesis, it just.. returns the contract object.. for reasons I don't understand.

kanej: Do we support the "setAddress(bytes,bool)" format for function invocation?

zoeyTM: That is an excellent question, let me check that.

zoeyTM: This code does not support that, which is a problem. Also worth noting that all of this only affects static calls and not regular calls, but still, we do need to support that case as well.

zoeyTM: We decided to add a temporary fix for supporting static calls with parameters, with future plans to revisit this with a more comprehensive solution, [captured here](https://github.com/NomicFoundation/ignition/issues/372)

Issue #369, opened 2023-07-31
Setup typeversions export in `core`


kanej:
We need to double check our typeversion exports in core. Specifically to confirm they are working with pnpm.

See the discussion in the typescript channel for more details.

kanej:
We think this is solved. Lets double check. See Pato's repo.

PR #371, opened 2023-07-31
Update `package.json`s

alcuadrado:
This PR makes sure that the subpath exports are properly set so that the different versions of TS can resolve them. This was based on [this repository I built](https://github.com/alcuadrado/nodejs-exports-and-typescript).

This PR also removes the `engines` fields, to be consistent with Hardhat's packages.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,7 +27,7 @@
   "typesVersions": {
     "*": {
       "ui-helpers": [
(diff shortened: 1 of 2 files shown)


alcuadrado:
Fixes #369

kanej (approved):
How do we indicate node version support - is it convention through the docs?

PR #382, opened 2023-08-03
Switch adapters in favor of provider injection

### LINKED ISSUE #339: Switch adapters to provider injection
Our adapters devolve down on to the hre provider, and the provider is a common interface point for js eth projects.

Instead of passing in adapters we support passing a eip1193 provider. This allows us to separate out the ethers types from the adapters api.

zoeyTM:
Made adapters internal and instead accept a generic EIP-1193 provider

fixes #339


kanej (approved):
LGTM!

PR #383, opened 2023-08-03
added a simple test for complete example

### LINKED ISSUE #337: bring back complete example test
Raised as feedback here: https://github.com/NomicFoundation/ignition/pull/319#discussion_r1273566401

The complete example has had its test removed. It was previously skipped. We should investigate whether the test can be brought back and updated to the new api in a way that adds value by either being illustrative or a useful regression test.

zoeyTM:
fixes #337


kanej (approved):
Looks good.

PR #386, opened 2023-08-08
refactor/rework-execution-strategy

kanej:


kanej:
We have move the refactoring to another PR: https://github.com/NomicFoundation/ignition/pull/387

PR #387, opened 2023-08-10
refactor: update state and errors (PR 1)

kanej:
This is a WIP branch for breaking work on error handling and state changes.

Currently it pulls in the new call example code and new core types under:

```shell
./packages/core/src/new-api/internal/new-execution
```

The build is broken. But commits should endeavor to:

* build in `./packages/core`
* test in `./packages/core`

We can fix linting and the integration tests as the last move.

## TODO

* [x] Setup `./new-execution` folder to allow development without breaking existing code
* [x] Add hardhat and hh based integration tests to core to test low level error handling code
* [x] Add new messages + reducer based on new types
* [x] Integrate on chain calls and new reducer into execution engine


packages/core/src/new-api/internal/execution/transaction-types.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
alcuadrado, line 1: This file was replaced by `packages/core/src/new-api/internal/new-execution/types/jsonrpc.ts`

packages/core/src/new-api/internal/new-execution/error-decoding.ts
```diff
@@ -0,0 +1,146 @@
+import {
```
alcuadrado, line 1: Maybe we should merge this into `./abi.ts`.

packages/core/src/new-api/internal/new-execution/basic-execution-strategy.ts
```diff
@@ -0,0 +1,227 @@
+import {
+  decodeArtifactCustomError,
+  decodeArtifactFunctionCallResult,
+  encodeArtifactDeploymentData,
+  encodeArtifactFunctionCall,
+  executeOnchainInteractionRequest,
+  executeStaticCallRequest,
+} from "./execution-strategy-helpers";
```
alcuadrado, line 8: This should be the only import that's not types or enums that an execution strategy should need.

packages/core/src/new-api/internal/new-execution/jsonrpc-calls.ts
```diff
@@ -0,0 +1,600 @@
+import { IgnitionError } from "../../../errors";
```
alcuadrado, line 1: This includes all the RPC calls that we were already doing. This replaces both the adapters and chain dispatcher.

packages/core/src/new-api/internal/new-execution/reducers/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { IgnitionError } from "../../../../errors";
+
+export function assertUnknownAction<T>(actionType: never): T {
```
alcuadrado, line 3: Let's try not to add this. If we design everything carefully we shouldn't need neither this nor type guards in the execution state and reducers.

packages/core/src/new-api/internal/new-execution/type-guards/execution-state.ts
```diff
@@ -0,0 +1,82 @@
+import {
```
alcuadrado, line 1: As mentioned in the previous file, we shouldn't need this. Let's try not to add these functions. They add a ton of code and potentially hide bugs. Now everything has a `type`.

packages/core/src/new-api/internal/new-execution/type-guards/execution-state.ts
```diff
@@ -0,0 +1,82 @@
...
+    isCallExecutionState(potential) ||
+    isStaticCallExecutionState(potential) ||
+    isContractAtExecutionState(potential) ||
+    isReadEventArgumentExecutionState(potential) ||
+    isSendDataExecutionState(potential)
+  );
+}
+
+export function isDeploymentExecutionState(
+  potential: unknown
```
alcuadrado, line 25: Thinking about every typeguard as taking `unknown` may seem to indicate that they are needed. But this one, for example, is only used in context where we already know we have an execution state, so all we need to do is using its `type`.

packages/core/src/new-api/internal/new-execution/types/execution-result.ts
```diff
@@ -0,0 +1,103 @@
+import {
+  FailedEvmExecutionResult,
+  SuccessfulEvmExecutionResult,
+} from "./evm-execution";
+
+/**
```
alcuadrado, line 6: I used to have an extra type: `SIMULATION_ERROR`.

I removed it as those errors shouldn't be considered permanent (i.e. not journaled), because a re-run may make them pass.

I think I regret having removed it. It creates this new entity in the system, where execution may result in an ExecutionResult or a simulation error. 

I'm starting to think that I should reintroduce it, and we can decide not to journal it anyways.

WDYT?

packages/core/src/new-api/internal/new-execution/types/execution-strategy.ts
```diff
@@ -0,0 +1,241 @@
...
+ */
+export type StaticCallRequest = Omit<StaticCall, "result">;
+
+/**
+ * The response of a static call request.
+ */
+export type StaticCallResponse = RawStaticCallResult;
+
+/**
+ * Each execution strategy defines how each type of execution state is executed.
```
alcuadrado, line 108: Please, pay extra attention to this.

packages/core/src/new-api/internal/new-execution/types/jsonrpc.ts
```diff
@@ -0,0 +1,70 @@
+/**
```
alcuadrado, line 1: All of this types used to store things in raw json-rpc format (e.g. numbers were 0x encoded strings). Now I changed them to more idiomatic types (e.g. numbers are `bigint`).

packages/core/src/new-api/internal/utils/collectLibrariesAndLink.ts
```diff
@@ -35,7 +35,11 @@ export async function collectLibrariesAndLink(
   )) {
     if (!isAddress(linkedLibraryAddress)) {
       throw new IgnitionValidationError(
-        `You tried to link the contract ${artifact.contractName} with the library ${linkedLibraryName}, but provided this invalid address: ${linkedLibraryAddress}`
+        `You tried to link the contract ${
+          artifact.contractName
+        } with the library ${linkedLibraryName}, but provided this invalid address: ${
```
alcuadrado, line 40: Just an eslint thing

packages/core/src/new-api/types/provider.ts
```diff
@@ -15,6 +13,6 @@ export interface RequestArguments {
  *
  * @beta
  */
-export interface EIP1193Provider extends EventEmitter {
+export interface EIP1193Provider {
```
alcuadrado, line 16: We don't use that interface, so we can avoid it, making mocking this way simpler.

packages/core/test-integrations/fixture-projects/dont-throw-on-reverts/hardhat.config.js
```diff
@@ -0,0 +1,9 @@
+/** @type import('hardhat/config').HardhatUserConfig */
+module.exports = {
+  solidity: "0.8.19",
+  networks: {
+    hardhat: {
+      throwOnTransactionFailures: false,
```
alcuadrado, line 6: We only need this for one test

packages/core/test-integrations/test-execution-result-fixtures.ts
```diff
@@ -0,0 +1,83 @@
+import { assert } from "chai";
```
alcuadrado, line 1: This test ensures that the fixture that we use in the `test/` folder is correct. It actually runs every call that the fixture is meant to simulate.

This allows us to keep `test/` independent from Hardhat network and fast, but be confident it is correct.

packages/core/test/new-api/new-execution/abi.ts
```diff
@@ -0,0 +1,174 @@
...
+import { decodeArtifactFunctionCallResult } from "../../../src/new-api/internal/new-execution/abi";
+import { EvmExecutionResultTypes } from "../../../src/new-api/internal/new-execution/types/evm-execution";
+import { artifact, fixture } from "../../helpers/execution-result-fixtures";
+
+describe("abi", () => {
+  // These tests validate that type conversions from the underlying abi library
+  // (ethers v6 as of this writing) are working as expected, and that no type
+  // of the library is used directly in the public API.
+  describe("Type conversions", () => {
+    // TODO @alcuadrado
```
alcuadrado, line 12: Consider these `TODO @alcuadrado`s as my plan for Monday

packages/core/test/new-api/new-execution/reducers/deployment-state-reducer.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    describe("initialization", () => {
+      beforeEach(() => {
+        updatedState = applyMessages([initializeNamedContractDeployMessage]);
+      });
+
+      it("should populate a deployment execution state for the future", () => {
+        const exState = updatedState.executionStates.future1;
+        assert.isDefined(exState);
+        assert(isDeploymentExecutionState(exState));
```
alcuadrado, line 93: This could be `assert.equal(exState.type, ...)`, so that we don't need a type guard, and also the error message would be better if it fails.

packages/hardhat-plugin/test/fixture-projects/minimal-new-api/contracts/C.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: MIT
```
alcuadrado, line 1: This file can now be deleted. If this fixture-project was added for my decoding code, it can be deleted in its entirety.

We can also move many tests to core/ now

alcuadrado:
Why does this PR make the example tests fail?

alcuadrado:
I split the linter into its own CI job so that the tests are still run in the presence of linter errors.

I found out that my integration tests were failing, only to discovered that the rest of the tests pass because all of them have an eslint-disable directive? We should fix the config instead.

packages/core/src/new-api/internal/new-execution/reducers/deployment-state-reducer.ts
```diff
@@ -0,0 +1,92 @@
...
+        executionStates: executionStatesReducer(state.executionStates, action),
+      };
+    default:
+      return assertUnknownAction(action);
+  }
+}
+
+function executionStatesReducer(
+  state: ExecutionStateMap = {},
+  action: Exclude<JournalMessage, RunStartMessage>
```
alcuadrado, line 50: I found something useful for the reducers. You don't need to do this exclude thing, which will break when you add more things.

Instead, if you first write the function call to `executionStatesReducer`, and then ask vscode to generate it, it has tighter types. 

That autogenerated code normally looks ugly, but if you add `import * as Messages from "../types/messages";` they look good.

Also, those tighter types eliminate the need to do `default: assert unknown`.

packages/core/src/new-api/internal/new-execution/types/messages.ts
```diff
@@ -0,0 +1,96 @@
...
+  | TransactionSendMessage
+  | TransactionConfirmMessage;
+
+export enum JournalMessageType {
+  RUN_START = "RUN_START",
+  DEPLOYMENT_EXECUTION_STATE_INITIALIZE = "DEPLOYMENT_EXECUTION_STATE_INITIALIZE",
+  DEPLOYMENT_EXECUTION_STATE_COMPLETE = "DEPLOYMENT_EXECUTION_STATE_COMPLETE",
+  CALL_EXECUTION_STATE_INITIALIZE = "CALL_EXECUTION_STATE_INITIALIZE",
+  CALL_EXECUTION_STATE_COMPLETE = "CALL_EXECUTION_STATE_COMPLETE",
+  NETWORK_INTERACTION_REQUEST = "NETWORK_INTERACTION_REQUEST",
```
alcuadrado, line 27: I think we are missing these:

`TRANSACTION_SIMULATION_ERROR`
`TRANSACTION_DROPPED`
`TRANSACTION_REPLACED_BY_USER_INTERFERENCE`

alcuadrado: Oh, simulation errors are encoded as `DEPLOYMENT_EXECUTION_STATE_COMPLETE` and `CALL_EXECUTION_STATE_COMPLETE`, right?

alcuadrado: Don't we need a message for `StaticCall` results?

PR #388, opened 2023-08-15
Switch configurations depending on detecting of auto-mining

### LINKED ISSUE #363: Switch configurations depending on detecting of auto-mining
Different default values for number of block confirmations should be set depending on whether an automining chain is detected.

If a user is overriding the values, these take precedence.

In the deploy function the `deployConfig` is merged with default values. At that point it should be determined if we are in an automining chain.

## Implementation details

* For hardhat and anvil: hardhat_getAutomine — Returns a boolean
* For ganache: web3_clientVersion and check if the string result includes "Ganache"

```js
if (isHardhatGetAutomineTrue) {
   return true;
}

if (isGanache) {
   return true; // This can be a false positive
}

return false;
```

We also want to add a validation on the passed in deployconfig that throws if the number of confirmations is below 1.

## TODO

- [x] Detect automining chain (local HH test node)
- [x] Rework defaults based on the check
- [x] Add validation check - 0 confirmations is an invalid value
- [x] Check the integration tests, unit tests and examples for config overrides that are no longer required

zoeyTM:
- [x] Detect automining chain (local HH test node)
- [x] Rework defaults based on the check
- [x] Add validation check - 0 confirmations is an invalid value
- [x] Check the integration tests, unit tests and examples for config overrides that are no longer required

fixes #363


packages/core/src/new-api/deploy.ts
```diff
@@ -48,11 +49,14 @@ export async function deploy({
 
   const chainDispatcher = new ChainDispatcherImpl(buildAdaptersFrom(provider));
 
+  const isAutominedNetwork = await checkAutominedNetwork(provider);
```
kanej, line 52: I think we should push this check down into the `deploy` function of `deployer`.

WDYT?

packages/hardhat-plugin/test/config.ts
```diff
@@ -1,39 +1,58 @@
...
+
+    it("should throw when given a `blockConfirmations` value less than 1", async function () {
+      const moduleDefinition = buildModule("FooModule", (m) => {
+        const foo = m.contract("Foo");
+
+        return { foo };
+      });
 
-    assert.deepStrictEqual(Object.keys(loadedOptions).sort(), configOptions);
+      await assert.isRejected(
```
kanej, line 52: I think the deploy function for `useEphemeralIgnitionProject` isn't passing the config from `hre.config.ignition` into the deploy call here.

packages/core/src/new-api/deploy.ts
kanej: After discussion we are happy with the current location. Async in a constructor would be bad.

packages/hardhat-plugin/test/config.ts
kanej: Done.

packages/core/src/new-api/internal/utils/check-automined-network.ts
```diff
@@ -0,0 +1,23 @@
...
+export async function checkAutominedNetwork(
+  provider: EIP1193Provider
+): Promise<boolean> {
+  const isHardhat = Boolean(
+    await provider.request({ method: "hardhat_getAutomine" })
+  );
+
+  if (isHardhat) {
+    return true;
+  }
```
alcuadrado, line 12:
```suggestion
  try {
    const isHardhat = Boolean(
      await provider.request({ method: "hardhat_getAutomine" })
    );

    if (isHardhat) {
      return true;
    }
  } catch {
    // If this method failed we aren't using Hardhat Network nor Anvil, so we 
    // just continue with the next check.
  }
```

PR #391, opened 2023-08-17
Add artifact function name validation for call and static call (PR 1)

### LINKED ISSUE #390: Update validation logic to leverage new low level calls
Pato implemented helpers to validate the function names, and to resolve them later. They enforce that the users use the bare name (e.g. foo not foo()) if there's no overload, as having a single representation is better for reconciliation.

There's a `validateArtifactFunctionName` that we need to plug into the validation logic. Then the abi encoding/decoding will be done by the execution strategy.

The function is here: https://github.com/NomicFoundation/ignition/blob/4b85d9a333d4ff9c17dc16b1f7b61cd1583a548f/packages/core/src/new-api/internal/new-execution/abi.ts#L159

Same to validate the library names: https://github.com/NomicFoundation/ignition/blob/4b85d9a333d4ff9c17dc16b1f7b61cd1583a548f/packages/core/src/new-api/internal/new-execution/libraries.ts#L26

zoeyTM:
fixes #390

### DIFF
--- a/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
+++ b/packages/core/src/new-api/internal/execution/chain-dispatcher.ts
@@ -172,25 +172,7 @@ export class ChainDispatcherImpl implements ChainDispatcher {
 
     const contractInstance = new Contract(contractAddress, abi, signer);
 
-    // todo: temp fix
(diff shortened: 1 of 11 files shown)


alcuadrado:
This looks good, but for some reason the CI isn't running.

PR #392, opened 2023-08-18
Remove temp fix for static call function names (PR 2)

### LINKED ISSUE #372: Write a more comprehensive solution for supporting static calls with arguments
Currently we have a hotfix in place, but we should flesh out a more intelligent solution that includes supporting overloaded functions.

We think the next step is investigation into whether the existing shim for calls during execution can now be removed given the helper functions that have been added and the enforcement during validation.

zoeyTM:
The ticket originally was for writing a more comprehensive solution for static call function names with args, but this is handled in validation now so this issue was largely resolved already. Just removing the temp fix now.

fixes #372


alcuadrado (approved):

alcuadrado:
We can ignore the linter errors. let's merge

PR #397, opened 2023-08-23
test/update state and errors with future processor test

kanej:
This also fixes some minor lookup bugs for network interactions, and a bug with simulation type conversion.

### DIFF
--- a/packages/core/src/new-api/internal/new-execution/future-processor/handlers/send-transaction.ts
+++ b/packages/core/src/new-api/internal/new-execution/future-processor/handlers/send-transaction.ts
@@ -4,26 +4,29 @@ import { NonceManager } from "../../nonce-management";
 import { TransactionTrackingTimer } from "../../transaction-tracking-timer";
 import { ExecutionResultType } from "../../types/execution-result";
 import {
-  DeploymentExecutionState,
   CallExecutionState,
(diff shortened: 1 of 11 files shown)


packages/core/src/new-api/internal/new-execution/future-processor/handlers/send-transaction.ts
```diff
@@ -77,7 +82,10 @@ export async function sendTransaction(
     lastNetworkInteraction,
     async (_sender: string) => nonceManager.getNextNonce(_sender),
     async (simulationResult) => {
-      const response = await strategyGenerator.next(simulationResult);
+      const response = await strategyGenerator.next({
+        type: OnchainInteractionResponseType.SIMULATION_RESULT,
+        result: simulationResult,
+      } as any);
```
kanej, line 88: If this isn't wrapped up as a `SimulationResult`, the basic strategy errors on the first assertion, claiming it shouldn't recieve raw simulation results.

I was unsure what the intent was here, and so haven't changed the types.

How do you want me to approach this?

kanej: Constraining the generator resolved the issue.

PR #398, opened 2023-08-23
Switch execution engine (PR 2)

alcuadrado:
This PR switches the old execution engine for the new one.

This effort meant changing a ton of code due to several interfaces having changed, so I did some cleanups in the meantime:

* I created helpers for the reconciler.
* The reconciler is now async, so we don't need the artifacts map.
* I improved the error messages that the reconciler creates, giving more info to help users understand what happened.
* Updated the `DeployConfig` as it now has different values, and moved the config resolution to `deploy()`, so that anything except for the public functions receives it "ready".
* Simplified the `MemoryJournal` while debugging the bug fixed in https://github.com/NomicFoundation/ignition/commit/c67a960c5a93fa6faa126266f8c9395d60dac8ea

I deleted the old execution engine and its tests, but some of them may be worth porting. These may be good candidates:

* `packages/core/test/new-api/execution/execution-engine.ts`
* `packages/core/test/new-api/execution/restart.ts`

This PR does not update the plugin tests and adds `.skip` to some that would not pass otherwise.


packages/core/src/new-api/internal/journal/memory-journal.ts
```diff
@@ -1,33 +1,26 @@
...
-  private messages: string[] = [];
+  private messages: JournalMessage[] = [];
 
   constructor(private _verbose: boolean = false) {}
 
   public record(message: JournalMessage): void {
     this._log(message);
 
-    this.messages.push(JSON.stringify(message, serializeReplacer.bind(this)));
+    this.messages.push(message);
```
kanej, line 19: That was there to check that we didn't introduce any messages that we can't serialize.

Happy for it to go now, do you think it is worth while adding an equivalent test? Or are changes now likely to come with their own tests.

kanej (approved):
Awesome. LGTM!

PR #401, opened 2023-08-24
Bring back UI for execution

### LINKED ISSUE #259: Show UI during execution
The previous cli ui was based on react ink. We want to bring it back. However we do not want to expose the entire internal execution state.

The initial idea is to inject an event emitter into the deploy process that the UI listens to (is this just the journal?). The UI builds up its own ui view orientated version of the current state and renders it with react ink. We suspect we can bring back alot of the previous react ink code (see older commits before it was removed to keep eslint happy).

## TODO

- [x] Design for re-integrating UI react code at the command line based on events
- [x] Bring back UI components
- [x] Wire through event emitter into the deploy and execution engine
- [ ] https://github.com/NomicFoundation/ignition/issues/402
- [ ] https://github.com/NomicFoundation/ignition/issues/419
- [ ] https://github.com/NomicFoundation/ignition/issues/433
- [ ] https://github.com/NomicFoundation/ignition/issues/434

zoeyTM:
fixes #259


zoeyTM:
Added a `VerboseEventHandler`

Screen Shot 2023-08-28 at 2 13 29 PM

alcuadrado:
Now that we decided to have independent `ExecutionEvent`s that don't necessarily map 1:1 with `JournalMessage`s, I think we can apply this simplification: Instead of having the deployment loader emit the `ExecutionEvent`s, emit them directly whenever needed.

I think this would mean emitting them in:

* `Deployer._getOrInitializeDeploymentState` when a new deployment gets started.
* `Deployer.deploy`, to emit the `BATCH_INITIALIZE` event.
* Maybe in the execution engine to emit an event when each batch started (maybe I missed this event)
* All the places where `applyNewMessage` is called, except maybe in the `Wiper`.

Also, how annoying is it to have to use this weird syntax?

```ts
public [ExecutionEventType.RUN_START](event: RunStartEvent): void {
// ....
}
```

zoeyTM:
@alcuadrado ah okay, i can update that!

As for the weird syntax, we don't technically have to do it that way. We could also do
```typescript
public RUN_START(event: RunStartEvent): void {
// ....
}
```

I just did it the current way because I figured it was one less thing we have to update if we ever change one of the enum values. Now that I'm typing this out, though, I'm thinking there's not really a reason besides convention that we couldn't do something like:
```typescript
export enum ExecutionEventType {
  RUN_START = "runStart",
  WIPE_EXECUTION_STATE = "wipeExecutionStart",
  // etc...
}
```

which would make the handler functions just normal function names. Thoughts?

packages/hardhat-plugin/src/ui/components/execution/FinalStatus.tsx
```diff
@@ -0,0 +1,117 @@
+import { Box, Text } from "ink";
+import flattenDeep from "lodash.flattendeep";
+
+import { UiFuture, UiFutureStatusType, UiState } from "../../types";
+
+import { AddressResults } from "./AddressResults";
+import { Divider } from "./Divider";
+
+export const FinalStatus = ({ state }: { state: UiState }) => {
+  const allFutures = flattenDeep(state.batches);
```
alcuadrado, line 10: This can be done with a builtin array method now

```suggestion
  const allFutures = state.batches.flat(Infinity);
```

https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/array/flat

kanej: Done

PR #408, opened 2023-08-24
test/reenable-integration-tests

kanej:
The execution tests and some harder problems remain skipped but the broad basic integration tests are re-enabled including:

* calls
* library
* module params
* static calls
* useModule
* contract

There is a small bugfix for a missing id when passing a transaction for a call.

### DIFF
--- a/packages/core/src/new-api/internal/new-execution/basic-execution-strategy.ts
+++ b/packages/core/src/new-api/internal/new-execution/basic-execution-strategy.ts
@@ -90,7 +90,7 @@ export class BasicExecutionStrategy implements ExecutionStrategy {
       {
         id: 1,
         type: NetworkInteractionType.ONCHAIN_INTERACTION,
-        to: undefined,
+        to: executionState.contractAddress,
         data: encodeArtifactFunctionCall(
           artifact,
           executionState.functionName,
--- a/packages/hardhat-plugin/test/events.ts
+++ b/packages/hardhat-plugin/test/events.ts
@@ -7,7 +7,7 @@ import { useEphemeralIgnitionProject } from "./use-ignition-project";
 describe("events", () => {
   useEphemeralIgnitionProject("minimal-new-api");
 
-  it("should be able to use the output of a readEvent in a contract at", async function () {
+  it.skip("should be able to use the output of a readEvent in a contract at", async function () {
     const moduleDefinition = buildModule("FooModule", (m) => {
       const account1 = m.getAccount(1);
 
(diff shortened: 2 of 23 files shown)


packages/core/src/new-api/internal/new-execution/basic-execution-strategy.ts
```diff
@@ -90,7 +90,7 @@ export class BasicExecutionStrategy implements ExecutionStrategy {
       {
         id: 1,
         type: NetworkInteractionType.ONCHAIN_INTERACTION,
-        to: undefined,
+        to: executionState.contractAddress,
```
kanej, line 93: This is the tweak to call transactions, `undefined` makes sense for a deploy but I think it should be `exState.contractAddress` for a general call.

packages/hardhat-plugin/test/events.ts
```diff
@@ -7,7 +7,7 @@ import { useEphemeralIgnitionProject } from "./use-ignition-project";
 describe("events", () => {
   useEphemeralIgnitionProject("minimal-new-api");
 
-  it("should be able to use the output of a readEvent in a contract at", async function () {
+  it.skip("should be able to use the output of a readEvent in a contract at", async function () {
```
kanej, line 10: These `events` tests will need more investigation. I suspect this is a data encoding/decoding issue but haven't dug in.

packages/hardhat-plugin/test/execution/deploy-contract-at-from-artifact.ts
```diff
@@ -13,7 +13,7 @@ import {
  * First deploy a working contract, then reuse it from a subsequent module
  * with a passed in artifact.
  */
-describe("execution - deploy contractAt from artifact", function () {
+describe.skip("execution - deploy contractAt from artifact", function () {
```
kanej, line 16: The execution tests have been skipped in general. They will all need revisited.

packages/hardhat-plugin/test/static-calls.ts
```diff
@@ -78,15 +78,16 @@ describe("static calls", () => {
     assert.equal(await result.foo.x(), Number(1));
   });
 
-  it("should be able to use the output of a static call function in a contract at (with function signature)", async function () {
+  // TODO: this needs more investigation
+  it.skip("should be able to use the output of a static call function in a contract at (with function signature)", async function () {
```
kanej, line 82: We are getting a validation based on overloads here. I suspect because the validation logic is misidentifying overloads.

packages/core/src/new-api/internal/new-execution/basic-execution-strategy.ts
alcuadrado: Yes, this was definitely a bug

PR #409, opened 2023-08-25
fix/static-call-validation

kanej:
A few checks where relying on the previous version of ethers. The ethers code has been moved down into abi, so it is encapsulated.

### DIFF
--- a/packages/core/src/new-api/internal/new-execution/abi.ts
+++ b/packages/core/src/new-api/internal/new-execution/abi.ts
@@ -12,7 +12,7 @@ import {
   UnsupportedOperationError,
 } from "../../../errors";
 import { Artifact } from "../../types/artifact";
-import { SolidityParameterType } from "../../types/module";
+import { ArgumentType, SolidityParameterType } from "../../types/module";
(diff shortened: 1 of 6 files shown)


packages/core/src/new-api/internal/new-execution/abi.ts
```diff
@@ -142,6 +142,43 @@ export function decodeArtifactFunctionCallResult(
...
+ * Validates that a function is valid for the given artifact. That means:
+ *  - It's a valid function name
+ *    - The function name exists in the artifact's ABI
+ *    - If the function is not overlaoded, its bare name is used.
+ *    - If the function is overloaded, the function name is includes the argument types
+ *      in parentheses.
+ * - The function has the correct number of arguments
+ * - The function is has a pure or view state mutability
+ */
+export function validateArtifactFunction(
```
alcuadrado, line 155: Maybe do the pure/view check optional and also reuse this function in the validation of call futures?

kanej: Done! Not the biggest fan of boolean flags but this cleans up `call` as well now.

alcuadrado (approved):

PR #416, opened 2023-08-27
Fix readEventArgument validation and decoding

alcuadrado:
This PR forbids using arguments with dynamic types as their values are hashed and used as topics, not stored.

It also fixes how the arguments are decoded, by providing ethers the topics and not just the data of the log.

It also re-enables the argument-related integration tests.

### DIFF
--- a/packages/core/src/new-api/internal/new-execution/abi.ts
+++ b/packages/core/src/new-api/internal/new-execution/abi.ts
@@ -176,22 +176,25 @@ export function validateArtifactEventArgumentParams(
 
   const eventFragment = getEventFragment(iface, eventName);
 
-  if (typeof argument === "string") {
-    for (const input of eventFragment.inputs) {
-      if (input.name === argument) {
-        return;
-      }
+  const paramType = getEventArgumentParamType(
+    emitterArtifact.contractName,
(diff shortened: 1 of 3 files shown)


packages/core/src/new-api/internal/new-execution/abi.ts
```diff
@@ -176,22 +176,25 @@ export function validateArtifactEventArgumentParams(
...
-      `Event ${eventName} of contract ${emitterArtifact.contractName} has only ${eventFragment.inputs.length} arguments, but argument ${argument} was requested`
-    );
+  if (paramType.indexed === true) {
+    // We can't access the value of indexed arguments with dynamic size
+    if (
+      paramType.isArray() ||
+      paramType.isTuple() ||
+      paramType.type === "bytes" ||
+      paramType.type === "string"
+    ) {
```
kanej, line 193: Are there any fixed sized types that also don't get indexed?

alcuadrado: As discussed on a call: no, because fixed-sized types have 32 bytes (i.e. an EVM word) or less.

kanej (approved):

PR #422, opened 2023-08-28
chore: re-enable samples

### LINKED ISSUE #421: re-enable samples
We want to bring back the samples projects to:

* provide a useful regression test
* to show users example ignition modules

kanej:
Upgrade the samples to toolbox 3.0.0 so that ethers v6 is used throughout the monorepo.

This lets us bring back the samples, with the exception of the full uniswap test as the test depends on the uniswap sdk which is ethers v5 based. That test remains skipped but a simpler uniswap test has been added that asserts that the deploy has completed.

Includes a minor fix to the basic execution strategy to fix the `to` of send requests.

Resolves #421.

### DIFF
--- a/examples/sample/test/Lock.js
+++ b/examples/sample/test/Lock.js
@@ -1,12 +1,12 @@
 const {
   time,
   loadFixture,
-} = require("@nomicfoundation/hardhat-network-helpers");
+} = require("@nomicfoundation/hardhat-toolbox/network-helpers");
 const { anyValue } = require("@nomicfoundation/hardhat-chai-matchers/withArgs");
 const { expect } = require("chai");
 const LockModule = require("../ignition/LockModule");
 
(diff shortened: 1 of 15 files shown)


examples/sample/test/Lock.js
```diff
@@ -51,7 +50,7 @@ describe.skip("Lock", function () {
         deployOneYearLockFixture
       );
 
-      expect(await ethers.provider.getBalance(lock.address)).to.equal(
+      expect(await ethers.provider.getBalance(lock.target)).to.equal(
```
alcuadrado, line 53: What's `.target`?

alcuadrado (approved):

examples/sample/test/Lock.js
kanej: This is coming from the new hardhat sample.

PR #423, opened 2023-08-29
New deployment result types

alcuadrado:


packages/core/src/new-api/internal/deployer.ts
```diff
@@ -52,12 +57,20 @@ export class Deployer {
...
+      ModuleIdT,
+      ContractNameT,
+      IgnitionModuleResultsT
+    >,
     deploymentParameters: DeploymentParameters,
     accounts: string[],
     defaultSender: string
-  ): Promise<DeploymentResult> {
+  ): Promise<DeploymentResult<ContractNameT, IgnitionModuleResultsT>> {
     await validateStageTwo(
```
alcuadrado, line 74: This is something you can take over if you want @kanej.

Here the validate functions are throwing an error that doesn't give you access to the futureId that caused the error.

Ideally, the validation function would return all the errors. A quick fix is changing the error they throw so that they include the futureId. We can do the proper fix later.

packages/hardhat-plugin/src/ignition-helper.ts
```diff
@@ -55,52 +81,123 @@ export class IgnitionHelper {
...
+${errorsList.join("\n")}`
+      );
+    }
 
-    for (const [key, future] of Object.entries(result.module.results)) {
-      const deployedContract = result.contracts[future.id];
+    if (result.type === DeploymentResultType.EXECUTION_ERROR) {
+      console.log(result);
+      throw new IgnitionError(
+        `The deployment wan't succesful: TODO: create a good error message`
```
alcuadrado, line 117: This is also something else that you can make progress on, @kanej

packages/hardhat-plugin/src/index.ts
```diff
@@ -125,70 +125,28 @@ task("deploy")
...
+        provider: hre.network.provider,
+        artifactResolver,
+        deploymentDir,
+        ignitionModule: userModule,
+        deploymentParameters: parameters ?? {},
+        accounts,
+        verbose: logs,
+      });
+
+      // TODO: print a better result
```
alcuadrado, line 148: I removed this just to make it compile. We should reintroduce better messages.

packages/hardhat-plugin/test/use-ignition-project.ts
```diff
@@ -14,6 +10,15 @@ import { IgnitionHelper } from "../src/ignition-helper";
 import { clearPendingTransactionsFromMemoryPool } from "./execution/helpers";
 import { waitForPendingTxs } from "./helpers";
 
+declare module "mocha" {
```
alcuadrado, line 13: @kanej I tried to type these helpers to be able to debug some things, but I don't think the other `deploy` helper can be typed.

To be honest, I find creating this kind of helpers really hard to reason, especially since they have type `any` unless you type them.

alcuadrado:
I accidentally broke a plugin test. That's confusing me a ton. I'll debug it early tomorrow.

alcuadrado:
Also, I think this works nice with TypeChain, but I still need to test it.

packages/hardhat-plugin/test/use-ignition-project.ts
kanej: I have pushed a temporary typing. I split out the two deploys (file based and ephemeral), one is now `runControlledDeploy` and is mainly used in the execution that is currently skipped.

I have no problem pulling these helpers out/replacing the tests with a different structure. We talked about moving the execution tests currently skipped into core, which may simplify the integration tests enough this issue goes away.

packages/core/src/new-api/internal/deployer.ts
kanej: I am tackling this now.

I am going to alter the validate stage two to return [futureId, validationErrorMessage] pairs (or some such) that can then be returned as a result if there are any validation failures.

kanej: Alright, I backed out of the larger change and did as you suggested.

Stage two now returns either a `ValidationErrorDeploymentResult` or null, and deploy returns the error result if detected.

packages/core/src/new-api/internal/validation/validateStageTwo.ts
```diff
@@ -19,105 +25,122 @@ export async function validateStageTwo(
...
-): Promise<void> {
+): Promise<ValidationErrorDeploymentResult | null> {
   const futures = getFuturesFromModule(module);
 
-  // originally, I wrote a getSubmodulesFromModule function similar to the one above
-  // that recursively retrieved all submodules regardless of how deeply nested they were.
-  // however, by taking only the top level submodules of the current depth and recursively
-  // validating each of those, we achieve the same effect.
-  const submodules = module.submodules;
-  for (const submodule of submodules) {
```
kanej, line 30: Get futures from module, gets the futures of all submodules as well recursively. Have I missed a case that submodule recursion gets us here?

packages/core/src/new-api/internal/deployer.ts
kanej: I have added an issue to expand out the number of validation messages returned: https://github.com/NomicFoundation/ignition/issues/424

packages/hardhat-plugin/src/ignition-helper.ts
kanej: What do you think of only showing the failed and timed out within the Ignition Error:

```
The deployment wan't succesful, there were timeouts and errors:

Timed out
  * SomeContract

Errors
  * MyContract: reverted with reason z
  * AnotherContract: reverted with reason y
```

packages/hardhat-plugin/test/calls.ts
```diff
@@ -155,9 +155,12 @@ describe("calls", () => {
       });
 
       const result = await this.deploy(moduleDefinition, {
-        Submodule: {
-          depositValue: BigInt(this.hre.ethers.parseEther("1")),
+        parameters: {
+          Submodule: {
```
alcuadrado, line 159: Good catch!

packages/core/src/new-api/internal/validation/validateStageTwo.ts
alcuadrado: Sounds right! We should have a small test for this though

packages/hardhat-plugin/src/index.ts
kanej: I think there is a rugged simplicity to stringified result.

packages/core/src/new-api/internal/validation/validateStageTwo.ts
kanej: Whoever does the validation multi-result work can add tests around this as well.

packages/hardhat-plugin/src/index.ts
kanej: I have added a basic console.log print out of results leveraging the same helper message as is used in the ignition helper exception code.

PR #427, opened 2023-08-29
fix: reconcile send data when undefined

### LINKED ISSUE #420: reconciliation fail after successful send
The reconciliation rules are too tight for send. A successful send on rerun fails in comparing the undefined data to '0x'

image

## Steps to reproduce

Run the complete example against a local node:

```sh
cd ./examples/complete
npx hardhat node
# in another window
npx hardhat deploy ./ignition/CompleteModule.js --network localhost
```

kanej:
Two runs where the send data is undefined now reconcile.

Fixes #420.


kanej:
One follow on issue is what to do about the error message display between a previous undefined and a now defined:

image

The reference to `0x` is confusing.

alcuadrado (approved):

PR #428, opened 2023-08-29
feat: reconciliation rule blocking timed out/failed futures

### LINKED ISSUE #412: Prevent resuming failed/timedout deployments
We need to add a check to the execution engine so that it fails to restart a failed/timedout deployment.

kanej:
If a future has previously timed out or failed (not just failed at simulation). Then the user must wipe the future with the wipe command before rerunning.

We enforce this with a reconciliation check. This check comes after the future level reconciliation checks (a design decision we may revisit - could it just be a standard rule?).

Resolves #412.

### DIFF
--- a/packages/core/src/new-api/internal/reconciliation/reconciler.ts
+++ b/packages/core/src/new-api/internal/reconciliation/reconciler.ts
@@ -1,9 +1,13 @@
+import { IgnitionError } from "../../../errors";
 import { ArtifactResolver } from "../../types/artifact";
 import { DeploymentParameters } from "../../types/deploy";
 import { Future, IgnitionModule } from "../../types/module";
 import { DeploymentLoader } from "../deployment-loader/types";
 import { DeploymentState } from "../new-execution/types/deployment-state";
-import { ExecutionState } from "../new-execution/types/execution-state";
+import {
+  ExecutionState,
+  ExecutionStatus,
+} from "../new-execution/types/execution-state";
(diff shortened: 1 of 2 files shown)


packages/core/src/new-api/internal/reconciliation/reconciler.ts
```diff
@@ -46,6 +50,10 @@ export class Reconciler {
       ]
     );
 
+    if (reconciliationFailures.length === 0) {
+      reconciliationFailures = this._reconcileNoErroredFutures(deploymentState);
+    }
+
     // TODO: Reconcile sender of incomplete futures.
```
alcuadrado, line 57:
```suggestion
```

Not relevant to this PR, but we already did this.

alcuadrado (approved):
I say we merge this here, and then think a bit more about the UX implications of this being part of reconciliation vs an ad-hoc check. The latter may allow us to give better error messages.

kanej:
> I say we merge this here, and then think a bit more about the UX implications of this being part of reconciliation vs an ad-hoc check. The latter may allow us to give better error messages.

Agreed. I will merge.

I initially had it as a post-reconciliation check. But I then needed to answer the question, "what is the return result?". I picked validation, then changed my mind and went with reconciliation. At that point though it seemed odd to have a post reconciliation check that returned a reconciliation failure, rather than an additional rule in reconciliation.

Happy to shift it out and improve the message, we just need to consider whether it should count as an additional phase and hence get a new return type.

Issue #426, opened 2023-08-29
Fix display of revert messages


kanej:
The decoding of the revert message shows a todo:

![image](https://github.com/NomicFoundation/ignition/assets/24030/010646ff-74a3-497d-b5f7-a2bfb4d91dab)

If you trigger a revert (as part of estimation say), instead of seeing the revert message you see a todo.

PR #429, opened 2023-08-29
Format the execution errors returned by deploy()

### LINKED ISSUE #410: Formatting a SolidityParameterType as string can throw
We have a function, [`solidityParamToString`](https://github.com/NomicFoundation/ignition/blob/316ec19b1df431a665e523cefd88ef4db7f0bf31/packages/core/src/new-api/internal/journal/utils/log.ts#L153), that turns a `SolidityParameterType` value as a string. This function can throw in cases where we have a recursive type with deeply nested bigints.

I think we and integrators are going to find ourselves in a similar situation many times, so we should have public helpers to do it.

I guess it would be something like:

```ts
function formatSolidityParameter(param: SolidityParameterType): string {
  if (Array.isArray(param)) {
    const values = param.map(solidityParamToString);

    return `[${values.join(", ")}]`;
  }

  if (typeof param === "object") {
    const values = Object.entries(param).map(
      ([key, value]) => `"${key}": ${solidityParamToString(value)}`
    );

    return `{${values.join(", ")}}`;
  }

  return param.toString();
}
```

alcuadrado:
Closes #410 and #426

### DIFF
--- a/packages/core/src/ui-helpers.ts
+++ b/packages/core/src/ui-helpers.ts
@@ -1,4 +1,3 @@
-/* eslint-disable import/no-unused-modules */
 export {
   StoredDeploymentDeserializer,
   StoredDeploymentSerializer,
@@ -6,3 +5,4 @@ export {
 export * from "./new-api/type-guards";
 export * from "./new-api/types/module";
 export * from "./new-api/types/serialized-deployment";
+export { formatSolidityParameter } from "./new-api/internal/formatters";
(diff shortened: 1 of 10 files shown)


packages/core/src/ui-helpers.ts
```diff
@@ -1,8 +1,8 @@
-/* eslint-disable import/no-unused-modules */
 export {
   StoredDeploymentDeserializer,
   StoredDeploymentSerializer,
 } from "./new-api/stored-deployment-serializer";
 export * from "./new-api/type-guards";
 export * from "./new-api/types/module";
 export * from "./new-api/types/serialized-deployment";
+export { formatSolidityParameter } from "./new-api/internal/formatters";
```
kanej, line 8: Just so I am clear on the reasoning.

Because this will inevitable be used in both the UI project and the cli, we move it to core and leverage it from there; as the alternative is duplication?

PR #439, opened 2023-09-07
Update event listener names

zoeyTM:


kanej (approved):

packages/core/src/types/execution-events.ts
```diff
@@ -422,13 +422,23 @@ export interface ExecutionEventTypeMap {
   [ExecutionEventType.SET_MODULE_ID]: SetModuleIdEvent;
 }
 
+/**
+ * A utility type for mapping enum values to function names
+ *
+ * @beta
+ */
+export type SnakeToCamelCase<S extends string> =
```
kanej, line 430: I am happy with this being in the public API, as opposed to the mapping type it replaced.

PR #456, opened 2023-09-11
feat: expose strategy in deploy's signature

### LINKED ISSUE #452: Enhance deploy signature to support strategy
Execution engine takes a execution strategy. This should be exposed through deploy, with a default of the basic strategy.

For consideration, do we need a reconcile step for strategy? This is so that, say a defender strategy, could check and throw if the detects a deployment that has previously run on an old version.

kanej:
Allow the user of deploy to provide their own strategy.

Resolves #452.


kanej:
We need to resolve the question of whether strategies are responsible for their own artifact loading or a loader of some form should be injected in.

kanej:
We are going to hold on exposing strategy through deploy.

Issue #451, opened 2023-09-11
Reintroduce Hold


kanej:
Our future status does not currently support `Hold`. We expect execution strategies to be leverage this feature in the near future, so we intend to add it to the types now.

PR #457, opened 2023-09-11
feat: add `Held` status to exState status

kanej:
To support execution strategies returning `hold` status in the future,
we are adding it into the execution status of exStates.

Resolves https://github.com/NomicFoundation/ignition/issues/451

## TODO

- [x] Add the status
- [x] Update the return result
- [x] Do we need a strategy message now that can return an on hold result?
- [x] Do we need rerun rules? Or can they be deferred until we release execution strategies.

### DIFF
--- a/packages/core/src/internal/deployer.ts
+++ b/packages/core/src/internal/deployer.ts
@@ -291,6 +291,26 @@ export class Deployer {
       successful: Object.values(deploymentState.executionStates)
         .filter((ex) => ex.status === ExecutionStatus.SUCCESS)
         .map((ex) => ex.id),
+      held: Object.values(deploymentState.executionStates)
+        .filter(canFail)
+        .filter((ex) => ex.status === ExecutionStatus.HELD)
+        .map((ex) => {
+          assertIgnitionInvariant(
+            ex.result !== undefined,
+            `Execution state ${ex.id} is marked as held but has no result`
+          );
+
+          assertIgnitionInvariant(
+            ex.result.type === ExecutionResultType.STRATEGY_HELD,
+            `Execution state ${ex.id} is marked as held but has ${ex.result.type} instead of a held result`
+          );
+
+          return {
+            futureId: ex.id,
+            heldId: ex.result.heldId,
+            reason: ex.result.reason,
+          };
+        }),
       timedOut: Object.values(deploymentState.executionStates)
         .filter(canTimeout)
         .filter((ex) => ex.status === ExecutionStatus.TIMEOUT)
@@ -304,7 +324,8 @@ export class Deployer {
         .map((ex) => {
           assertIgnitionInvariant(
             ex.result !== undefined &&
-              ex.result.type !== ExecutionResultType.SUCCESS,
+              ex.result.type !== ExecutionResultType.SUCCESS &&
+              ex.result.type !== ExecutionResultType.STRATEGY_HELD,
             `Execution state ${ex.id} is marked as failed but has no error result`
           );
 
--- a/packages/core/src/internal/execution/types/execution-result.ts
+++ b/packages/core/src/internal/execution/types/execution-result.ts
@@ -12,6 +12,7 @@ export enum ExecutionResultType {
   REVERTED_TRANSACTION = "REVERTED_TRANSACTION",
   STATIC_CALL_ERROR = "STATIC_CALL_ERROR",
   STRATEGY_ERROR = "STRATEGY_ERROR",
+  STRATEGY_HELD = "STRATEGY_HELD",
 }
 
 /**
@@ -60,6 +61,16 @@ export interface StrategyErrorExecutionResult {
 }
 
 /**
+ * The execution strategy returned a strategy-specific hold e.g.
+ * waiting for off-chain multi-sig confirmations.
+ */
+export interface StrategyHeldExecutionResult {
+  type: ExecutionResultType.STRATEGY_HELD;
+  heldId: number;
+  reason: string;
+}
+
+/**
  * A deployment was successfully executed.
  */
 export interface SuccessfulDeploymentExecutionResult {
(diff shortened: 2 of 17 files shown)


kanej:
I am opting to leave the run logic until we are actually exposing strategies that can use held.

packages/core/src/internal/execution/types/execution-result.ts
```diff
@@ -59,6 +60,16 @@ export interface StrategyErrorExecutionResult {
   error: string;
 }
 
+/**
+ * The execution strategy returned a strategy-specific hold e.g.
+ * waiting for off-chain multi-sig confirmations.
+ */
+export interface StrategyHeldExecutionResult {
```
kanej, line 67: The appropriate (failable) exstates can return a complete message where this Hold is the result.

packages/core/src/internal/execution/types/execution-result.ts
```diff
@@ -59,6 +60,16 @@ export interface StrategyErrorExecutionResult {
...
+  type: ExecutionResultType.STRATEGY_HELD;
+  heldId: number;
```
kanej, line 69: This is different from other examples of `Results`. We infer the networkInteractionId through the network interactions. But there is no other place where the heldId exists except within the logic of the strategy.

I have opted to include the heldId in the result, so it can be fed through to the results and UI.

packages/core/src/internal/deployer.ts
```diff
@@ -291,6 +291,26 @@ export class Deployer {
       successful: Object.values(deploymentState.executionStates)
         .filter((ex) => ex.status === ExecutionStatus.SUCCESS)
         .map((ex) => ex.id),
+      held: Object.values(deploymentState.executionStates)
+        .filter(canFail)
+        .filter((ex) => ex.status === ExecutionStatus.HELD)
+        .map((ex) => {
+          assertIgnitionInvariant(
+            ex.result !== undefined,
+            `Execution state ${ex.id} is marked as held but has no error result`
```
alcuadrado, line 300:
```suggestion
            `Execution state ${ex.id} is marked as held but has no result`
```

We should consider `held` a normal result, not an error.

packages/core/src/internal/deployer.ts
```diff
@@ -291,6 +291,26 @@ export class Deployer {
...
+          );
+
+          assertIgnitionInvariant(
+            ex.result.type === ExecutionResultType.STRATEGY_HELD,
+            `Execution state ${ex.id} is marked as held but has ${ex.result.type} instead of a held error result`
```
alcuadrado, line 305:
```suggestion
            `Execution state ${ex.id} is marked as held but has ${ex.result.type} instead of a held result`
```

packages/core/src/internal/deployer.ts
```diff
@@ -291,6 +291,26 @@ export class Deployer {
...
+          );
+
+          return {
+            futureId: ex.id,
+            heldId: ex.result.heldId,
+            reason: "The execution was held",
```
alcuadrado, line 311: These messages should come from the strategy

packages/core/src/internal/journal/utils/emitExecutionEvent.ts
```diff
@@ -217,6 +217,12 @@ function convertExecutionResultToEventResult(
         error: "Transaction reverted",
       };
     }
+    case ExecutionResultType.STRATEGY_HELD: {
```
alcuadrado, line 220: Sorry for being repetitive, but held shouldn't be considered an error.

packages/core/src/internal/journal/utils/emitExecutionEvent.ts
```diff
@@ -241,6 +247,12 @@ function convertStaticCallResultToExecutionEventResult(
         error: result.error,
       };
     }
+    case ExecutionResultType.STRATEGY_HELD: {
```
alcuadrado, line 250: Same here

packages/core/src/internal/views/has-execution-failed.ts
```diff
@@ -20,6 +20,7 @@ export function hasExecutionFailed(
 
   return (
     exState.status === ExecutionStatus.FAILED ||
-    exState.status === ExecutionStatus.TIMEOUT
+    exState.status === ExecutionStatus.TIMEOUT ||
+    exState.status === ExecutionStatus.HELD
```
alcuadrado, line 24: Same

packages/hardhat-plugin/src/utils/error-deployment-result-to-exception-message.ts
```diff
@@ -78,6 +79,14 @@ function _convertExecutionError(result: ExecutionErrorDeploymentResult) {
     sections.push(`Failures:\n\n${errorList.join("\n")}`);
   }
 
+  if (messageDetails.held) {
+    const errorList = result.held.map(
```
alcuadrado, line 83: Same

packages/core/src/internal/journal/utils/emitExecutionEvent.ts
kanej: I have added another event result (held), to allow mapping from held to a non-error result. And wired it through the UI.

packages/core/src/internal/views/has-execution-failed.ts
kanej: This func is used to determine whether the next batch should be processed or whether execution should halt and the result be returned.

I have swapped this function with a `has-execution-succeeded` function for use in the batch processing logic. This means we can consider `Held` not an error, but still consider it as something worth halting processing for, without distorting the meaning of `hasExecutionFailed`.

packages/hardhat-plugin/src/utils/error-deployment-result-to-exception-message.ts
kanej: I think this one is legitimate. An execution that is halted but not complete is error/failed. We need to be able to display failures/timeouts/held all at once.

Taking a look at the result type:

image

We could add another result type, but I think what we really want to do is rename `ExecutionErrorDeploymentResult` to something like `ExecutionHaltedDeploymentResult`?

alcuadrado: Sorry, I just meant the name of the variable `errorList` 😳

It could be `heldResons` or something like that. This is not important though.

alcuadrado (approved):

PR #462, opened 2023-09-12
Add support for return value indexing in `readEventArg` and `staticCall`

### LINKED ISSUE #396: Refine read event arg and static call api
Both read event arg and static call should be able to select the field of the result tuple by name or index.

Change readEventArgument parameter `argumentName` to be called `argument`, it should accept either string or number. Where the number is the index into the tuple.

We need to enhance staticCall so that there is an argument parameters after args. Argument is the number index for the result tuple.

Validation rules for out of range index and ensure that if the arg is a string that the name exists.

Take a look at the abi module, as some validation may already be in place.

zoeyTM:
fixes #396

### DIFF
--- a/packages/core/src/internal/reconciliation/futures/reconcileReadEventArgument.ts
+++ b/packages/core/src/internal/reconciliation/futures/reconcileReadEventArgument.ts
@@ -47,7 +47,7 @@ export function reconcileReadEventArgument(
 
   result = compare(
     future,
-    "Argument name",
+    "Argument name or index",
     executionState.nameOrIndex,
     future.nameOrIndex
   );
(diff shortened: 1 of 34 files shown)


packages/core/src/internal/reconciliation/futures/reconcileReadEventArgument.ts
```diff
@@ -48,8 +48,8 @@ export function reconcileReadEventArgument(
   result = compare(
     future,
     "Argument name",
```
alcuadrado, line 50:
```suggestion
    "Argument name or index",
```

alcuadrado: I guess we should add an equivalent reconciliation rule to static call

alcuadrado:
Good work! I left two minor comments about reconciliation

packages/core/src/internal/reconciliation/futures/reconcileReadEventArgument.ts
zoeyTM: Done and done

PR #469, opened 2023-09-14
Refactor previous run timeout/failure check

### LINKED ISSUE #431: Reconsider previous run timeout/error check
Currently we ban a rerun if a future is in the timeout state or error state. This is done as a special phase within reconciliation. All other reconciliations are run first, then the final guard is whether the future had already started and so the state needs wiped.

Can this:
* become a normal reconciliation rule? Probably not
* be moved out as a post-reconciliation check, that either exceptions or returns its own deployment result type

The goal here is to enable the best error message to the user for the problem.

zoeyTM:
- Moved logic for checking for previous run errors out of reconciliation flow
- Created a new `DeploymentResult` type to encapsulate the previous run errors
- Updated the UI to display more information to the user for the new `DeploymentResult` type (see screenshot)

fixes #431 

Screen Shot 2023-09-13 at 8 10 59 PM

### DIFF
--- a/packages/core/src/internal/deployer.ts
+++ b/packages/core/src/internal/deployer.ts
@@ -1,6 +1,5 @@
 import type { IgnitionModule, IgnitionModuleResult } from "../types/module";
 
-import { IgnitionError } from "../errors";
 import { isContractFuture } from "../type-guards";
 import { ArtifactResolver } from "../types/artifact";
 import {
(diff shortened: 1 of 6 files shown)


zoeyTM:
I wanted to mention, while this works for now, I think that we may want to revisit this in the future potentially and consider enriching the existing ReconciliationError type with more data for the user instead of creating a new type like we've done here. Maybe a subtype on the ReconciliationError type? 

My thought is that this is not the only "reconciliation error" that we may want to provide more detailed info about to the user in the future.

alcuadrado:
LGTM

Re your comment @zoeyTM: do you think that can be done in a backward compatible way in the future? For example, we may pass a flag to `deploy/plan` so that it returns more "structured" errors instead of `string`s, or expose a method to get the errors.

alcuadrado (approved):

PR #471, opened 2023-09-14
feat: rename plan to visualize

### LINKED ISSUE #450: Refactor plan return type
Make plan return results similar to deploy. The intention here is to support rerunning plan after a deployment run and getting a visual indication of progress by adding type changes in the public api now.

## TODO

- [ ] Rename plan to `visualize` (or show)?
- [ ] Change the result type to match as much of deploy as makes sense (validation fails, reconciliation fails, execution results)
- [ ] After the plan result types have been recorded we want to review the names of the serialization types, and whether we can drop additional information (like chainId)

As a follow on we intend to enhance the plan features to support showing execution information, this is captured in #449.

kanej:
We intend to support viewing an already run deployment, hence plan makes less sense. Instead we will go with the more generic visualize for the moment.

Resolves #450 

## TODO

- [x] Rename plan to `visualize` (or show)?
- ~Change the result type to match as much of deploy as makes sense (validation fails, reconciliation fails, execution results)~
- [x] After the plan result types have been recorded we want to review the names of the serialization types, and whether we can drop additional information (like chainId)


kanej:
We have opted to not export visualize from core for the moment. Instead the visualize task will only leverage the `IgnitionModule` and the serializer.

The serializer has been tightened up to only be about IgnitionModules.

alcuadrado (approved):

PR #473, opened 2023-09-14
feat: store debug files with build infos

kanej:
If a build info is stored for a future id, record a debug file as a pointer. The debug file is a sibling to the artifact file, so under `./ignition/deployments/<depid>/artifacts`:
    
- `LockModule#Lock.json`
- `LockModule#Lock.dbg.json`
    
The dbg file points at the Ignition build info file via a relative path.

### DIFF
--- a/packages/core/src/internal/deployment-loader/file-deployment-loader.ts
+++ b/packages/core/src/internal/deployment-loader/file-deployment-loader.ts
@@ -81,7 +81,10 @@ export class FileDeploymentLoader implements DeploymentLoader {
     await writeFile(artifactFilePath, JSON.stringify(artifact, undefined, 2));
   }
 
-  public async storeBuildInfo(buildInfo: BuildInfo): Promise<void> {
+  public async storeBuildInfo(
+    futureId: string,
+    buildInfo: BuildInfo
+  ): Promise<void> {
     await this._initialize();
 
     const buildInfoFilePath = path.join(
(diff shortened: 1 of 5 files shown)


packages/core/src/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -90,6 +93,23 @@ export class FileDeploymentLoader implements DeploymentLoader {
...
+      this._paths.artifactsDir,
+      `${futureId}.dbg.json`
+    );
+
+    await writeFile(
+      debugInfoFilePath,
+      JSON.stringify(
+        {
+          _format: "hh-sol-dbg-1",
+          buildInfo: `../build-info/${buildInfo.id}.json`,
```
kanej, line 107: This path could be calculated relatively with something like `path.relative`. That would get you an OS specific format. Is that what is expected in HH debug files?

alcuadrado: It's currently computed with:

```ts
    const relativePathToBuildInfo = path.relative(
      path.dirname(artifactPath),
      pathToBuildInfo
    );
```

alcuadrado: which would be

```ts
    const relativePathToBuildInfo = path.relative(
      this._paths.artifactsDir,
      debugInfoFilePath,
    );
```

packages/core/src/internal/deployment-loader/types.ts
```diff
@@ -19,7 +19,7 @@ export interface DeploymentLoader {
     contractName: string,
     artifact: Artifact
   ): Promise<void>;
-  storeBuildInfo(buildInfo: BuildInfo): Promise<void>;
+  storeBuildInfo(futureId: string, buildInfo: BuildInfo): Promise<void>;
```
kanej, line 22: To track the link between future and build info, the `storeBuildInfo` is enhanced with the future id.

Note we are calling store build info multiple times if more than one artifact has the same build info. But that is no different than before this change.

alcuadrado: Makes sense. We can optimize it later if needed.

alcuadrado (approved):
Left a small comment but LGTM

packages/core/src/internal/deployment-loader/file-deployment-loader.ts
kanej: Awesome. Change applied.

PR #476, opened 2023-09-18
Support returning multiple validation errors (PR 1)

### LINKED ISSUE #424: support multiple validation error messages displayed at once
Currently our validation during stage 2 limits the number of validation errors to one per deployment. The types support more. We should rework the validtion so that multiple fails can be retrieved at once.

zoeyTM:
fixes #424


kanej:
@zoeyTM can you walk me through our validation stage 1 vs validation stage 2 division. If we swap in full reconciliation/execution info from visualization does this distinction go away?

zoeyTM:
@kanej I believe the answer is yes. Originally, we separated validation into stages so that stage one could be used by the visualization task, but the only difference between the stages is the inclusion of execution info such as accounts and deployment parameters. If we include that info in visualization like we plan to, then I don't believe there's any reason for the stages to be separate anymore. 

@alcuadrado may have a different take, but that's my perspective at least

kanej:
> @kanej I believe the answer is yes. Originally, we separated validation into stages so that stage one could be used by the visualization task, but the only difference between the stages is the inclusion of execution info such as accounts and deployment parameters. If we include that info in visualization like we plan to, then I don't believe there's any reason for the stages to be separate anymore.
> 
> @alcuadrado may have a different take, but that's my perspective at least

Ugh. I suspected this was the case. Alright, let me take another look over it. Do you have a feel for whether that merge would be better done as a separate PR?

zoeyTM:
> > @kanej I believe the answer is yes. Originally, we separated validation into stages so that stage one could be used by the visualization task, but the only difference between the stages is the inclusion of execution info such as accounts and deployment parameters. If we include that info in visualization like we plan to, then I don't believe there's any reason for the stages to be separate anymore.
> > @alcuadrado may have a different take, but that's my perspective at least
> 
> Ugh. I suspected this was the case. Alright, let me take another look over it. Do you have a feel for whether that merge would be better done as a separate PR?

Short answer: I think a separate PR would be better.

I was going to say it should be done after we add the execution data to visualize so we don't break it, but it doesn't actually look like we're validating in that task at all anymore. Given that, I think merging the validation stages themselves would be relatively trivial, so most of the work would just be modifying the tests.

alcuadrado (approved):

PR #478, opened 2023-09-19
Support using the result from `m.staticCall` or `m.readEventArgument` as value option

### LINKED ISSUE #455: Expand value and from to support staticCall/readEventArg
We should expand the options of the `value` property on the actions of the api to include `staticCall` and `readEventArg`.

zoeyTM:
fixes #455

### DIFF
--- a/packages/core/src/internal/execution/future-processor/helpers/future-resolvers.ts
+++ b/packages/core/src/internal/execution/future-processor/helpers/future-resolvers.ts
@@ -12,6 +12,7 @@ import {
   ContractFuture,
   Future,
   ModuleParameterRuntimeValue,
+  ReadEventArgumentFuture,
   SolidityParameterType,
   StaticCallFuture,
 } from "../../../../types/module";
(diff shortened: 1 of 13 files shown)


packages/core/src/internal/execution/future-processor/helpers/future-resolvers.ts
```diff
@@ -30,23 +35,33 @@ import { convertEvmValueToSolidityParam } from "../../utils/convert-evm-tuple-to
...
+    result = resolveModuleParameter(givenValue, {
+      deploymentParameters,
+    });
+  }
 
   assertIgnitionInvariant(
-    typeof moduleParam === "bigint",
-    "Module parameter used as value must be a bigint"
+    typeof result === "bigint",
+    "Module parameter or future result used as value must be a bigint"
```
kanej, line 61: Nice catch.

kanej (approved):
LGTM. Nicely done.

PR #479, opened 2023-09-20
Centralize error descriptions (PR 1)

### LINKED ISSUE #418: Centralise error descriptions
Follow the pattern in Hardhat where we have one place where error description are defined so they can be reviewed in one place and intentionally improved as error reports come in.

zoeyTM:
- add `errors-list.ts`, mimicking the similar structure in hardhat
- centralized all currently thrown error messages into the above structure
- adjusted `IgnitionError` to require one of the new `ErrorDescriptor` types as an input, rather than a raw string
- removed `IgnitionValidationError` as a class since its stack capturing is no longer functioning as originally intended 
- refactored all previous instances of `IgnitionValidationError` into `IgnitionError`

fixes #418

### DIFF
--- a/packages/core/src/errors-list.ts
+++ b/packages/core/src/errors-list.ts
@@ -1,0 +1,65 @@
+export const ERROR_PREFIX = "IGN";
+
+export interface ErrorDescriptor {
+  number: number;
+  // Message can use templates. See applyErrorMessageTemplate
+  message: string;
+}
+
+export function getErrorCode(error: ErrorDescriptor): string {
+  return `${ERROR_PREFIX}${error.number}`;
+}
+
+export const ERROR_RANGES: {
+  [category in keyof typeof ERRORS]: {
+    min: number;
+    max: number;
+    title: string;
+  };
+} = {
+  GENERAL: {
+    min: 1,
+    max: 99,
+    title: "General errors",
+  },
+  INTERNAL: {
+    min: 100,
+    max: 199,
+    title: "Internal errors",
+  },
+};
+
+export const ERRORS = {
+  GENERAL: {
+    TEST: {
+      number: 1,
+      message: "Test error",
+    },
+  },
+  INTERNAL: {
+    TEMPLATE_INVALID_VARIABLE_NAME: {
+      number: 100,
+      message: "Invalid variable name: %variable%",
+    },
+    TEMPLATE_VARIABLE_NOT_FOUND: {
+      number: 101,
+      message: "Variable not found: %variable%",
+    },
+    TEMPLATE_VALUE_CONTAINS_VARIABLE_TAG: {
+      number: 102,
+      message: "Value contains variable tag: %variable%",
+    },
+  },
+};
+
+/**
+ * Setting the type of ERRORS to a map let us access undefined ones. Letting it
+ * be a literal doesn't enforce that its values are of type ErrorDescriptor.
+ *
+ * We let it be a literal, and use this variable to enforce the types
+ */
+const _PHONY_VARIABLE_TO_FORCE_ERRORS_TO_BE_OF_TYPE_ERROR_DESCRIPTOR: {
+  [category: string]: {
+    [name: string]: ErrorDescriptor;
+  };
+} = ERRORS;
--- a/packages/core/src/errors-list.ts
+++ b/packages/core/src/errors-list.ts
@@ -52,6 +52,11 @@ export const ERROR_RANGES: {
     max: 699,
     title: "Wipe errors",
   },
+  VALIDATION: {
+    min: 700,
+    max: 799,
+    title: "Validation errors",
+  },
 };
 
 export const ERRORS = {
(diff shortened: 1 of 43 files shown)


zoeyTM:
One open question for this PR is whether or not we want to add an `IgnitionPluginError` and/or `NomicIgnitionPluginError` similar to what hardhat has.

zoeyTM:
> One open question for this PR is whether or not we want to add an `IgnitionPluginError` and/or `NomicIgnitionPluginError` similar to what hardhat has.

Okay I added these and a new question has popped up: should our `hardhat-plugin` throw `NomicIgnitionPluginError` or `NomicLabsHardhatPluginError` (or `HardhatPluginError`)? Technically it's both an Ignition plugin and a Hardhat plugin, so I'm not quite sure

packages/core/src/errors-list.ts
```diff
@@ -0,0 +1,322 @@
+export const ERROR_PREFIX = "IGN";
```
zoeyTM, line 1: Most of this file is basically copied structurally from the similar files in Hardhat.

packages/core/src/errors-list.ts
```diff
@@ -0,0 +1,322 @@
...
+ * ErrorDescriptor is a type that describes an error.
+ * It's used to generate error codes and messages.
+ *
+ * @beta
+ */
+export interface ErrorDescriptor {
+  number: number;
+  // Message can use templates. See applyErrorMessageTemplate
+  message: string;
+}
```
zoeyTM, line 13: I think Hardhat has extra fields here, but I figured we start with the minimum and can always add more later.

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
 /**
- * All exceptions intenionally thrown with Ignition-core
- * extend this class.
+ * Base error class extended by all custom errors.
+ * Placeholder to allow us to customize error output formatting in the future.
  *
  * @alpha
  */
-export class IgnitionError extends Error {
+export class CustomError extends Error {
```
zoeyTM, line 9: Hardhat does a lot of clever stuff to modify the displayed output of errors inside this class. We currently aren't doing those things, but I wanted to leave room for us to add it later if we decide to.

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
- * This error class represents issue detected by Ignition-cores
- * validation phase on the user inputed module. Validation errors
- * capture the stack to the action within the offending module,
- * to enhance the locality of the validation error message.
+ * This class is used to throw errors from Ignition plugins made by third parties.
  *
- * * @alpha
+ * @alpha
  */
-export class IgnitionValidationError extends IgnitionError {
```
zoeyTM, line 23: I removed the validation specific error in favor of a validation category in `errors-list.ts`. All errors thrown in Ignition core are now `IgnitionError`s

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
-   */
-  public resetStackFrom(f: () => any) {
-    Error.captureStackTrace(this, f);
+/**
+ * This class is used to throw errors from *core* Ignition plugins.
+ * If you are developing a third-party plugin, use IgnitionPluginError instead.
+ *
+ * @alpha
+ */
+export class NomicIgnitionPluginError extends IgnitionPluginError {
```
zoeyTM, line 72: I added this and the above `IgnitionPluginError` again mimicking the similar classes from Hardhat.

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
+ *  - Values can't contain variable tags.
+ *  - If a variable is not present in the template, but present in the values
+ *    object, an error is thrown.
+ *
+ * @param template - The template string.
+ * @param values - A map of variable names to their values.
  *
  * @alpha
  */
-export class UnsupportedOperationError extends IgnitionError {
```
zoeyTM, line 54: This error was only being used in one place, so I removed it as well and added the message for it to `errors-list.ts`

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
-  constructor(message: string) {
-    super(message);
+export function applyErrorMessageTemplate(
```
zoeyTM, line 108: copy/pasted from Hardhat

packages/core/src/errors.ts
```diff
@@ -1,58 +1,176 @@
...
+  }
+
+  private readonly _isIgnitionPluginError = true;
+
+  public readonly pluginName: string;
+
+  constructor(pluginName: string, message: string) {
     super(message);
+    this.pluginName = pluginName;
+    Object.setPrototypeOf(this, IgnitionPluginError.prototype);
```
alcuadrado, line 62: Can we leave the comment that was here before? also, `new.target` is dynamic, so it works with extensions to this tupe.

alcuadrado (approved):

PR #482, opened 2023-09-20
fix: display all deployed contracts (PR 3)

### LINKED ISSUE #480: deployed addresses only show top level module
Take the `SubmodulesExample` from the ignition examples repo:

image

We are not listing all the deployed contracts in the results section. Only the contracts deployed by the top level module.

kanej:
The result was constrained to only those contracts returned by the module. It now returns all contracts and contractAts that have been deployed in this run and previous runs.

The `ignition-helper` now takes the responsibility of resolving the modules results to contracts (and only the result contracts).

Note the types have not been altered.

Resolves #480.

### DIFF
--- a/examples/ens/.gitignore
+++ b/examples/ens/.gitignore
@@ -7,3 +7,4 @@ typechain
 #Hardhat files
 cache
 artifacts
+ignition/deployments
--- a/packages/core/src/deploy.ts
+++ b/packages/core/src/deploy.ts
@@ -58,7 +58,7 @@ export async function deploy<
   deploymentParameters: DeploymentParameters;
   accounts: string[];
   defaultSender?: string;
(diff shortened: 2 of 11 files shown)


kanej:
I am unsure of this change. Currently deploy returns a deployment result. That deployment result type encodes the return results of the module passed in.

That misled us, I think the deployed contracts should include the contracts, librarys and contractsAts that are deployed in the module and its submodules.

This PR expands the returned results to include those contracts in submodules. However, I have not changed the type.

I think there are two approaches we can consider:

* return all the contracts (but don't change the type) - as this PR does
* change the type of the deploy result to not encode the module return, but insist on encoding it for the `deploy` function of the `ignition-helper`

alcuadrado (approved):

PR #483, opened 2023-09-20
Lazy load external packages within plugin (PR 2)

### LINKED ISSUE #13: Load Ignition core lazily in the hardhat plugin
Ignition uses some heavy packages (ink, react), therefore the hardhat plugin shouldn't load Ignition directly. Instead, it should be lazy-loaded when needed.

zoeyTM:
fixes #13

### DIFF
--- a/packages/hardhat-plugin/src/index.ts
+++ b/packages/hardhat-plugin/src/index.ts
@@ -1,23 +1,10 @@
+import type { DeploymentParameters } from "@nomicfoundation/ignition-core";
+
 import "@nomicfoundation/hardhat-ethers";
-import {
-  deploy,
-  DeploymentParameters,
-  IgnitionModuleSerializer,
-  wipe,
-} from "@nomicfoundation/ignition-core";
 import { existsSync, readdirSync, readJSONSync } from "fs-extra";
 import { extendConfig, extendEnvironment, task } from "hardhat/config";
 import { lazyObject } from "hardhat/plugins";
 import path from "path";
-import Prompt from "prompts";
-
-import { HardhatArtifactResolver } from "./hardhat-artifact-resolver";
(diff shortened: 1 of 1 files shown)


fvictorio:
I tested this locally and `npx hardhat` goes from taking 500ms to taking 2000ms when including the `hardhat-ignition` plugin.

I suspect the cause is this one:

```typescript
export { buildModule } from "@nomicfoundation/ignition-core";
```

I'm not 100% sure what's the best way to do this "dynamic". This monstrosity seems to compile fine:

```typescript
export const buildModule: typeof import("@nomicfoundation/ignition-core").buildModule =
  (...args) => {
    const { buildModule: coreBuildModule } =
      require("@nomicfoundation/ignition-core") as typeof import("@nomicfoundation/ignition-core");

    return coreBuildModule(...args);
  };
```

fvictorio:
Ugh, I tried that using `main`, sorry, I'll give it another try in this branch -_-

fvictorio:
Ok, these are the proper measurements:

1. `npx hardhat` without the ignition plugin: 500ms
2. `npx hardhat` with the current version of ignition: 2000ms
3. `npx hardhat` with these changes: 1500ms
4. `npx hardhat` with these changes plus the change in my comment above: 550ms

So yeah, we should make that change (or an equivalent one)

kanej:
@fvictorio, @schaable Would the eslint plugin work for our repo?

fvictorio:
I think so. We should make some changes and publish it.

fvictorio:
Opened https://github.com/NomicFoundation/hardhat/issues/4423

alcuadrado:
I think this looks good enough. Let's merge it, and then we revisit it once the slow imports eslint plugin gets published.

alcuadrado (approved):

PR #487, opened 2023-09-21
feat: put in place alternate advanced UI (PR 4)

kanej:
The simple text UI is intended to match the ink version without ink. So we would have a text ui that shows batching and a pretty version of results, and a verbose version that matches a more logging orientated output.

## Preview

new-ui

## Method

I have copied the ink based UI handler, but for the moment replace the simple ui handler at the command line, this allows for quick comparison between a deploy with the current ink based ui and the new non-ink based ui.

The first step is replace all of the result views with close equivalents based on console logging and chalk.

The next step is replacing the batching (based on the captured event info in the handler state), maybe to something like:

```shell
batch #0

- MyModule#Contract1 deploying...
- MyModule#Contract.call running
- OtherModule#Contract3 deploying...
```

Then rewriting the set of lines on updates to futures in the batch?

Once we are happy with this we swap the ink based handler out for the new handler and bring back the verbose handler.

## TODO

- [x] Replace successful deployment result view
- [x] Replace validation error result view
- [x] Replace reconciliation error result view
- [x] Replace rerun error result view
- [x] Replace execution error result view
- [x] Replace batching display
- [x] Remove old ui and package deps


alcuadrado (approved):
I mostly reviewed the structure and it looks good and simple. Let's merge this

PR #497, opened 2023-09-27
Update visualization UI styles (PR 5)

### LINKED ISSUE #493: Apply new styling based on figma
See the figma doc for more details:

image

## TODO

- [x] Rework futures list in to batches
- [x] Implement linking between futures in batches
- [ ] Implement highlighting of diagram on hover of future in batch
- [x] Rename section to Execution batches
- [ ] The summary should show the contract name being deployed, and an `x4` after it to indicate the number of instances
- [x] The tooltip should explain the concept of `futures`
- [x] The different batches buttons on the diagram, on hover of batch number should highlight in the mermaid diagram the futures that make up that batch.

zoeyTM:
fixes #493


zoeyTM:
progress pics

Screen Shot 2023-09-26 at 9 36 58 PM
Screen Shot 2023-09-26 at 9 37 09 PM

packages/ui/package.json
```diff
@@ -14,6 +14,7 @@
     "prepack": "npm run build"
   },
   "dependencies": {
+    "@fontsource-variable/inter": "^5.0.8",
```
kanej, line 17: Are we pulling in this font as it is an exact match for figma?

Do we have the option of inlining it - rather than pulling in a dep?

zoeyTM: @kanej this font is the one from figma, yes. I did some light research into importing fonts with vite and this seemed to be the easiest solution I found.

kanej (approved):
We will need another round of tweaks and fixes, but that is coming in as a follow on PR.

PR #499, opened 2023-09-28
Add batching information to visualize (UI PR 3)

### LINKED ISSUE #494: Add batching information
Display the batch information within the list.

Rename the futures list to `Execution batches`, then group the futures by batch with a title on each batch.

## Design

image

## Technical

We need to a `visualize` api in core. The end result is going to be an end point that returns the result type as deploy (maybe called status). For the moment though it should just return the batch information.

zoeyTM:
fixes #494


zoeyTM:
@kanej everything is connected and working on the UI side. Should be able to just plug in the data from core and have it Just Work™️

kanej (approved):

PR #503, opened 2023-09-30
feat/status-command (PR 6)

### LINKED ISSUE #496: Rename ignition-info to status
Rename the `ignition-info` task to status.

We also want to display `validation`/`reconciliation`/`timeouts`/`execution` failures in addition to the successful deployment addresses.

kanej:
Rename the `ignition-info` command to status and have it display errored and timed out futures as well as the deployed contracts list.

Resolves #496


packages/hardhat-plugin/src/index.ts
```diff
@@ -227,55 +232,106 @@ task("visualize")
     }
   );
 
-task("ignition-info")
-  .addParam("deploymentId")
-  .addFlag("json", "format as json")
```
kanej, line 232: I have opted to remove the json option as status does more than just show the deployed addresses now.

packages/hardhat-plugin/src/index.ts
```diff
@@ -227,55 +232,106 @@ task("visualize")
...
+      "deployments",
+      id
+    );
+
+    const statusResult = await status(deploymentDir);
+
+    if (statusResult.started.length > 0) {
+      console.log("");
+      console.log(
+        chalk.bold(`⛔ Deployment ${id} exited early and needs to be rerun`)
```
kanej, line 250: If there is a simulation error, we don't update the future state, we kill the process leaving started futures.

If you call status on a deployment after the simulation error you will see this message.

packages/hardhat-plugin/src/index.ts
alcuadrado: Agreed with this

packages/hardhat-plugin/src/index.ts
alcuadrado: How is this different from the `"There are futures that were started but have not completed:"`?

alcuadrado: What I mean is that they feel duplicated

packages/hardhat-plugin/src/index.ts
```diff
@@ -227,55 +229,108 @@ task("visualize")
...
+    }
+
+    if (
+      statusResult.timedOut.length > 0 ||
+      statusResult.failed.length > 0 ||
+      statusResult.held.length > 0
+    ) {
+      console.log("");
+      console.log(
+        chalk.bold(`⛔ Deployment ${id} has failed or halted futures`)
```
alcuadrado, line 275: I think using the word halted here can be confusing, as it's the only place we use it.

alcuadrado (approved):
I left some small comments but approved otherwise

packages/hardhat-plugin/src/index.ts
kanej: That is a good point. How about "Deployment XXX has either failed or timed out". Or do you think we do a case statement on the text depending on the combination of failed/timed out/held?

packages/hardhat-plugin/src/index.ts
kanej: That is interesting. I was trying to sign post the user in the case where there was a simulation error; which I think is likely to be the most common case here. `There are futures that were started but have not completed`, is accurate but obscure for that case is my suspicion.

packages/hardhat-plugin/src/index.ts
kanej: I added a case statement.

packages/hardhat-plugin/src/index.ts
kanej: I have updated the text and combined the two separate lines into one.

PR #505, opened 2023-10-02
fix/additional-nonces-checks (PR 7)

### LINKED ISSUE #411: Not all accounts' nonces are sync before execution
Before starting to execute a module we run [`getNonceSyncMessages`](https://github.com/NomicFoundation/ignition/blob/316ec19b1df431a665e523cefd88ef4db7f0bf31/packages/core/src/new-api/internal/new-execution/nonce-management.ts#L42C23-L42C43) to sync the nonces from the previous with the actual ones, and to prevent us from execution if there are user transactions without enough confirmations.

We only check the accounts that we have already used, but we have to check every transaction that's either a `from` in a future, or the default sender.

kanej:
Include the future senders from the module when syncing nonces for checks. 

The senders for all futures without exStates that will be used if the module executes, are now added to the syncing taking into account usage of default sender when `from` is undefined.

Resolves #411

### DIFF
--- a/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
+++ b/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
@@ -1,16 +1,14 @@
-import { IgnitionError } from "../../errors";
-import { ERRORS } from "../../errors-list";
-import { getPendingNonceAndSender } from "../views/execution-state/get-pending-nonce-and-sender";
-import { getPendingOnchainInteraction } from "../views/execution-state/get-pending-onchain-interaction";
-
-import { JsonRpcClient } from "./jsonrpc-client";
-import { DeploymentState } from "./types/deployment-state";
-import { ExecutionSateType } from "./types/execution-state";
+import { IgnitionError } from "../../../errors";
+import { ERRORS } from "../../../errors-list";
+import { getPendingOnchainInteraction } from "../../views/execution-state/get-pending-onchain-interaction";
+import { JsonRpcClient } from "../jsonrpc-client";
+import { DeploymentState } from "../types/deployment-state";
+import { ExecutionSateType } from "../types/execution-state";
 import {
   JournalMessageType,
   OnchainInteractionDroppedMessage,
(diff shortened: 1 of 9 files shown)


packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
```diff
@@ -279,3 +242,43 @@ function createMapFromSenderToNonceAndTransactions(
...
+function _resolveFutureSenders(
+  ignitionModule: IgnitionModule<string, string, IgnitionModuleResult<string>>,
+  accounts: string[],
+  defaultSender: string,
+  exStateIds: string[]
+): string[] {
+  const futures = getFuturesFromModule(ignitionModule);
+
+  const senders: string[] = futures
+    .filter((f) => !exStateIds.includes(f.id))
```
kanej, line 259: Exclude existing exStates. They get dealt with by a different path.

packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
```diff
@@ -279,3 +242,43 @@ function createMapFromSenderToNonceAndTransactions(
...
+
+  return uniq(senders);
+}
+
+function _pickFrom(
+  future: Future,
+  accounts: string[],
+  defaultSender: string
+): string | null {
+  if (isNamedContractAtFuture(future)) {
```
kanej, line 271: I took this pattern as opposed to a sophisticated filter higher up as it gives nicer typing under conditions where we change futures that breaks the `f.from` here.

packages/core/src/type-guards.ts
```diff
@@ -170,7 +170,7 @@ export function isNamedStaticCallFuture(
  */
 export function isReadEventArgumentFuture(
   future: Future
-): future is StaticCallFuture<string, string> {
+): future is ReadEventArgumentFuture {
```
kanej, line 173: I assume this has just never been used.

alcuadrado (approved):

PR #507, opened 2023-10-02
fail gracefully if program to open visualize output is not available

### LINKED ISSUE #504: visualize fails if xdg-open or equivalent is missing
I'm getting this bug

```
$ npx hardhat 
visualize ignition/BasicDeploymentsModule.js
Deployment visualization written to /workspaces/ignition-example/cache/visualization/index.html
/bin/sh: 1: xdg-open: not found
An unexpected error occurred:

Error: Command failed: xdg-open /workspaces/ignition-example/cache/visualization/index.html
/bin/sh: 1: xdg-open: not found
```

We should check if the utility to open the HTML file exists before running it

zoeyTM:
fixes #504

### DIFF
--- a/packages/hardhat-plugin/src/utils/open.ts
+++ b/packages/hardhat-plugin/src/utils/open.ts
@@ -17,8 +17,6 @@ export function open(filePath: string): void {
   try {
     execSync(`${command} ${filePath}`, { stdio: "ignore" });
   } catch {
-    console.log(
-      "No program found to open the file. Please open the file manually in your browser."
-    );
+    // do nothing
   }
 }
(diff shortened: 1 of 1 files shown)


zoeyTM:
Screen Shot 2023-10-02 at 4 11 25 PM

packages/hardhat-plugin/src/utils/open.ts
```diff
@@ -14,5 +14,11 @@ export function open(filePath: string): void {
       command = "xdg-open";
   }
 
-  execSync(`${command} ${filePath}`);
+  try {
+    execSync(`${command} ${filePath}`, { stdio: "ignore" });
+  } catch {
+    console.log(
+      "No program found to open the file. Please open the file manually in your browser."
+    );
```
alcuadrado, line 22: What do you think about this alternative? As we are already printing the path, the user will know how to open it, and there's no clear expectation of auto-opening it.

```suggestion
```

zoeyTM: Yep, sounds good to me!

PR #508, opened 2023-10-03
fix: nonce-syncing check of mempool transactions (PR 8)

### LINKED ISSUE #506: rerun bug
On a rerun with an inflight transaction we are getting a false error message about the user sending a transaction on the same account.

Originally reported by Pato.

kanej:
There was a wrong check on whether the transactions against the last network interaction are still in the mempool. If at least one transaction is in the mempool we don't do the further checks that assume the transaction with that nonce has been either dropped or replaced by the user.

I have added a test for the failing case of rerun at the command line.

Fixes #506.

### DIFF
--- a/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
+++ b/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
@@ -117,12 +117,29 @@ export async function getNonceSyncMessages(
         transactions.map((tx) => jsonRpcClient.getTransaction(tx))
       );
 
-      if (fetchedTransactions.some((tx) => tx === undefined)) {
+      // If at least one transaction for the future is still in the mempool,
+      // we do nothing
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
```diff
@@ -117,12 +117,29 @@ export async function getNonceSyncMessages(
         transactions.map((tx) => jsonRpcClient.getTransaction(tx))
       );
 
-      if (fetchedTransactions.some((tx) => tx === undefined)) {
+      // If at least one transaction for the future is still in the mempool,
+      // we do nothing
+      if (fetchedTransactions.some((tx) => tx !== undefined)) {
```
kanej, line 122: The fix

alcuadrado (approved):
Approved, will get deeper into the tests in the following pr

PR #512, opened 2023-10-03
feat/additional nonce sync check (PR 9)

kanej:
If there are pending Ignition transactions, we still want to check if the user has sent additional transactions on top of that.

A test has been added to cover the case as well.


alcuadrado (approved):
Approved and added a few commits

PR #517, opened 2023-10-05
feat/scoped-tasks (PR 1)

### LINKED ISSUE #442: update cli tasks to use Hardhat namespacing
Hardhat namespacing of tasks is about to drop in hardhat.

Ignition should leverage the namespacing feature, so that our command set becomes:

```shell
npx hardhat ignition deploy ./ignition/MyModule.ts
npx hardhat ignition plan ./ignition/MyModule.ts
npx hardhat ignition info ./ignition/MyModule.ts
```

We should also review the parameters that ignition accepts for each task.

## TODO

- [ ] Switch commands to use an ignition namespace
- [ ] Review parameter names and defaults

kanej:
Scope all of Ignition's tasks under the `ignition` keyword.

To allow this we now depend on hardhat v2.18.0 and above.

The flags and parameters have been rationalized as well, and missing descriptions added.

Specifically:

* deploy has had its `force` flag removed as it is no longer used.
* wipe has had the param deployment changed to `id` to match the usage in our other tasks

Resolves #442.

## Preview

image


packages/hardhat-plugin/src/index.ts
```diff
@@ -173,9 +178,14 @@ task("deploy")
     }
   );
 
-task("visualize")
+ignitionScope
+  .task("visualize")
   .addFlag("quiet", "Disables logging output path to terminal")
```
kanej, line 183: We could remove this. It doesn't seem to be providing a huge value add. It is inherited from a previous iteration.

packages/hardhat-plugin/src/index.ts
```diff
@@ -15,6 +15,8 @@ import "./type-extensions";
 /* ignition config defaults */
 const IGNITION_DIR = "ignition";
 
+const ignitionScope = scope("ignition", "Deploy contracts");
```
alcuadrado, line 18:
```suggestion
const ignitionScope = scope("ignition", "Deploy your smart contracts using Hardhat Ignition");
```

packages/hardhat-plugin/src/index.ts
```diff
@@ -54,18 +56,22 @@ extendEnvironment((hre) => {
...
+    "The name of the module file within the Ignition modules directory, or a path to the module file"
+  )
   .addOptionalParam(
     "parameters",
-    "A JSON object as a string, of the module parameters, or a relative path to a JSON file"
+    "A relative path to a JSON file to use for the module parameters,"
   )
   .addOptionalParam("id", "set the deployment id")
-  .addFlag("force", "restart the deployment ignoring previous history")
   .addFlag(
```
alcuadrado, line 70: I think we should remove this flag for now. It's unclear if it adds value when compared to the other UI. If we get complains about the normal UI, we can consider reintroducing it.

To be clear, I'd just remove the flag

alcuadrado (approved):
approved with some small comments

packages/hardhat-plugin/src/index.ts
kanej: It was being used to suppress opening during test runs. I am going to leave it for now.

PR #518, opened 2023-10-05
Add support for chains with less blocks than the amount of required confirmations

### LINKED ISSUE #516: Running against non-auto mining chain in first few blocks
If you set a local hh node to non-automine, and try and run against it very early, you _sometimes_ see this error:

image

I suspect the failure is in the issue is in our nonce syncing. We check for a confirmed count x blocks in the past, but what if that leads to a negative number? See:

https://github.com/NomicFoundation/hardhat-ignition/blob/98eeb89b746640f09500417673948936586553d1/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts#L86

We never ask for a negative block number (add validation to the rpc count method). And in this case we should figure out whether we should just look back to the genesis block or whether custom logic is required.

## Reproduce steps

1. Go to sample examples
2. Set the hh config so the hardhat network doesn't automine and it has a 10s interval
3. Start a new node
4. Run an ignition deployment against the node immediately

alcuadrado:
Fixes #516


packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
```diff
@@ -98,9 +101,14 @@ export async function getNonceSyncMessages(
...
 
+    const hasPendingTransactions =
+      safeConfirmationsCount === undefined
+        ? pendingCount > 0
+        : safeConfirmationsCount !== pendingCount;
+
     // Case 0: We don't have any pending transactions
     if (pendingTransactions.length === 0) {
-      if (safeConfirmationsCount !== pendingCount) {
+      if (hasPendingTransactions) {
```
kanej, line 111: We need to differentiate the names of the variables, so it is clear that we mean `ignition tracked pending transactions` and `onchainPendingCount`

kanej: maybe `hasPendingTransactions` should become `hasPendingTransactionsThatHaveNotConfirmed` or some such?

packages/core/test/execution/nonce-management/get-nonce-sync-messages.ts
```diff
@@ -145,394 +144,499 @@ describe("execution - getNonceSyncMessages", () => {
...
-          transactionCountEntries: {
-            [exampleAccounts[1]]: {
-              pending: pendingCount,
-              latest: latestCount,
-              number: () => safestCount,
+  describe("second deployment run", () => {
+    const latestBlockNumbers = [0, 3, 10];
+
+    for (const latestBlockNumber of latestBlockNumbers) {
+      describe(`With latest block ${latestBlockNumber}`, () => {
```
kanej, line 232:
```suggestion
      describe(`With latest block #${latestBlockNumber} and ${requiredConfirmations} required confirmations`, () => {
```

To make explicit that it is the no of confirmations interacting with the block number that makes each of these cases.

kanej (approved):
I left a few naming comments but looks good.

PR #522, opened 2023-10-09
chore: move UI deps to dev deps (PR 2)

### LINKED ISSUE #521: chore: cleanup ui deps
The dependencies in the UI packages should be `dev-dependencies`. The packages exposes the `visualize` html report as a file, but this is built during packaging - at which point the dependencies are no longer needed.

kanej:
Move the ui projects dependencies to dev-dependencies.

The UI npm package consists of a single `index.html` file that is used from the Ignition plugin. This is built during publishing. There is no need for any additional dependencies.

Resolves #521.


alcuadrado (approved):
I also made the ui a dependency of the plugin, and not a peerDependency.

Ideally, the generated template should be copied to the plugin instead of using a dependency.

PR #524, opened 2023-10-09
chore: remove serialize error dep (PR 3)

### LINKED ISSUE #523: chore: remove serialize-error dep
Our restructuring of error handling means we no longer have to serialize node error objects to the journal. We can remove the dependency on `serialize-error`.

kanej:
Our restructuring of error handling means we no longer pass error objects as results to messages. We can remove this dep.

Resolves #523.

### DIFF
--- a/config/typescript/tsconfig.json
+++ b/config/typescript/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
     "module": "commonjs",
     "declaration": true,
     "declarationMap": true,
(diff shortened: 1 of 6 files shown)


kanej:
Removing the serialized-error dep broke our tsc build. It exposed that
we are using features that require library support from typescript of at
least `es2022` to support cause on error objects and `at` on array
objects.

Just the error.cause part of es2022 assumes node 16.9 or above. We
technically support node 16.0 and above.

I have replaced our `at` usage with lodash `last`. I have added a hack
to get around the typesystem to let us pass a second arg to error in the
constructor on the assumption it is ignored.

PR #534, opened 2023-10-10
Fix bug and hide diagram for modules containing only one future

### LINKED ISSUE #529: bug in visualize jump to future
In the visualize report, the arguments and details of one future can contain a link to another future. Clicking the link expands that future and scrolls the future into view.

image

If the future being navigated to is closed/unexpanded this works. If the future is open the scroll into view fails, and the future is collapsed.

If a user clicks on a link to an already open future. The future should remain open. The future should be scrolled into view.

### LINKED ISSUE #532: Hide mermaid diagram for simple graphs
For simple modules, for instance where there is only one future, the mermaid diagram is confusing to new users.

image

If there is only one future we hide the "visualize batches" part and diagram of the deployment flow section. Instead we show a note that: “Module diagram will show once you have more than 1 future”

zoeyTM:
fixes #532 
fixes #529

### DIFF
--- a/packages/ui/src/components/socials/index.tsx
+++ b/packages/ui/src/components/socials/index.tsx
@@ -9,7 +9,7 @@ export enum SocialsEnum {
 }
 
 const SOCIALS_LINKS = {
-  [SocialsEnum.GITHUB]: "https://github.com/NomicFoundation/hardhat-ignition",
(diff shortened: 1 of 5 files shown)


kanej (approved):
LGTM.

PR #539, opened 2023-10-10
improvement: change `id` to `deployment-id` in tasks

### LINKED ISSUE #538: Rename `id` to `deployment-id` in tasks
Tasks that accept a deployment id should have their parameter renamed from `id` to `deployment-id`.

kanej:
Tasks that accept a deployment id have had their param renamed to the more explicit `--deployment-id`.

Updates `--future` to `--future-id in the wipe task as well, to be consistent.

Resolves #538.


packages/hardhat-plugin/src/index.ts
```diff
@@ -221,15 +221,15 @@ ignitionScope
 
 ignitionScope
   .task("status")
-  .addParam("id", "The id of the deployment to show")
+  .addParam("deploymentId", "The id of the deployment to show")
```
alcuadrado, line 224: Unrelated, but this one could be positional, right?

packages/hardhat-plugin/src/index.ts
```diff
@@ -318,15 +318,12 @@ ignitionScope
 
 ignitionScope
   .task("wipe")
-  .addParam("id", "The id of the deployment that has the future to wipe")
-  .addParam("future", "The id of the future to wipe")
+  .addParam("deploymentId", "The id of the deployment with the future to wipe")
```
alcuadrado, line 321: Should these be positional too?

alcuadrado (approved):
Left some comments, but these changes look good to me

packages/hardhat-plugin/src/index.ts
kanej: Switched to positional

packages/hardhat-plugin/src/index.ts
kanej: I have switched both `deploymentId` and `futureId` to be positional.

PR #541, opened 2023-10-10
handle error thrown by status task

### LINKED ISSUE #527: Improve no deployment error message in status
If you run the status command with a non-existant deployment id you get:

image

At the least the stack trace should be removed. Maybe a pattern to follow is HHs:

image

zoeyTM:
fixes #527 

Screen Shot 2023-10-10 at 5 09 59 PM

### DIFF
--- a/packages/hardhat-plugin/src/index.ts
+++ b/packages/hardhat-plugin/src/index.ts
@@ -1,9 +1,8 @@
-import type {
+import {
   DeploymentParameters,
   IgnitionError,
   StatusResult,
 } from "@nomicfoundation/ignition-core";
-
 import "@nomicfoundation/hardhat-ethers";
 import chalk from "chalk";
 import { readdirSync } from "fs-extra";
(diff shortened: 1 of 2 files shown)


packages/hardhat-plugin/src/index.ts
```diff
@@ -247,7 +248,15 @@ ignitionScope
       id
     );
 
-    const statusResult = await status(deploymentDir);
+    let statusResult: StatusResult;
+    try {
+      statusResult = await status(deploymentDir);
+    } catch (e) {
+      throw new NomicLabsHardhatPluginError(
```
alcuadrado, line 255: I think it'd be better to selectively rethrow, something like this:


```
catch (e) {
  if (e instanceOf IgnitionError && shouldBeHardhatPluginError(e)) {
    throw new ...
  }
  
  throw e;
}
```

```
function shouldBeHardhatPluginError(error:IgnitionError):boolean {
   /// white list of errors by their code
}
```

alcuadrado: that way we can use that logic on all the tasks, and make sure we are handling all the errors we care about

zoeyTM: done!

kanej (approved):
I am going to merge this.

I think we should reuse it across our other tasks, but lets do that as a separate PR.

PR #542, opened 2023-10-10
feat: expose `buildModule` on its own entry point

### LINKED ISSUE #540: export `buildModule` from `@nomicfoundation/hardhat-ignition/modules`
We currently re-export `buildModule` from core in the `main` file of the `@nomicfoundation/hardhat-ignition` package. This works but is dangerous, as the rest of the file loads in significant Hardhat specific plugin code.

Instead we will add another entry point into `@nomicfoundation/hardhat-ignition` at `modules` and re-export `buildModule` from there.

kanej:
We re-export `buildModule` in `hardhat-ignition` from a `modules` entry point to avoid clashes with Hardhat plugin setup. Now to setup a module you need to import:

```
import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";
```

Resolves #540.

## Testing 

I created a the HH sample project in both js and ts modes and was able to leverage the new entrypoint.

### DIFF
--- a/examples/complete/ignition/modules/CompleteModule.js
+++ b/examples/complete/ignition/modules/CompleteModule.js
@@ -1,5 +1,5 @@
 // ./ignition/CompleteModule.js
-const { buildModule } = require("@nomicfoundation/hardhat-ignition");
+const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");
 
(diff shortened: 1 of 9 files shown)


alcuadrado:
LGTM.

I also checked all the imports of `buildModule` and they were all updated.

alcuadrado (approved):

PR #544, opened 2023-10-12
Add error rethrowing to all plugin tasks

### LINKED ISSUE #543: extend HH error remapping to other tasks
Currently the HH error remapping covers status. It should be extended to:

* deploy
* visualize
* wipe

Judgement should be used in determining if there are obvious errors in those tasks that should be added to the white list. For instance for wipe, what is the error if the deployment id doesn't exist and what is the error if the future doesn't exist.

zoeyTM:
- added error rethrowing to all tasks
- added two new error codes to the whitelist
- fixed a bug in the `wipe` task

fixes #543


zoeyTM:
As a general comment, what is our internal criteria for whether an error code should be part of the whitelist?

kanej (approved):
LGTM!

PR #546, opened 2023-10-12
fix: timeout display

kanej:
Init events where overwriting the batch init for futures, and setting them to pending (aka timeout).

The fix is to set them to unstarted (the same as the batch init).

This fixes the flash of pending/timed out displayed for futures.

At the same time several of the future events have been updated to redisplay the batch. This is a separate bug, but fixing it makes the initial issue more obvious.

Fixes https://github.com/NomicFoundation/hardhat-ignition/issues/545.

### Before (with proper batch refreshes)

before

### After

after

### DIFF
--- a/packages/hardhat-plugin/src/ui/pretty-event-handler.ts
+++ b/packages/hardhat-plugin/src/ui/pretty-event-handler.ts
@@ -412,7 +412,7 @@ export class PrettyEventHandler implements ExecutionEventListener {
       const f: UiFuture = {
         futureId,
         status: {
-          type: UiFutureStatusType.PENDING,
+          type: UiFutureStatusType.TIMEDOUT,
         },
       };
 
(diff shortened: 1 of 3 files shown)


packages/hardhat-plugin/src/ui/pretty-event-handler.ts
```diff
@@ -134,155 +135,65 @@ export class PrettyEventHandler implements ExecutionEventListener {
...
-      status: this._getFutureStatusFromEventResult(event.result),
-    };
-
-    this.state = {
-      ...this.state,
-      batches: this._applyUpdateToBatchFuture(updatedFuture),
-    };
-
-    this._redisplayCurrentBatch();
+    this._setFutureStatusCompletAndRedisplayBatch(event);
```
alcuadrado, line 144: Complete*

There's an e missing

alcuadrado (approved):
There's a small typo, but otherwise LGTM

PR #549, opened 2023-10-12
Make ignition-core a dependency, and not a peerDependency

alcuadrado:
As we aren't asking the user to use core directly, there's no need for it to be a peer dependency. Making it a dependency leads to a simpler ux.


kanej (approved):
LGTM

PR #558, opened 2023-10-16
feat: change contractAt signature

### LINKED ISSUE #557: Update `contractAt` signature for artifacts
To bring `contractAt` inline with other optional artifact signatures, we should change the signature to:

```ts
m.contractAt(name, artifact, address)
```

Effectively swapping the position of artifact and address.

kanej:
Switch the `artifact` version of `contractAt` to:

```js
m.contractAt(name, artifact, address)
```

This brings it into line with other artifact based signatures.

Resolves #557.

### DIFF
--- a/packages/core/src/internal/module-builder.ts
+++ b/packages/core/src/internal/module-builder.ts
@@ -18,24 +18,24 @@ import {
   AccountRuntimeValue,
   AddressResolvableFuture,
   ArgumentType,
-  ContractAtFuture,
-  ContractDeploymentFuture,
-  LibraryDeploymentFuture,
   CallableContractFuture,
+  ContractAtFuture,
+  ContractCallFuture,
+  ContractDeploymentFuture,
   ContractFuture,
   FutureType,
   IgnitionModule,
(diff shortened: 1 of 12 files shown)


packages/core/src/internal/module-builder.ts
```diff
@@ -633,19 +638,32 @@ class IgnitionModuleBuilderImplementation<
...
+    if (isArtifactType(addressOrArtifact)) {
+      if (
+        !(
+          typeof optionsOrAddress === "string" ||
+          isFuture(optionsOrAddress) ||
+          isModuleParameterRuntimeValue(optionsOrAddress)
+        )
+      ) {
+        this._throwErrorWithStackTrace(
+          `Invalid parameter "address" provided to contractAt "${contractName}" in module "${this._module.id}"`,
```
kanej, line 650: This is following the pattern from `contract`.

There is a later further validation of the address field with `_contractAtFromArtifact`, but the pattern does not seem to be to pull these up, but instead to give specific type based error here.

PR #560, opened 2023-10-16
Define call ids based on the entire id of the contract being called, not just its name (PR 2)

alcuadrado:
This PR introduces a small change that makes working with multiple instances of the same contract in a module easier.

### DIFF
--- a/packages/core/src/internal/utils/future-id-builders.ts
+++ b/packages/core/src/internal/utils/future-id-builders.ts
@@ -5,6 +5,12 @@
 const MODULE_SEPERATOR = "#";
 
 /**
+ * The separator in ids that depend on futures that belong to a submodule.
+ * This separator is used to split the submodule and the rest of the dependency's id.
+ */
+const SUBMODULE_SEPARATOR = "-";
+
+/**
  * The seperator in ids that indicated different subparts of the future key.
  */
 const SUBKEY_SEPERATOR = ".";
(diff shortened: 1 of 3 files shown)


packages/core/src/internal/utils/future-id-builders.ts
```diff
@@ -4,6 +4,12 @@
  */
 const MODULE_SEPERATOR = "#";
 
+/**
+ * The separator in ids that depend on futures that belong to a submodule.
+ * This separator is used to split the submodule and the rest of the dependency's id.
+ */
+const SUBMODULE_SEPARATOR = "-";
```
alcuadrado, line 11: I accept better alternatives. I tried "##", but I think we'd regret it in the future.

kanej: What do you think of `~`?

We allow the user to provide `-` for deployment ids, so I worry we would create ambiguity in parsing.

Alternatively we could ban `-` in deployment ids and all other ignition ids to ensure `-` only has the meaning _SUBMODULE_SEPARATOR_.

alcuadrado: That's a good point.

I hesitated about `~` initially, as it has some special meaning in UNIX systems, like `~`, `~pato`. But in our case it will always be in the middle of two words, so it shouldn't be a problem.

PR #564, opened 2023-10-16
Better validation error messages

alcuadrado:
Before this PR validation errors weren't displaying their stack traces. This was problematic, as those are needed to be able to debug your errors. We intentionally construct stack traces that can be helpful for that in `core`.

After this PR, the plugin doesn't wrap those validation errors in a `HardhatPluginError`, instead, it prints them and throws a more basic plugin error.

While doing this, I introduce three small improvements:

* `IgnitionError` doesn't pass the `cause` property if it's undefined, as that interferes with Node's default error formatting.
* `IgnitionError` now has a `errorNumber` property, that can be used to check which error was thrown.
* When creating a plugin error, we pass the `cause`/`parent` error wherever appropriate.

### DIFF
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -8,7 +8,7 @@ import { ERRORS, ErrorDescriptor, getErrorCode } from "./errors-list";
  */
 export class CustomError extends Error {
   constructor(message: string, cause?: Error) {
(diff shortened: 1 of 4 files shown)


alcuadrado:
Reviewing one commit at a time should be simpler.

PR #565, opened 2023-10-16
fix: rerun deploying panel display (PR A)

kanej:
Our display of the deployment starting panel is dependent on the `RUN_START` event. Previously this event is triggered only once for a deployment on initialization of the deployment state on the very first run of the deployment. At the least this is a misnomer.

Several moves have been made here:

1. The UI only event DEPLOYMENT_START now occurs at the start of a deploy for both fresh runs and reruns.
2. The `RUN_START` event now occurs at the start of `executeModule` in the execution engine. So there is a `RUN_START` message in the journal for each run.
3. The execution engine is no longer invoked if there are no batches to run (i.e. no uncompleted futures). This means no `RUN_START` event if nothing is executed.
4. The `RUN_START` event no longer clears the previous execution states (a bug that was masked by it only ever being run once in the previous version).

An integration test has been added to pin the `RUN_START` event behaviour of not wiping previous run history.

These changes gives us the correct UI combinations:

- Show Deploying on fresh deploy
- Only show complete on rerun with no new futures
- Show deploying on rerun with new futures

Fixes https://github.com/NomicFoundation/hardhat-ignition/issues/566.

## Preview

Now on the second run with a new contract `AnotherContract` we see the Ignition panel:

image

And on a rerun with no new futures we see:

image


kanej:
@alcuadrado there are two considerations here I think.

The first is whether we want a new run event on every run in the journal. I think recording an entry for a new run makes sense, though perhaps adding a timestamp would be the way to go. An alternative though would be to rename the journalable event from `RUN_START` to `DEPLOYMENT_INIT`, then instead of having a `RUN_START` at the beginning of execution phase, we put in a UI only event, and use that for the cli UI ignition running display.

The second issue, is that there is a duplicate call to get chainId. The first is to do the check that we haven't changed the chainId. The second is in the execution engine to record the start of the run (the chainId is needed for that). We could do the chainId call once at the start of deploy and pass  the information down (reworking the execution engine signature).

alcuadrado:
I think we should not touch the journal if not necessary. Imagine having the journal in git, and having to commit or reset the file on every run. That'd be unnecessarily annoying.

alcuadrado:
Execution events are different from the journal messages. They normally match, but they don't have to.

IMO the correct approach would be emitting an execution event on every run, but not recording a journal message.

alcuadrado:
> An alternative though would be to rename the journalable event from `RUN_START` to `DEPLOYMENT_INIT`, then instead of having a `RUN_START` at the beginning of execution phase, we put in a UI only event, and use that for the cli UI ignition running display.

Yes, let's do this instead.

> The second issue, is that there is a duplicate call to get chainId. The first is to do the check that we haven't changed the chainId. The second is in the execution engine to record the start of the run (the chainId is needed for that). We could do the chainId call once at the start of deploy and pass the information down (reworking the execution engine signature).

I'm not worried about that duplicated call. I'm pretty sure hardhat caches that info.

kanej:
We swapped to the version where the original journalable `RUN_START` event was renamed to `DEPLOYMENT_INITIALIZE`, and only sets the chainId.

A new UI only event `RUN_START` was then added. That combined with only triggering the `RUN_START` event if there were futures to execute, fixed the issue.

PR #569, opened 2023-10-17
Better duplicated id errors

alcuadrado:
This PR changes the errors shown when ids are duplicated.

It splits the errors into two cases, with different messages:

1. An autogenerated id is duplicated.
2. An id provided by the user is duplicated.

It also changes the wording, so that we use the full id ("Module#autoId") only in the autogenerated case, and just the id ("userProvidedId") when the user provides it. This change is because the user may not recognize that "Module#userProvidedId" refers to the string they provided, as it has been changed.


kanej (approved):
LGTM. I moved the duplicate id recording out of the assert.

PR #573, opened 2023-10-19
fix: display wrong chainId correctly

### LINKED ISSUE #571: Misdisplay on the changed chainId warning
The Ignition starting message is visible if you trigger a changed `chainId` warning:

image

kanej:
Move the deployment start event to after the state initialization where the chainId check is. This works, but I suspect we should consider dealing with unknown exceptions occuring during UI processing.

The chainId check now throws an Ignition error from a new `DEPLOY` range of errors. This error has been added to the whitelist (we don't need to see the stack trace).

## Preview

image

Fixes #571.


packages/core/src/internal/deployer.ts
```diff
@@ -74,10 +76,10 @@ export class Deployer {
     accounts: string[],
     defaultSender: string
   ): Promise<DeploymentResult> {
-    this._emitDeploymentStartEvent(ignitionModule.id, this._deploymentDir);
-
     let deploymentState = await this._getOrInitializeDeploymentState();
 
+    this._emitDeploymentStartEvent(ignitionModule.id, this._deploymentDir);
```
kanej, line 81: This resolves the issue with the `Ignition starting...` text appearing before the chainId error.

There is a larger change that this suggests though, which is to rework our unexpected exception handling to clear the UI.

packages/core/src/errors-list.ts
```diff
@@ -68,6 +68,11 @@ export const ERROR_RANGES: {
     max: 899,
     title: "Status errors",
   },
+  DEPLOY: {
```
kanej, line 71: I added a deploy section, for errors that are expected in the deploy task outside of execution.

packages/core/src/internal/deployer.ts
alcuadrado: I don't full understand this, but let's merge this pr, as it looks good as a small/local fix.

PR #575, opened 2023-10-19
fix: stop future senders clearing out pending trans (PR A)

### LINKED ISSUE #574: rerun blocked by sent transactions message
Killing a deployment midway, then rerunning can lead to the `403` Ignition error saying you have sent transactions. Even though all the transactions where sent on the previous run by Ignition and should be tracked.

## Reproduction steps

1. Update the `examples/sample` module to deploy multiple contracts one after the other using `after`.
2. Set the node so that automining is false and the interval is long enough to kill and restart a deploy.
3. Start the first run and ctrl c after the first batch has started
4. Rerun the deployment
5. You should see the 403 error

image

kanej:
The future senders update the account to nonce+transactions list in such away that existing pending transactions for the account are untouched.

No test has been added, but the case has been covered by enhancing the example module to include an unexecuted future in the module.

Fixes #574.

### DIFF
--- a/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
+++ b/packages/core/src/internal/execution/nonce-management/get-nonce-sync-messages.ts
@@ -291,7 +291,9 @@ function createMapFromSenderToNonceAndTransactions(
   );
 
   for (const futureSender of futureSenders) {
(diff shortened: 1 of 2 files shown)


alcuadrado:
Good catch!

PR #578, opened 2023-10-22
Small mermaid fixes

alcuadrado:
This PR introduces some small changes to the mermaid graph:

- It swaps the order of the graph, as when writing docs with Fran we noticed it was confusing in the top-to-bottom order, so I switched it to bottom-to-top
- I changed the function that escapes ids so that it escapes `~` because we are using them now, and also tweaked it a bit to try to avoid clashes.
- While doing this I realized that some ids weren't escaped, so i fixed that.

### DIFF
--- a/packages/ui/src/utils/to-escaped-id.ts
+++ b/packages/ui/src/utils/to-escaped-id.ts
@@ -3,7 +3,7 @@
  */
 export function toEscapedId(id: string): string {
   return id
-    .replaceAll("(", "__)")
+    .replaceAll("(", "__")
     .replaceAll(")", "___")
     .replaceAll(",", "____")
     .replaceAll("~", "_____")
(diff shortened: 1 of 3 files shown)


packages/ui/src/utils/to-escaped-id.ts
```diff
@@ -3,8 +3,10 @@
  */
 export function toEscapedId(id: string): string {
   return id
-    .replace("(", "_")
-    .replace(")", "_")
-    .replace(",", "_")
-    .replace(" ", "_");
+    .replaceAll("(", "__)")
```
kanej, line 6: Is it safe for me to remove the `)` in `"__)"`, or is there a move I am not seeing?

I think line 7 replaces it.

alcuadrado: OMG! That explains why i struggled so much with those tests! It's just a typo

kanej (approved):
LGTM

PR #582, opened 2023-10-23
Add a message when a deployment is being resumed

alcuadrado:


packages/hardhat-plugin/src/ui/helpers/cwd-relative-path.ts
```diff
@@ -0,0 +1,12 @@
+import path from "path";
+import process from "process";
+
+export function pathFromCwd(thePath: string): string {
+  const cwd = process.cwd();
+
+  if (thePath.startsWith(cwd)) {
+    return `./${path.relative(process.cwd(), thePath)}`;
```
kanej, line 8: This will likely look wrong on windows: `./users\\example`.

PR #593, opened 2023-10-25
Changed the rendered text for previously successful depoyment

### LINKED ISSUE #586: Indicate that nothing was executed on rerun
> If I rerun a deployment that is already complete, and doesn’t need any further action, the output shouldn’t be “successfully deployed”.

We don't give enough of a visual cue on a rerun where no futures are executed. We should successful deploy when nothing was deployed.

zoeyTM:
fixes #586

### DIFF
--- a/packages/hardhat-plugin/src/ui/pretty-event-handler.ts
+++ b/packages/hardhat-plugin/src/ui/pretty-event-handler.ts
@@ -261,8 +261,10 @@ export class PrettyEventHandler implements ExecutionEventListener {
       batches: this._applyResultToBatches(this.state.batches, event.result),
     };
 
+    const anythingDone = this.state.currentBatch > 0;
+
     // If batches where executed, rerender the last batch
-    if (this.state.currentBatch > 0) {
+    if (anythingDone) {
       this._redisplayCurrentBatch();
     } else {
       // Otherwise only the completion panel will be shown so clear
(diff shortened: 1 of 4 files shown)


zoeyTM:
Screen Shot 2023-10-24 at 10 59 54 PM

packages/hardhat-plugin/src/ui/pretty-event-handler.ts
```diff
@@ -261,16 +261,23 @@ export class PrettyEventHandler implements ExecutionEventListener {
       batches: this._applyResultToBatches(this.state.batches, event.result),
     };
 
+    const anythingDone = this.state.currentBatch > 0;
+
     // If batches where executed, rerender the last batch
-    if (this.state.currentBatch > 0) {
+    if (anythingDone) {
       this._redisplayCurrentBatch();
```
kanej, line 268: Lets pull out `this.state.currentBatch > 0` as a util function, and use it here but also use it in `calculateDeploymentCompleteDisplay`.

zoeyTM:
updated messaging:

Screen Shot 2023-10-26 at 1 37 39 AM

kanej (approved):

packages/hardhat-plugin/src/ui/pretty-event-handler.ts
kanej: I have pushed a commit that takes the util function route. It means that `calculateDeploymentCompleteDisplay` can keep the interface of accepting `UiState` and then deriving what to display, so better encapsulation.
Looking at what had to be done to get there though makes me feel it might not have been worth it.

PR #606, opened 2023-10-27
Checksum deployed addresses

### LINKED ISSUE #600: Addresses should show checksums
We are displaying addresses without the checksums: https://docs.ethers.org/v5/api/utils/address/#address-formats

The recorded deployed addresses are missing their checksums as well.

image

zoeyTM:
fixes #600

### DIFF
--- a/packages/core/src/internal/execution/abi.ts
+++ b/packages/core/src/internal/execution/abi.ts
@@ -1,10 +1,11 @@
-import type {
+import {
   EventFragment,
   Fragment,
   FunctionFragment,
   Interface,
   ParamType,
   Result,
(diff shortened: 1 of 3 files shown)


kanej (approved):
LGTM! Nice work on the `getAddress` and tracking down the source of the unchecksumed addresses.

I moved the conversion up a layer to the json-rpc client. I think our rule should be that any `address` returned on any object from json-rpc should be checksum formatted. Which I think is the case with these changes.

Well done on the address comparison catch. I added a comparator utility function. We should look out for other places where we might be comparing two addresses in other situations.

PR #612, opened 2023-10-30
Viem support

### LINKED ISSUE #603: Create new view package that mirrors `hardhat-plugin`
- [x] #610

zoeyTM:
Adds a new package for Viem support and a new package separating out the existing Ethers support.

`@nomicfoundation/hardhat-ignition` remains and contains most of the Ignition functionality, but to leverage Ignition in tests or scripts within hardhat now requires installing one of the new viem/ethers extension packages:

- @nomicfoundation/hardhat-ignition-ethers
- @nomicfoundation/hardhat-ignition-viem

The `@nomicfoundation/hardhat-ignition` installs a stub IgnitionHelper under `hre.ignition` that will throw a warning if the user attempts to use it without ethers/viem support installed.

The integration tests under `hardhat-plugin` have been switched to use `viem` from `ethers`.

## Viem Support

The viem extension will provide viem contract instances as results from `hre.ignition.deploy(...)` calls. The instances will be typed based upon either the codegen of `hardhat-viem` for HH contracts or the type of the abi of the artifact for `m.contract("MyContract", myArtifact)` module builder calls.

> Note: you likely need to add `as const` to your artifact declaration to get the correct viem type support.

fixes #603 

## TODO

- [x] Create `packages/hardhat-plugin-viem` as a copy of `packages/hardhat-plugin`
- [x] Update viem plugin to use viem instead of ethers
- [x] Update all references to ethers across the viem plugin
- [x] Update `hardhat-ignition` integration tests
- [x] Create sample project for viem
- [x] Add hardhat-viem type generation support to the viem `hre.ignition.deploy` method
- [x] Add viem typing support to passed in artifacts.


kanej (approved):
This looks good and makes sense. I have rebased against development and squashed the commits down. I have also set the PR to draft to avoid the danger of us merging it before we have the other PRs in this line of work finished.

.gitignore
```diff
@@ -92,9 +92,6 @@ out
 .nuxt
 dist
 
-# Typedocs documentation files
-typedocs/
```
kanej, line 96: Typedocs is no longer used.

packages/core/src/internal/execution/abi.ts
```diff
@@ -862,3 +865,11 @@ function hasDynamicSize(paramType: ParamType): boolean {
     paramType.type === "string"
   );
 }
+
+/**
+ * A utility function to capture the cast from the `Abi` type to the
+ * ethers type.
+ */
+function toEthersAbi(abi: Abi): InterfaceAbi {
+  return abi as any[];
```
kanej, line 874: The abi of abi type I assume to be a valid json api snippet, and so convertible to ethers unchanged.

packages/core/src/types/artifact.ts
```diff
@@ -1,12 +1,14 @@
+import type { Abi } from "abitype";
+
 /**
  * An compilation artifact representing a smart contract.
  *
  * @beta
  */
-export interface Artifact {
+export interface Artifact<AbiT extends Abi = Abi> {
```
kanej, line 8: Artifact now types abi using the "abitype" library, ensuring we are able to provide viem typing support.

packages/hardhat-plugin-ethers/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
```
kanej, line 1: We need to consider whether this should just be populated with the existing hardhat-plugin changelog and similarly for the viem package.

packages/hardhat-plugin-viem/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "@types/chai": "^4.2.22",
+    "@types/chai-as-promised": "^7.1.4",
+    "@types/sinon": "^10.0.13"
+  },
+  "peerDependencies": {
+    "@nomicfoundation/hardhat-viem": "^1.0.0",
+    "hardhat": "^2.18.0"
+  },
+  "dependencies": {
+    "@nomicfoundation/hardhat-ignition": "^0.11.2",
```
alcuadrado, line 65: This should be a peer dependency

packages/hardhat-plugin-viem/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "@nomicfoundation/ignition-core": "^0.11.2",
+    "abitype": "0.10.2",
+    "viem": "1.18.1"
```
alcuadrado, line 68: peer dependecy

packages/hardhat-plugin-viem/src/modules.ts
```diff
@@ -0,0 +1 @@
+export { buildModule } from "@nomicfoundation/ignition-core";
```
alcuadrado, line 1: Let's not export this here. Users will import from `hardhat-ignition`

alcuadrado: @fvictorio wdty? this is a similar dynamic to the toolbox and the network helpers. With the difference that we have two different plugins: ethers, and viem. My worry is that if we reexport we'd partition the userbase & docs into two

packages/hardhat-plugin-ethers/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@types/sinon": "^10.0.13"
+  },
+  "peerDependencies": {
+    "@nomicfoundation/hardhat-ethers": "^3.0.4",
+    "hardhat": "^2.18.0"
+  },
+  "dependencies": {
+    "@nomicfoundation/ignition-core": "^0.11.2",
+    "@nomicfoundation/hardhat-ignition": "^0.11.2",
+    "ethers": "^6.7.0"
```
alcuadrado, line 66: peer dependencies

alcuadrado: ignition-core is also part of the public API, the modules, so it should be peer too

packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
```diff
@@ -0,0 +1,230 @@
...
+  private static async _convertArtifactToViemContract(
+    hre: HardhatRuntimeEnvironment,
+    future:
+      | ContractDeploymentFuture
+      | LibraryDeploymentFuture
+      | ContractAtFuture,
+    deployedContract: { address: string }
+  ) {
+    const publicClient = await hre.viem.getPublicClient();
+    const [walletClient] = await hre.viem.getWalletClients();
```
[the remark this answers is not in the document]
alcuadrado: Still, let's not block on this yet. We can release with a disclaimer that it doesn't work with read-only networks, but we'd have to fix it quick after the release. Or is it too annoying?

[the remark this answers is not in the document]
alcuadrado: Note that this function is only used in scripts and tests. So walletClients are gonna be available the vast majority of the time.

packages/core/src/types/artifact.ts
alcuadrado: I think we can use a `const` type param https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#const-type-parameters

packages/hardhat-plugin-viem/package.json
kanej: I have changed both the viem and ethers packages to have peer dependencies instead: https://github.com/NomicFoundation/hardhat-ignition/pull/612/commits/c395a4a4dfdd24438781efe4effbfa8f0a74ba5e

packages/hardhat-plugin-viem/src/modules.ts
fvictorio: I'm not sure I understand that are the options here

packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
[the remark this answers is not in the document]
fvictorio: We throw an error. That can probably be improved, but no one has complained so far: https://github.com/NomicFoundation/hardhat/blob/3a657398a5f0ccaa9d8f365bb2665ba54c33e9ea/packages/hardhat-viem/src/internal/contracts.ts#L232-L243

alcuadrado (approved):

packages/hardhat-plugin-viem/src/modules.ts
kanej: We have opted for option 2.

PR #622, opened 2023-10-31
Validate that module id's are not duplicated

### LINKED ISSUE #608: Validation on duplicate module ids in visualization
@maguroid reported an issue where visualization failed when running against this module:

https://github.com/maguroid/hardhat-ignition-demo/blob/factory-pattern/ignition/modules/Vault.ts

It appears to reduce down to:

```js
import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";
import VaultFactory from "./VaultFactory";

import hre from "hardhat";

const hex3 = hre.ethers.toBeHex(3, 20);

const VaultFactory = buildModule("Vault", (m) => {
  const vaultFactory = m.contract("VaultFactory");

  return {
    vaultFactory,
  };
});

const UniswapV3Swapper = buildModule("UniswapV3Swapper2", (m) => {
  const uniswapV3Swapper = m.contract("UniswapV3Swapper", [hex3]);

  return {
    uniswapV3Swapper,
  };
});

// * NOTES: deployment with factory currently not working as expected.
// *        deployed address not stored in json file, and visualization throws error
export default buildModule("Vault", (m) => {
  const { vaultFactory } = m.useModule(VaultFactory);
  const { uniswapV3Swapper } = m.useModule(UniswapV3Swapper);

  return {
    vaultFactory,
    uniswapV3Swapper,
  };
});
```

I suspect this is the reuse of the module id. Where the top level module has id `Vault` but so does one of its submodules.

We should add `buildModule` validation to cover this case.

zoeyTM:
fixes #608

### DIFF
--- a/packages/hardhat-plugin/src/utils/shouldBeHardhatPluginError.ts
+++ b/packages/hardhat-plugin/src/utils/shouldBeHardhatPluginError.ts
@@ -9,9 +9,9 @@ import type { IgnitionError } from "@nomicfoundation/ignition-core";
  *    - If there's an exception that doesn't fit in either category, let's discuss it and review the categories.
  */
 const whitelist = [
-  200, 201, 202, 203, 403, 404, 405, 600, 601, 602, 700, 701, 702, 703, 704,
-  705, 706, 707, 708, 709, 710, 711, 712, 713, 714, 715, 716, 717, 718, 719,
-  720, 721, 722, 723, 724, 725, 726, 800, 900,
+  200, 201, 202, 203, 204, 403, 404, 405, 600, 601, 602, 700, 701, 702, 703,
+  704, 705, 706, 707, 708, 709, 710, 711, 712, 713, 714, 715, 716, 717, 718,
+  719, 720, 721, 722, 723, 724, 725, 726, 800, 900,
 ];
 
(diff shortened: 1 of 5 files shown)


packages/hardhat-plugin/src/utils/shouldBeHardhatPluginError.ts
```diff
@@ -9,9 +9,9 @@ import type { IgnitionError } from "@nomicfoundation/ignition-core";
  *    - If there's an exception that doesn't fit in either category, let's discuss it and review the categories.
  */
 const whitelist = [
-  200, 201, 202, 203, 403, 404, 405, 600, 601, 602, 700, 701, 702, 703, 704,
-  705, 706, 707, 708, 709, 710, 711, 712, 713, 714, 715, 716, 717, 718, 719,
-  720, 721, 722, 723, 724, 725, 726, 800, 900,
+  200, 201, 202, 203, 204, 403, 404, 405, 600, 601, 602, 700, 701, 702, 703,
```
kanej, line 12: I included it in the whitelist to get the nicer error message:

image

kanej (approved):
Looks good. I added 204 to the whitelist to get the nicer HH error display. LGTM!

PR #623, opened 2023-10-31
Allow sending to AccountRuntimeValue

### LINKED ISSUE #618: send should support account as a to parameter
The API should be modified to support passing an AccountRuntimeValue to the `to` of `send`. Currently it errors:

image

zoeyTM:
fixes #618


kanej (approved):
LGTM!

I have added additional tests for serialization and reconciliation.
I also tightened an address comparison, which led to adding an extra validation in buildModule to ensure to addresses are addresses.

PR #624, opened 2023-10-31
Catch base fee error and rethrow as ignition error

### LINKED ISSUE #594: Fix error output for fee exceeding block gas limit
Error output when `initialBaseFeePerGas` in hardhat config is greater than the `blockGasLimit` is an ugly stack trace.

- [ ] Add a validation to check for this case
- [ ] Add a new validation error to ensure the output looks good

zoeyTM:
fixes #594 

screenshot (with redacted file system info in stack trace):

Screen Shot 2023-10-31 at 11 28 30 AM

### DIFF
--- a/packages/core/src/errors-list.ts
+++ b/packages/core/src/errors-list.ts
@@ -174,6 +174,11 @@ export const ERRORS = {
       message:
         "The next nonce for %sender% should be %expectedNonce%, but is %pendingCount%. Please make sure not to send transactions from %sender% while running this deployment and try again.",
     },
+    BASE_FEE_EXCEEDS_GAS_LIMIT: {
+      number: 406,
+      message:
+        "The configured base fee exceeds the block gas limit. Please reduce the configured base fee or increase the block gas limit.",
+    },
   },
   RECONCILIATION: {
     INVALID_EXECUTION_STATUS: {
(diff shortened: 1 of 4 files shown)


kanej:
Can you add a set of reproduction steps so I can test the error?

zoeyTM:
@kanej 

Steps to reproduce:

1. set the `hardhat.config.js` of an example to:
```
hardhat: {
      mining: {
        auto: false,
      },
      blockGasLimit: 1,
      initialBaseFeePerGas: 1_000_000_000,
    },
```
2. run `npx hardhat node`
3. deploy the example module with `--network localhost`

packages/core/src/errors-list.ts
```diff
@@ -169,6 +169,11 @@ export const ERRORS = {
       message:
         "The next nonce for %sender% should be %expectedNonce%, but is %pendingCount%. Please make sure not to send transactions from %sender% while running this deployment and try again.",
     },
+    BASE_FEE_EXCEEDS_GAS_LIMIT: {
```
kanej, line 172: What do you think of adding it to the whitelist. Based on the rules I think it can be argued either way, but I suspect the stacktrace just isn't useful in this context.

kanej (commented):
Lets add an integration test for this.

packages/core/src/errors-list.ts
zoeyTM: I'm on board with that idea

kanej (approved):
Looks good. Lets get this **squash** merged!

PR #625, opened 2023-11-02
Restructure plugins for reduced code duplication (VIEM 2)

### LINKED ISSUE #616: Determine duplication and potential for sharing up code
Based on the viem plugin PR  https://github.com/NomicFoundation/hardhat-ignition/pull/612, determine the extent to which a shared package would minimize duplication and how much DRYing up is possible between two HH plugin packages.

## TODO

- [ ] Can tasks be shared up?
- [ ] Can the UI be moved into a shared package?
- [ ] How much of ignition helper can be extracted and shared?
- [ ] Where do the integration tests go?

zoeyTM:
This is a first pass at restructuring the plugins. The goals of this restructure are as follows:

1. Reduce duplicated code as much as possible between the plugins
2. Allow the flexibility of a lighter weight plugin for users who only want to deploy from the CLI (using `hardhat-ignition`) and are less interested in utilizing Hardhat Ignition in a testing environment

I believe this first pass achieves both of those goals, though this may not be an exhaustive set of changes and more may be added. Some notes on decisions I made:

- Created `hardhat-plugin-ethers` to mirror `hardhat-plugin-viem`, both of which have been stripped of all UI and task related code. All that remains in these plugins is their respective `IgnitionHelper` classes and the snippet of code that initializes it into the global `hre` object.
- Detect if another plugin has already initialized `hre.ignition` and throw, thereby enforcing mutual exclusivity between our viem and ethers plugins.
- The viem and ethers plugins each have `hardhat-ignition` (the "base" plugin containing all of the task and UI code) listed as a peer dependency. Additionally, they are both importing `hardhat-ignition` in their respective index files to ensure that the base plugin and its tasks are available when the "extension plugins" are used.

***Note:*** I haven't updated docs or tests yet as I wanted to get the green light on the overall design before I potentially waste time moving tests multiple times. Additionally, there are some open questions and considerations that I will add as comments to the PR below. These are either alternate ideas I had or potential implications that should be considered due to assumptions I've made so far.

closes #616


packages/hardhat-plugin-ethers/src/index.ts
```diff
@@ -0,0 +1,24 @@
+import "@nomicfoundation/hardhat-ignition";
```
zoeyTM, line 1: What are the implications (if any) of potentially importing this twice? i.e. if a user has already imported the base plugin in their Hardhat config file.

Is this something we can/should check for?

packages/hardhat-plugin-ethers/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { extendEnvironment } from "hardhat/config";
+import { HardhatPluginError, lazyObject } from "hardhat/plugins";
+
+import "./type-extensions";
+
+/**
+ * Add an `ignition` object to the HRE.
+ */
+extendEnvironment((hre) => {
+  if (hre.ignition !== undefined && hre.ignition.type !== "ethers") {
```
zoeyTM, line 12: Should we perhaps assign the respective `IgnitionHelper` instances each to their own unique properties on the `hre` object so they could be used simultaneously?

packages/hardhat-plugin-ethers/src/utils/error-deployment-result-to-exception-message.ts
```diff
@@ -0,0 +1,141 @@
+import {
```
zoeyTM, line 1: This util is the only one currently being shared between both ethers and viem plugins. Is this something we should maybe export from the base plugin?

packages/hardhat-plugin-ethers/src/index.ts
kanej: I think you have right with the check and enforcement of only one of either viem or ethers.

packages/hardhat-plugin-ethers/src/utils/error-deployment-result-to-exception-message.ts
kanej: I have exported it from the base plugin.
On the adding abstractions vs duplication scale I tend towards duplication until we know we really have the abstraction; however I think if `error-deployment-result-to-exception-message` changed in one without changing in another that would be something that right now violates the design. So I think we share it up from `hardhat-plugin`.

packages/hardhat-plugin-ethers/src/index.ts
kanej: I think this is now taken care of with the check for other ignition plugins.
I will run a manual test to see.

kanej (approved):
Lets get this merged into the Viem 1 base branch.

PR #626, opened 2023-11-02
Support create2

### LINKED ISSUE #629: Proof of concept create2 spike
Create a proof of concept execution strategy that uses allows Ignition to run the sample and complete projects with contract deployment done through a create2 factory.

kanej:
This is a proof of concept of how to support create2 in Ignition.

It adds a new execution strategy that alters the way contract deployment futures are handled. The contract is proxied through a create2 factory rather than becoming a deployment transaction.

This PR uses [createX](https://github.com/pcaversaccio/createx) as the create2 factory. If deploying to `sepolia` a known address is used. If deploying ephemerally, the create2 factory is deployed as part of a new `init` step in the execution strategy, then leveraged when processing futures.

Resolves #629 

## Testing Locally

To deploy a module using `create2`

```shell
cd examples/ts-sample
npx hardhat ignition deploy ./ignition/modules/LockModule.ts --strategy create2
```

A createX factory will be deployed and the Lock contract (with passed value) will be created via the factory (note this affects the set owner!).

image

## Future Considerations

- How should we support injecting of custom strategies? This POC accepts an enumeration of `{ basic, create2 }`
- What UI indications do we need to give that `create2` is being used?
- Do we need a validation to ensure contract deploys won't count as a duplicate under `create2` (same salt + bytecode)?
- How are we handling salts and configuring them?
- Are happy to deploy a factory if not present onchain
- Are we happy with a default deploy to 31337 on account 0
- How should we handle confirmations for the factory deploy?
- How are we supporting `init` functions on contracts for `create2`
- We need createX development so we get:
  - a list of chains with a working factory
  - signed transaction for autodeployment
  - an audit?

## Testing Plan

This PR and the strategies config PR include both unit and integration tests. A manual round of tests will include:

* [x] Deploy the rocket example to sepolia with `create2`
* [x] Deploy the rocket example to an alternate chain, confirm sepolia and alternate chain have the same address
* [x] Deploy an example to sepolia with `--verify` enabled
* [x] Run through each of the `./example` projects with create2 enabled
* [x] Confirm reconcilliation fails on changed salt between runs
* [x] Create a deployment, deploy with simple, add a future, continue with create2
* [x] Run an example with createX Permissioned Deploy Protection
* [x] Run an example with createX Cross-Chain Redeploy Protection


alcuadrado:
> A createX factory will be deployed and the Lock contract (with passed value) will be created via the factory (note this affects the set owner!).

Good catch! We should update that example everywhere so that the owner is received.

> How should we support injecting of custom strategies? This POC accepts an enumeration of { basic, create2 }

Let's leave this out of scope for now. We'll have to define something eventually, but not needed yet.

> What UI indications do we need to give that create2 is being used?

I think everywhere where it says "Deploying [ module]" we should add the strategy

> Are we happy with a default deploy to 31337 on account 0
> How should we handle confirmations for the factory deploy?
> How are we supporting init functions on contracts for create2

I think that to get this out quickly we can deploy the factory outside of ignition core for development networks and wait for that tx to confirm before calling ignition.

We should have a way for the strategy to be able to drive its initialization, but I think it'd be better to have this out sooner.

> We need createX development so we get:
>  * a list of chains with a working factory
All of them should be in the same address, so it's a matter of doing an `eth_getCode` to that address and see if there's code there.

> signed transaction for autodeployment
I'll ask pascal about this

> an audit?
Same

alcuadrado:
> How are we handling salts and configuring them?

This is trickier than what I had in mind actually. Because we don't have a mechanism to pass "strategy config" to core.

Maybe we can make the usage of crate x transparent, and let users handle the nonce. This would mean that the strategy gets to handle some sort of config. 

One potential way to implement this can be by accepting a strategy object in ignition core.

packages/core/src/internal/execution/create2-execution-strategy.ts
```diff
@@ -0,0 +1,342 @@
...
+
+      return;
+    }
+  }
+
+  public async *executeDeployment(
+    executionState: DeploymentExecutionState
+  ): DeploymentStrategyGenerator {
+    const artifact = await this._loadArtifact(executionState.artifactId);
+    const salt = ethers.id("test");
```
zoeyTM, line 179: Was this intentionally left as `test` ?

zoeyTM:
> an audit ?

I do want to point out that createX is explicitly [not audited](https://github.com/pcaversaccio/createx?tab=readme-ov-file#security-considerations).

packages/core/src/internal/execution/create2-execution-strategy.ts
kanej: It hasn't been intentionally left as `test`. We need to review how we approach setting a salt and make sure we are happy with it.

alcuadrado (approved):

PR #631, opened 2023-11-09
Add type support for ignition tests using hardhat-viem (VIEM 2)

zoeyTM:
Note: currently this only supports types for local contracts. Support for imported contracts and contracts with external artifacts will come after an update to `hardhat-viem`


packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
```diff
@@ -49,54 +57,29 @@ export class IgnitionHelper {
...
-      throw new HardhatPluginError("hardhat-ignition-viem", message);
-    }
-
-    return this._toViemContracts(ignitionModule, result);
+      successfulDeploymentResult
+    );
   }
 
-  private async _toViemContracts<
+  private static async _toViemContracts<
```
zoeyTM, line 77: What's the reasoning behind making this static but not the ethers equivalent functions?

packages/hardhat-plugin-ethers/src/index.ts
```diff
@@ -12,7 +12,7 @@ extendEnvironment((hre) => {
   if (
     hre.ignition !== undefined &&
     hre.ignition.type !== "stub" &&
-    hre.ignition.type !== "ethers"
+    hre.ignition.type !== "viem"
```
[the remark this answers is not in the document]
kanej: Agreed. Good catch, clearly I didn't think that through.

packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
kanej: Agreed. I will bring the ethers version inline.

kanej (approved):
I am going to merge this into the base viem branch, so it can all be reviewed in one place.

PR #632, opened 2023-11-16
Added `verify` task

### LINKED ISSUE #630: Verification as ignition deploy post-step
Support verification against etherscan as a `verify` subtask of `ignition`.

```shell
npx hardhat ignition verify <deploymentId>
```

Also support passing a `--verify` parameter to the `deploy` task. If passed immediately start a verification pass after a successful deployment.

The configuration setup for the verification should match the `hardhat-verify` plugin: https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#usage

## Implementation

Given a deployment, with successfully deployed contracts, cycle through each deployment future, retrieve the constructor args, looks up the artifact/build information under the deployments directory and then invoke the Etherscan class from the verify plugin, passing:

* Contract address - from the exState
* Contract source code - from the build info
* Contract name - constructed from the artifact’s sourceName + contractName
* Compiler version - from the build info long compiler version
* Encoded constructor arguments - from the exState

The configuration should match the exact config structure of the verify plugin.

We should aim for parity with verify-plugin in terms of blockexplorer support.

See the verify docs for what it exposes to other plugins: https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#using-programmatically

## TODO

- [x] Implement with minimal config assuming only standard etherscan
- [x] Investigate testing approach - how is this done in verify plugin
- [x] Testing should include:
  - [x] simple contract
  - [x] contract with complex constructor args
  - [x] contract with libraries

zoeyTM:
Adds a `verify` task to the `ignition` scope to verify contracts from a given deployment on Etherscan. 

Usage: `npx hardhat ignition verify <deploymentId>`

Currently only verifies on Etherscan. 

fixes #630


packages/core/src/internal/deployment-loader/file-deployment-loader.ts
```diff
@@ -117,6 +117,26 @@ export class FileDeploymentLoader implements DeploymentLoader {
     );
   }
 
+  public async readBuildInfo(futureId: string): Promise<BuildInfo> {
```
kanej, line 120: This is added onto the file deployment loader but the deployment loader interface right? So we aren't expanding the api, just leveraging the file deployment loader as the most sensible place to put the `readBuildInfo` logic?

packages/core/src/verify.ts
```diff
@@ -0,0 +1,64 @@
...
+      assertIgnitionInvariant(
+        exState.type === "DEPLOYMENT_EXECUTION_STATE",
+        "execution state is not a deployment"
+      );
+
+      const { contractName, constructorArgs } = exState;
+
+      return {
+        address: contract.address,
+        compilerVersion: buildInfo.solcLongVersion.startsWith("v")
```
kanej, line 53: What are the cases that trigger this?

packages/core/test/ignition-module-serializer.ts
```diff
@@ -90,6 +90,7 @@ describe("stored deployment serializer", () => {
     const fakeArtifact: Artifact = {
```
kanej, line 90: We should probably share up `fakeArtifact` into the test helpers? What do you think?

packages/hardhat-plugin/src/utils/verifyEtherscanContract.ts
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    const { message: guid } = await etherscanInstance.verify(
+      address,
+      sourceCode,
+      name,
+      compilerVersion,
+      args
+    );
+
+    await new Promise((res) => setTimeout(res, 1000));
```
kanej, line 19: Do we need to retry up to a certain timeout?

packages/hardhat-plugin/test/verify.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { Etherscan } from "@nomicfoundation/hardhat-verify/etherscan";
+
+import { assert } from "chai";
+
+import { getApiKeyAndUrls } from "../src/utils/getApiKeyAndUrls";
+import { verifyEtherscanContract } from "../src/utils/verifyEtherscanContract";
+
+import { useEphemeralIgnitionProject } from "./use-ignition-project";
+
+describe("verify", function () {
```
kanej, line 11: Maybe create a verify folder and split the apikeylist tests from the fuller tests?

packages/hardhat-plugin/test/verify.ts
```diff
@@ -0,0 +1,126 @@
...
+                sepolia: "sepoliaApiKey",
+              },
+              "mainnet"
+            ),
+          /No etherscan API key configured for network mainnet/
+        );
+      });
+    });
+
+    describe("verifyEtherscanContract", function () {
```
kanej, line 75: What happens in the case where the bytecode has been verified in a previous verification (say the constructor args varied but the contract deploy bytecode is the same?).

kanej (commented):
I am not sure from the code, do we cover the cases of:

* verifying a contract that uses libraries
* verifying a contract that uses a library in its constructor?

packages/core/src/internal/deployment-loader/file-deployment-loader.ts
zoeyTM: Correct. This did require disabling the type test for `FileDeploymentLoader`, however.

packages/core/src/verify.ts
zoeyTM: `hardhat-verify` throws an error if the `compilerVersion` doesn't start with `v`, and the version string saved to the build-info file doesn't start with one.

packages/hardhat-plugin/test/verify.ts
zoeyTM: Etherscan will auto-verify any verified bytecode that it recognizes and will return the `Already Verified` message that the `verify` task handles internally

packages/core/test/ignition-module-serializer.ts
zoeyTM: After checking in the helpers, it turns out that `fakeArtifact` already exists as a helper, but we're not using it. I added a commit propagating that across the tests.

packages/core/src/verify.ts
```diff
@@ -0,0 +1,77 @@
...
+    throw new IgnitionError(ERRORS.VERIFY.NO_CONTRACTS_DEPLOYED, {
+      deploymentDir,
+    });
+  }
+
+  const chainConfig = builtinChains.find(
+    (c) => c.chainId === deploymentState.chainId
+  );
+
+  // this case is a failure on our part
```
alcuadrado, line 41: The main difference with hardhat-verify is that it allows you to set a new chain in your config. Maybe `verify` in core can take an optional chain config, and the hardhat plugin allow you to set it in the hardhat config, and then pass it to core.

I think this is important because it's common to deploy the same thing to multiple chains.

WDYT?

packages/core/src/verify.ts
```diff
@@ -0,0 +1,77 @@
...
+    verifyResult.push({
+      address: contract.address,
+      compilerVersion: buildInfo.solcLongVersion.startsWith("v")
+        ? buildInfo.solcLongVersion
+        : `v${buildInfo.solcLongVersion}`,
+      sourceCode: JSON.stringify(buildInfo.input),
+      name: `${artifact.sourceName}:${contractName}`,
+      args: encodeDeploymentArguments(artifact, constructorArgs),
+    });
+  }
```
alcuadrado, line 74: I'm not entirely sure if this won't lead to OOM errors. It doesn't store the entire build info, so the chances are lower though.

packages/hardhat-plugin/package.json
```diff
@@ -92,6 +92,7 @@
     "hardhat": "^2.18.0"
   },
   "dependencies": {
+    "@nomicfoundation/hardhat-verify": "^2.0.1",
```
alcuadrado, line 95: It should also be a peer dependency

packages/core/src/verify.ts
kanej: Would switching to the core function to an async generator that yields one verify result at a time resolve this?

packages/core/src/verify.ts
kanej: I think we should expand the list of chains to match the HH-viem list, and support the "other networks" config setup that it supports:

https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#adding-support-for-other-networks

packages/hardhat-plugin/src/utils/verifyEtherscanContract.ts
kanej: Looking at the implementation in HH-viem, it already polls. @schaable is the 1s timeout necessary or can we call `getVerificationStatus` immediately?

schaable: It shouldn't be necessary. `verify` returns immediately, and `getVerificationStatus` does the polling.

kanej:
I did a round of manual testing today (all against sepolia), and was able to:

* verify multiple contracts in one deployment
* library + contract mix
* library + contract where the library was used in the contract constructor
* using a custom chain
* Already verified

@zoeyTM can I ask for several changes:

* If the user adds `--verify` can we do the equivalent of the validation checks in the verify task at the start of deploy (so no deploy if no etherscan api key)
* in the case where `--verify` is passed can we separate the deployed addresses from the verifying lines with a new heading, so:
image
Becomes something like:
```
LockModule#WAIIT...

Verifying deployed contracts

Verifying contract "contracts/..."
```
* In the verifying contract text can we replace the FQN with the future id:
image
Becomes something like:
```
Verifying contract for LockModule#WAIIIT ...
```

This is looking good.

zoeyTM:
@kanej 

> - If the user adds --verify can we do the equivalent of the validation checks in the verify task at the start of deploy (so no deploy if no etherscan api key)
> - Support etherscan api key for sepolia network as discussed
> - in the case where --verify is passed can we separate the deployed addresses from the verifying lines with a new heading

All three of these are complete now, however:

> - In the verifying contract text can we replace the FQN with the future id

I have two concerns regarding this:

1. My understanding is that futures will not always map 1 to 1 with contracts, and a single future id could therefore result in multiple contract deployments, which, with this change, would make the verify messaging ambiguous for these types of futures
2. While futures are our internal construct for managing and executing a deployment, in my opinion, that distinction ends after the deployment completes. Verify as a task has no concept of futures, and outside the context of an Ignition deployment, I believe that the FQN conveys more meaningful information for interactions with Etherscan

kanej:
> @kanej
> 
> > * If the user adds --verify can we do the equivalent of the validation checks in the verify task at the start of deploy (so no deploy if no etherscan api key)
> > * Support etherscan api key for sepolia network as discussed
> > * in the case where --verify is passed can we separate the deployed addresses from the verifying lines with a new heading
> 
> All three of these are complete now, however:
> 
> > * In the verifying contract text can we replace the FQN with the future id
> 
> I have two concerns regarding this:
> 
>     1. My understanding is that futures will not always map 1 to 1 with contracts, and a single future id could therefore result in multiple contract deployments, which, with this change, would make the verify messaging ambiguous for these types of futures
> 
>     2. While futures are our internal construct for managing and executing a deployment, in my opinion, that distinction ends after the deployment completes. Verify as a task has no concept of futures, and outside the context of an Ignition deployment, I believe that the FQN conveys more meaningful information for interactions with Etherscan

Those are fair points. Lets leave it for now.

kanej:
I reviewed the library changes where the info is read from the deployment. I made a change to read the more specific info for the libraries from the execution state, rather than all deployed contracts.

packages/core/src/verify.ts
```diff
@@ -0,0 +1,166 @@
...
+  SourceToLibraryToAddress,
+  VerifyResult,
+} from "./types/verify";
+
+/**
+ * Retrieve the information required to verify all contracts from a deployment on Etherscan.
+ *
+ * @param deploymentDir - the file directory of the deployment
+ * @param customChains - an array of custom chain configurations
+ * @param libraryInfo - an object containing library names and addresses
```
alcuadrado, line 27:
```suggestion
```

packages/core/src/index.ts
```diff
@@ -15,4 +15,6 @@ export * from "./types/module-builder";
 export * from "./types/provider";
 export * from "./types/serialization";
 export * from "./types/status";
+export * from "./types/verify";
+export { verify } from "./verify";
```
alcuadrado, line 19: I found this name a bit confusing because you are the one who is doing the verification, and this verb seems to indicate that calling it will run the verification. Let's rename it to something like `getVerificationInformation`.

packages/core/src/verify.ts
alcuadrado: (this UI can be a bit confusing, it's a deletion because we don't have that param anymore)

packages/hardhat-plugin/src/index.ts
```diff
@@ -311,6 +338,90 @@ ignitionScope
...
+ignitionScope
+  .task("verify")
+  .addPositionalParam("deploymentId", "The id of the deployment to verify")
+  .setDescription(
+    "Verify contracts from a deployment against the configured block explorers"
+  )
+  .setAction(async ({ deploymentId }: { deploymentId: string }, hre) => {
+    const { verify } = await import("@nomicfoundation/ignition-core");
+    const { Etherscan } = await import(
+      "@nomicfoundation/hardhat-verify/etherscan"
```
alcuadrado, line 350: This should be imported into the top-level scope of this file. It will import its type-extensions, so we wouldn't need ours.

If you depend on another plugin, you should always do this, as you may be depending on one of its config validation/environment extensions/tasks/etc.

packages/hardhat-plugin/src/index.ts
```diff
@@ -311,6 +338,90 @@ ignitionScope
...
+    );
+    const { getApiKeyAndUrls } = await import("./utils/getApiKeyAndUrls");
+    const { shouldBeHardhatPluginError } = await import(
+      "./utils/shouldBeHardhatPluginError"
+    );
+    const { verifyEtherscanContract } = await import(
+      "./utils/verifyEtherscanContract"
```
alcuadrado, line 357: We should be ok with importing this top-level, if they aren't importing anything slow in their top-level.

packages/core/src/verify.ts
```diff
@@ -0,0 +1,166 @@
...
+    artifact,
+    libraries
+  );
+
+  if (sourceToLibraryAddresses === null) {
+    return JSON.stringify(buildInfo.input);
+  }
+
+  const { input } = buildInfo;
+  input.settings.libraries = sourceToLibraryAddresses;
```
alcuadrado, line 135: This will lead to a different metadata hash. If we can avoid this, we should. Can we pass the libraries to the `Etherscan` object as a separate param?

zoeyTM: Unfortunately the `Etherscan` class doesn't accept libraries as a param, however, `hardhat-verify` seems to be adding library info [the same way](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-verify/src/internal/tasks/etherscan.ts#L298).

packages/hardhat-plugin/src/index.ts
kanej: I have pulled it in directly. Note this required me to bump the version of HH toolbox used in the samples.

PR #635, opened 2023-12-01
chore: add a script for updating versions

kanej:
Previously publishing required doing an npm version update then manual setting of dependencies.

Instead now, we run a script that:

* performs the npm version call
* reads the new version and applies it to the dependencies
* run a final npm install to update the package.json

To update the version run:

```shell
npm run update-version patch #minor #major
```


alcuadrado:
Why this instead of changesets? I think it'd be simpler

kanej:
Lets go with changeset.

PR #638, opened 2023-12-07
chore: cleanup eslint config in examples

kanej:
We were seeing errors on some sample projects eslinting in vscode. This was due to a misconfiguraiont in their eslintrc that was pulling in the tsconfig wrong. Eslint then fell back to the root repo directories tsconfig, which doesn't exist - hence the error.

The solution was to remove the failing import and rely on the default.

I also excluded the eslint config file from eslint for each of the examples.


zoeyTM (approved):
Great find! LGTM

PR #651, opened 2023-12-12
Add --reset flag to deploy task

### LINKED ISSUE #649: Add `--reset` flag to `deploy` task
When deploying against a local Hardhat node (chainId: 31337), a deployment folder will be created containing useful information like the contract abis.

When developing against smart contract deployed locally via Ignition (e.g. with a wagmi frontend), it is cumbersome to have to `hardhat ignition wipe <deploymentid> <futureId>` before redeploying your contract every time you make a modification.

One approach would be a `--reset` flag on the `deploy` that clears the previous deployment before redeploying:

```shell
npx hardhat ignition deploy ./ignition/modules/Apollo.ts --reset --network localhost
```

A confirmation message should be shown for network's that have a `chainId` other than 31337.

zoeyTM:
Adds a `--reset` flag to wipe the deployments state before a deployment. Also includes a confirmation when used on a non-hardhat network.

resolves #649


zoeyTM:
Screen Shot 2023-12-11 at 10 24 18 PM

packages/hardhat-plugin-viem/scripts/compile-test-fixture-projects.ts
```diff
@@ -18,10 +18,6 @@ const main = async () => {
       fixtureProjectName
     );
 
-    if (fs.existsSync(path.join(fixtureProjectDir, "./artifacts"))) {
```
kanej, line 21: Was this causing an issue?

The intent was that you only need to do a precompile of the fixture projects to get the viem package build working once - to generate the contract type information. Is there a case I didn't understand?

packages/hardhat-plugin/src/index.ts
```diff
@@ -123,6 +126,13 @@ ignitionScope
...
 
+      const deploymentId = resolveDeploymentId(givenDeploymentId, chainId);
+
+      const deploymentDir =
+        hre.network.name === "hardhat"
+          ? undefined
+          : path.join(hre.config.paths.ignition, "deployments", deploymentId);
+
       if (chainId !== 31337) {
         const prompt = await Prompt({
```
kanej, line 137: Not for this PR. I think we should look at removing the prompt dependency. We are doing enough low level console work to justify putting in the equivalent to let us drop a dep.

kanej (changes requested):
Looks good.

Can I request two changes:

1. Can you add an `@nomicfoundation/hardhat-plugin` file based project integration test that: runs a deployment that succeeds, then reruns a deployment modified in such a way (e.g. same future id but different contract) it would be blocked if rerun. But will succeed when run with the reset flag.
2. Can you add a warning message and cancel if you run reset with an in-memory deployment.

If you have a different approach for the test we should consider it as I suspect we might run a fowl of the clearLine bug.

packages/hardhat-plugin-viem/scripts/compile-test-fixture-projects.ts
kanej: Out of band we discussed and agreed to remove the line.

If you are making changes to the viem tests this can lead to odd results.

packages/hardhat-plugin/test/test-helpers/use-ignition-project.ts
```diff
@@ -62,7 +62,10 @@ export function useFileIgnitionProject(
     const hre = require("hardhat");
 
     const deploymentDir = path.join(
-      path.resolve(__dirname, "../fixture-projects/minimal/"),
+      path.resolve(
+        __dirname,
+        `../fixture-projects/${fixtureProjectName}/ignition`
```
kanej, line 67: Well ...

kanej (approved):

kanej:
I have done of manual tests. No issues against local/sepolia.

PR #653, opened 2023-12-13
Wipe deployment state if hardhat-network has been restarted

### LINKED ISSUE #650: Deploying to a cleared local hardhat node should ignore previous deployment
Ignition treats deploying to a local Hardhat network node (chainId: 31337) like other networks. It creates deployment files under the `./ignition/deployments` folder.

However, if the Hardhat node is stopped and restarted, then the deployment state will need to be wiped to allow for a redeployment of the module.

It would be better if, when deploying against a network with chainId 31337, Ignition automatically detects that the network node has restarted since the previous run. If the network node has restarted, the deployment should be reset with a complete wipe of the previous deployment state.

## A potential approach

Pull the Hardhat node instance id if the network has chainId 31337:

```typescript
const metadata = await network.provider.send("hardhat_metadata");
const instanceId = metadata.instanceId; // random identifier of the session
```

Then to record that instance id against the deployment initialize message to the journal

```ts
export interface DeploymentInitializeMessage {
  type: JournalMessageType.DEPLOYMENT_INITIALIZE;
  chainId: number;
  instanceId?: number;
}
```

On a rerun of the deployment, Ignition should compare the instance ids, if they differ, the deployment state should be cleared.

zoeyTM:
In the past, we've had the intention of keeping ignition-core agnostic of hardhat, however, with this issue, while we do need to store this piece of data (`instanceId`) somewhere to compare against future runs and journal does seem like the most appropriate place, that would also require breaking that separation of intent between ignition-core and hardhat as `instanceId` is a hardhat-specific piece of data. we would also need to add a hardhat-specific method to `JsonRpcClient` to retrieve the `instanceId`

To maintain that separation, I went a different direction with this implementation. The logic for this feature exists entirely inside `hardhat-plugin`, but it does require adding one new (small) file to the `ignition/deployments` directory to map deployment IDs to hardhat instance IDs. I think this approach is preferable since this is a hardhat-specific feature to begin with.

This implementation does have one caveat, and that's in the case of a user 

1. with an existing deployment to a non-hardhat chain 
2. that deployment uses a custom `deploymentId`
3. the user tries deploying to hardhat-network using _**the same custom `deploymentId`**_

If those criteria are met, the existing deployment to a non-hardhat chain would be wiped. I think this is a bit of a contrived edge case, so I'm okay with leaving that possibility if needed, but any suggestions on removing that possibility would be more than welcome as well.

resolves #650


zoeyTM:
Also closes #677

packages/hardhat-plugin/src/index.ts
```diff
@@ -163,6 +172,32 @@ ignitionScope
...
+
+        const instanceFile: {
+          [deploymentId: string]: string;
+        } = instanceFileExists ? require(instanceFilePath) : {};
+
+        const metadata = (await hre.network.provider.request({
+          method: "hardhat_metadata",
+        })) as { instanceId: string };
+
+        if (instanceFile[deploymentId] !== metadata.instanceId) {
```
alcuadrado, line 193:
```suggestion
        if (instanceFile[deploymentId] !== undefined && instanceFile[deploymentId] !== metadata.instanceId) {
```

PR #654, opened 2023-12-13
Use readline in place of process.stdout

### LINKED ISSUE #644: TypeError: `process.stdout.clearLine` is not a function
### Description : 

Its seems `process.stdout.clearLine` might not work in some OS / node versions, recently encountered this error in our gh-actions CI job : 
https://github.com/scaffold-eth/scaffold-eth-2/actions/runs/7158553408/job/19490782791

https://github.com/NomicFoundation/hardhat-ignition/blob/4e06abfacc3638913d6698c9e429a4a393bb09c0/packages/hardhat-plugin/src/ui/pretty-event-handler.ts#L455

I think this was the cause in our case

### Similar issues reported on other repo's : 

https://forum.cloudron.io/topic/9219/calling-build-inside-a-docker-container

https://github.com/CookPete/auto-changelog/issues/73

https://github.com/hallh/elasticsearch-primary-balancer/issues/9

### Possible solution is to use `clearLine` from node's  [`readline`](https://nodejs.org/api/readline.html): 

```ts
import { clearLine } from 'readline'

clearLine(process.stdout, 0)
```

It was suggested here : https://stackoverflow.com/questions/34570452/node-js-stdout-clearline-and-cursorto-functions#comment80319576_34570694

zoeyTM:
- replaced `process.stdout` in favor of native nodejs `readline`
- also removed a redundant github workflow that ran integration tests separately since we run them as part of the top level `npm test` now

resolves #644

### DIFF
--- a/packages/core/test-integrations/new-api/ci/basic-success.ts
+++ b/packages/core/test-integrations/new-api/ci/basic-success.ts
@@ -3,13 +3,14 @@ import { assert } from "chai";
 import { useHardhatProject } from "../../helpers/hardhat-projects";
 
 // This test exists to ensure Ignition succeeds in a CI environment.
+// This is a test that the UI, runs even in constrained terminal environments.
 // It should always pass locally.
-describe("CI - basic success case", function () {
+describe("Running deployment in CI environment", function () {
   this.timeout(60000);
 
   useHardhatProject("ci-success");
 
-  it("should succeed in a CI environment", async function () {
+  it("should succeed with UI in a CI environment", async function () {
     await assert.isFulfilled(
       this.hre.run(
         { scope: "ignition", task: "deploy" },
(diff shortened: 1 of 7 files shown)


packages/core/test-integrations/new-api/ci/basic-success.ts
```diff
@@ -0,0 +1,23 @@
+import { assert } from "chai";
+
+import { useHardhatProject } from "../../helpers/hardhat-projects";
+
+// This test exists to ensure Ignition succeeds in a CI environment.
+// This is a test that the UI, runs even in constrained terminal environments.
+// It should always pass locally.
+describe("Running deployment in CI environment", function () {
```
kanej, line 8: Nicely done on the test.

kanej (approved):
This looks good.

Lets (squash) merge this!

alcuadrado (approved):

PR #655, opened 2023-12-14
Add sample project for upgradeable proxy usage

### LINKED ISSUE #641: Provide sample project for open zeppelin upgrade contracts
Create a sample project to show how to deploy an upgradable contract in one ignition module and a separate ignition module for doing a further upgrade.

zoeyTM:
resolves #641


examples/upgradeable/README.md
```diff
@@ -0,0 +1,19 @@
+# Hardhat Sample for Hardhat Ignition
+
+This project is Hardhat's sample project enhanced with Hardhat Ignition.
```
kanej, line 3: We should update this intro to explain that we are showing upgradeable contracts.

examples/upgradeable/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Deploying
+
+To run the Ignition deploy against the ephemeral hardhat network:
+
+```shell
+npx hardhat ignition deploy ./ignition/modules/LockModule.js
```
kanej, line 10: This needs updated

examples/upgradeable/README.md
zoeyTM: Oh whoops! good catch

examples/upgradeable/ignition/modules/ProxyModule.js
```diff
@@ -0,0 +1,85 @@
+// ./ignition/LockModule.js
```
alcuadrado, line 1:
```suggestion
```

examples/upgradeable/ignition/modules/ProxyModule.js
```diff
@@ -0,0 +1,85 @@
...
+  // The TransparentUpgradeableProxy contract creates the ProxyAdmin within its constructor.
+  // To read more about how this proxy is implemented, you can view the source code and comments here:
+  // https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.1/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
+  const proxy = m.contract("TransparentUpgradeableProxy", [
+    demo,
+    proxyAdminOwner,
+    "0x",
+  ]);
+
+  // We need to get the address of the ProxyAdmin contrac that was created by the TransparentUpgradeableProxy
```
alcuadrado, line 26:
```suggestion
  // We need to get the address of the ProxyAdmin contract that was created by the TransparentUpgradeableProxy
```

jkirkpatrick:
OZ upgradable contracts should not accept constructor parameters... How do pass parameters when using a custom initializer? How do you specify the initializer or upgrade functions, e.g. `updateToAndCall`?

examples/upgradeable/ignition/modules/ProxyModule.js
```diff
@@ -0,0 +1,59 @@
...
+  // to treat the contract at the proxy address as an instance of the Demo contract.
+  // This allows us to interact with the underlying Demo contract via the proxy from within tests and scripts.
+  const demo = m.contractAt("Demo", proxy);
+
+  // Return the contract instance, along with the original proxy and proxyAdmin contracts
+  // so that they can be used by other modules, or in tests and scripts.
+  return { demo, proxy, proxyAdmin };
+});
+
+module.exports = demoModule;
```
devon-n, line 59: demoModule is being exported here but the `UpgradeModule.js` imports it as ProxyModule which is the name of the not exported module. This could be confusing.

examples/upgradeable/test/ProxyDemo.js
devon-n, line 1: The tests don't upgrade the existing contract but deploys a completely new contract that is upgraded inside its own describe.

Could a `beforeEach()` could be used to show upgrading an existing instance of a contract.

Where the first version of the contract is used deployed in the `beforeEach()` then is upgraded elsewhere.

Or could `ignition.deploy(ProxyModule);` be used to get the first version of the contract and upgrade that version using ignition modules?

zoeyTM: The tests are testing exactly what they are meant to. There is no fundamental difference between what you're describing and what's already happening. This is a non-issue.

examples/upgradeable/README.md
```diff
@@ -0,0 +1,25 @@
...
+To deploy the an example proxy contract against the ephemeral Hardhat network:
+
+```shell
+npx hardhat ignition deploy ./ignition/modules/ProxyModule.js
+```
+
+To deploy an example of a proxy contract being upgraded against the ephemeral Hardhat network:
+
+```shell
+npx hardhat ignition deploy ./ignition/modules/UpgradeModule.js
```
kanej, line 16: I split these two calls apart @zoeyTM

kanej (approved):
I did a run through locally.
I pushed a few changes, specifically:

1. splitting up the commands in the readme
2. including the example in the `pnpm` workspace ([this commit](https://github.com/NomicFoundation/hardhat-ignition/pull/655/commits/b6731772ba49a688307c6375cf676843222d104d))
3. Removing unused variables in the test file

I think all that is left is to rebase on development and bump the version of the example to the latest `0.15.5`.

zoeyTM:
@kanej merging because it's already rebased and it doesn't look like we update the versions of the example projects. We can update them later all at once if we want. 

I also tried adding the ENS project to the workspace, but it's still bugged. I briefly tried to fix it, but it seems like a much more involved dive is needed and it's not worth the time right now.

PR #657, opened 2023-12-14
repository: add devcontainer support.

marianfe:
Adding .devcontainer support for the repository. 

Tested by creating a new named container in vscode based off of this branch, and verifying: 
  npm install
  npm run build
  npm run test

### DIFF
--- a/packages/hardhat-plugin/README.md
+++ b/packages/hardhat-plugin/README.md
@@ -1,5 +1,6 @@
 ![hardhat_Ignition_banner](https://github.com/NomicFoundation/hardhat-ignition/assets/24030/cc73227b-8791-4bb3-bc9a-a39be69d215f)
 [![npm](https://img.shields.io/npm/v/@nomicfoundation/hardhat-ignition.svg)](https://www.npmjs.com/package/@nomicfoundation/hardhat-ignition) [![hardhat](https://hardhat.org/buidler-plugin-badge.svg?1)](https://hardhat.org) [![Open in Dev Containers](https://img.shields.io/static/v1?label=Dev%20Containers&message=Open&color=blue&logo=visualstudiocode)](https://vscode.dev/redirect?url=vscode://ms-vscode-remote.remote-containers/cloneInVolume?url=https://github.com/NomicFoundation/hardhat-ignition.git)
+
 ---
 
 # Hardhat Ignition


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "Hardhat Ignition",
+    "image": "mcr.microsoft.com/devcontainers/base:bullseye",
+    "features": {
+        "ghcr.io/devcontainers/features/node:1": {
+          "version": "16" /* Keep in sync with the oldest version of Node.js that Hardhat supports */
```
kanej, line 6: We should bump this to node 18.

Node 16 EOL got brought forward: https://nodejs.org/en/blog/announcements/nodejs16-eol

Our supported Node.js versions are those in Active LTS or maintenance LTS + 2 months after they leave maintenance LTS.

Node 16 has been 2 months out of maintenance LTS as of a few weeks ago.

kanej (changes requested):
What do you think of running `npm install` as a post create command?

```json
{
  // ...
  "postCreateCommand": "npm install"
}
```

This eliminates the install step as a first action in a new container.

kanej:
What do we think of adding a `open in dev container` badge to the root repo readme?

https://code.visualstudio.com/docs/devcontainers/create-dev-container#_add-an-open-in-dev-container-badge

kanej:
Is there anything we need to consider or do to support devs using a devcontainer but bringing their own local configuration?

This is not something I have looked into but maybe there is a best practice guide we can find/follow?

kanej (approved):

kanej (approved):
Lets get this (squash) merged!

kanej:
> > Is there anything we need to consider or do to support devs using a devcontainer but bringing their own local configuration?
> > This is not something I have looked into but maybe there is a best practice guide we can find/follow?
> 
> As far as i can see, the standard recommended approach for customizing containers seems to be to push dotfiles to containers as described here: this https://code.visualstudio.com/docs/devcontainers/containers#_personalizing-with-dotfile-repositories
> 
> if i interpret this recommendation correctly, i would think that we don't need to do anything specific within our repo or devcontainer, instead this setup is left up to each dev. e.g. our repo should have the devcontainer json config file that i am adding now, and then each dev should follow that guide to set up custom dotfiles (and then push them automatically to the container via the method explained there). however @alcuadrado is probably more of an expert, adding him here for an opinion? thanks!

Thanks @marianfe-n for looking at that. Based on the vscode docs I agree, we don't need to make any changes to this PR, it is good to go.

PR #674, opened 2024-01-15
Add jsdoc comments for all module builder and ignition helper functions

### LINKED ISSUE #642: Ensure all module api and ignition helper classes are documented
To better support intellisense discovery of how to leverage different parts of the api we should document with appropriate comments all the methods/functions in our api.

## TODO

- [x] Each method on the ModuleBuilder object injected as part of buildModule should be documented
- [x] Do a run through of locally developing some contracts and then iterate on the docs to best support new users to Ignition

zoeyTM:
resolves #642

### DIFF
--- a/packages/core/src/types/module-builder.ts
+++ b/packages/core/src/types/module-builder.ts
@@ -128,19 +128,70 @@ export interface SendDataOptions {
  * @beta
  */
 export interface IgnitionModuleBuilder {
+  /**
+   * Returns an account runtime value that can be used to specify the
+   * account to use for a transaction.
+   *
+   * @param accountIndex - The index of the account to return
+   *
+   * @example
+   * ```
+   * const owner = m.getAccount(1);
+   * const myContract = m.contract("MyContract", { from: owner });
+   * // can also be used anywhere an address is expected
+   * m.send("sendToOwner", owner, { value: 100 });
+   * ```
+   */
   getAccount(accountIndex: number): AccountRuntimeValue;
 
+  /**
+   * Allows you to specify a parameter that can be dynamically set during deployment.
+   *
+   * @param parameterName - The name of the parameter
+   * @param defaultValue - The default value to use if the parameter is not provided
+   *
+   * @example
+   * ```
+   * const amount = m.getParameter("amount", 100);
+   * const myContract = m.contract("MyContract", { value: amount });
+   * ```
+   */
   getParameter<ParamTypeT extends ModuleParameterType = any>(
     parameterName: string,
     defaultValue?: ParamTypeT
   ): ModuleParameterRuntimeValue<ParamTypeT>;
 
+  /**
+   * Deploys a contract.
+   *
+   * @param contractName - The name of the contract to deploy
+   * @param args - The arguments to pass to the contract constructor
+   * @param options - The options for the deployment
+   *
+   * @example
+   * ```
+   * const myContract = m.contract("MyContract", [], { value: 100 });
+   * ```
+   */
   contract<ContractNameT extends string>(
     contractName: ContractNameT,
     args?: ArgumentType[],
     options?: ContractOptions
   ): NamedArtifactContractDeploymentFuture<ContractNameT>;
 
+  /**
+   * Deploys a contract.
+   *
+   * @param contractName - The name of the contract to deploy
+   * @param artifact - The artifact of the contract to deploy
+   * @param args - The arguments to pass to the contract constructor
+   * @param options - The options for the deployment
+   *
+   * @example
+   * ```
+   * const myContract = m.contract("MyContract", [], { value: 100 });
+   * ```
+   */
   contract<const AbiT extends Abi>(
     contractName: string,
     artifact: Artifact<AbiT>,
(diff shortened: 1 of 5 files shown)


packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
  * @beta
  */
 export interface IgnitionModuleBuilder {
+  /**
+   * Returns an account runtime value that can be used to specify the
+   * account to use for a transaction.
```
kanej, line 133:
```suggestion
   * Returns an account runtime value representing the 
   * Hardhat account the underlying transactions for a 
   * future will be sent from
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   * const owner = m.getAccount(1);
+   * const myContract = m.contract("MyContract", { from: owner });
+   * // can also be used anywhere an address is expected
+   * m.send("sendToOwner", owner, { value: 100 });
+   * ```
+   */
   getAccount(accountIndex: number): AccountRuntimeValue;
 
+  /**
+   * Allows you to specify a parameter that can be dynamically set during deployment.
```
kanej, line 148:
```suggestion
   * A parameter whose value can be set at deployment.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   * const myContract = m.contract("MyContract", { value: amount });
+   * ```
+   */
   getParameter<ParamTypeT extends ModuleParameterType = any>(
     parameterName: string,
     defaultValue?: ParamTypeT
   ): ModuleParameterRuntimeValue<ParamTypeT>;
 
+  /**
+   * Deploys a contract.
```
kanej, line 165:
```suggestion
   * Deploy a contract.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   *
+   * @param contractName - The name of the contract to deploy
+   * @param args - The arguments to pass to the contract constructor
+   * @param options - The options for the deployment
```
kanej, line 169: Can we break out the options and document them.

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   * ```
+   */
   contract<ContractNameT extends string>(
     contractName: ContractNameT,
     args?: ArgumentType[],
     options?: ContractOptions
   ): NamedArtifactContractDeploymentFuture<ContractNameT>;
 
+  /**
+   * Deploys a contract.
```
kanej, line 183:
```suggestion
   * Deploy a contract.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   */
   contract<const AbiT extends Abi>(
     contractName: string,
     artifact: Artifact<AbiT>,
     args?: ArgumentType[],
     options?: ContractOptions
   ): ContractDeploymentFuture<AbiT>;
 
+  /**
+   * Deploys a library.
```
kanej, line 203:
```suggestion
   * Deploy a library.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   * const myLibrary = m.library("MyLibrary", { from: owner } );
+   * ```
+   */
   library<LibraryNameT extends string>(
     libraryName: LibraryNameT,
     options?: LibraryOptions
   ): NamedArtifactLibraryDeploymentFuture<LibraryNameT>;
 
+  /**
+   * Deploys a library.
```
kanej, line 220:
```suggestion
   * Deploy a library.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   * ```
+   */
   library<const AbiT extends Abi>(
     libraryName: string,
     artifact: Artifact<AbiT>,
     options?: LibraryOptions
   ): LibraryDeploymentFuture<AbiT>;
 
+  /**
+   * Calls a contract function.
```
kanej, line 239:
```suggestion
   * Call a contract function.
```

packages/core/src/types/module-builder.ts
```diff
@@ -128,44 +128,149 @@ export interface SendDataOptions {
...
+   */
   call<ContractNameT extends string, FunctionNameT extends string>(
     contractFuture: CallableContractFuture<ContractNameT>,
     functionName: FunctionNameT,
     args?: ArgumentType[],
     options?: CallOptions
   ): ContractCallFuture<ContractNameT, FunctionNameT>;
 
+  /**
+   * Statically calls a readonly contract function and returns the result.
```
kanej, line 260: We should add a bit of explanation on the the following line about what a static call is, and that it won't update the blockchain.

packages/core/src/types/module-builder.ts
```diff
@@ -174,6 +279,18 @@ export interface IgnitionModuleBuilder {
     options?: StaticCallOptions
   ): StaticCallFuture<ContractNameT, FunctionNameT>;
 
+  /**
+   * Creates a contract instance at the specified address.
```
kanej, line 283: I think this should be more: 
```Creates a future representing an existing deployed contract so that it can be referenced in subsequent futures.```

What do you think?

File: packages/core/src/types/module-builder.ts [hunk 11]
```diff
@@ -193,6 +323,32 @@ export interface IgnitionModuleBuilder {
     options?: ContractAtOptions
   ): ContractAtFuture<AbiT>;
 
+  /**
+   * Reads an event argument from a contract.
```
kanej, line 327: Maybe a follow up line, on why you want to do that, i.e. so you can use the value in subsequent futures as a parameter.

packages/core/src/types/module-builder.ts
```diff
@@ -204,6 +360,26 @@ export interface IgnitionModuleBuilder {
     options?: ReadEventArgumentOptions
   ): ReadEventArgumentFuture;
 
+  /**
+   * Sends a transaction.
```
kanej, line 364: We should expand on this to explain it allows you to send an arbitrary transaction, so you can transfer eth or invoke with raw data.

packages/core/src/types/module-builder.ts
```diff
@@ -216,6 +392,26 @@ export interface IgnitionModuleBuilder {
     options?: SendDataOptions
   ): SendDataFuture;
 
+  /**
+   * Allows you to use the results of another module within this module.
```
kanej, line 396:
```suggestion
   * Allows you to deploy then use the results of another module within this module.
```

packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
```diff
@@ -51,6 +51,13 @@ export class EthersIgnitionHelper {
     this._deploymentDir = deploymentDir;
   }
 
+  /**
+   * Deploys the given Ignition module and returns the results returned by the module.
```
kanej, line 55:
```suggestion
   * Deploys the given Ignition module and returns the results of the module as Ethers contract instances.
```

packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
```diff
@@ -45,6 +45,13 @@ export class ViemIgnitionHelper {
     this._deploymentDir = deploymentDir;
   }
 
+  /**
+   * Deploys the given Ignition module and returns the results returned by the module.
```
kanej, line 49:
```suggestion
   * Deploys the given Ignition module and returns the results of the module as Viem contract instances.
```

kanej (changes requested):
This is looking good.
I suspect we will be iterating on these docs forever, but I left a few suggestions.

packages/core/src/types/module-builder.ts
zoeyTM: I believe they are documented already, but I will double check

packages/core/src/types/module-builder.ts
zoeyTM: This was 100% a "best effort" attempt on my part as I couldn't really think of a succinct way of saying it.

kanej (approved):
LGTM!

PR #678, opened 2024-01-16
Made all errors internal

### LINKED ISSUE #581: Consider moving `core/src/errors.ts` and `core/src/error-list.ts` to internal.
I think these files may be exposing unnecessary things.

zoeyTM:
resolves #581


kanej (changes requested):
I removed the API tags (i.e. @alpha & @beta) for all the functions/constants in internal. This threw up a few failures:

image

For those on the list that are really types lets add a `./src/types/error.ts` with those. For the others, they should be moved back out of internal into the top level, with references to the internal elements.

The goal is to minimize what is publicly exposed and to be explicit between what is public/internal.

kanej (approved):
Looks good. Lets get this squash merged!

PR #679, opened 2024-01-18
Add the ability to use AccountRuntimeValues as default values for mod…

### LINKED ISSUE #673: Document optional owner usage
Many contracts take the owner as an input parameter. I want to be able to set the owner of these contracts dynamically, but fallback to the HardHat account.

## Current state

For a lot of these contracts, I would like to do something like this:

```typescript
const owner = m.getParameter<string>('owner', m.getAccount(0))
```

However, this isn't possible because the types are different (`string` and `AccountRuntimeValue` are not comparable, and there is no way to convert between the two types)

## Alternatives

Other than adding a way to convert between these types, another way to tackle this would be to instead encourage people to use indices instead of owners

i.e something like this:

```ts
const ownerIndex = m.getParameter<number>('ownerIndex', 0);
m.getAccount(ownerIndex );
```

This code won't work (for the same reason as above), but it could be that implementing conversion between numbers like this is easier than converting between accounts

zoeyTM:
Allows for the workflow of:
```
const owner = m.getParameter<string>('owner', m.getAccount(0))
```

Resolves #673

### DIFF
--- a/packages/core/src/types/module.ts
+++ b/packages/core/src/types/module.ts
@@ -395,7 +395,7 @@ export interface ModuleParameterRuntimeValue<
  * @beta
  */
 export interface ModuleParameters {
-  [parameterName: string]: ModuleParameterType;
+  [parameterName: string]: SolidityParameterType;
 }
 
 /**
--- a/packages/core/src/types/module.ts
+++ b/packages/core/src/types/module.ts
@@ -344,7 +344,7 @@ export type SolidityParameterType =
  *
  * @beta
  */
-export type ModuleParameterType = SolidityParameterType;
+export type ModuleParameterType = SolidityParameterType | AccountRuntimeValue;
 
 /**
  * The different runtime values supported by Ignition.
--- a/packages/core/src/ignition-module-serializer.ts
+++ b/packages/core/src/ignition-module-serializer.ts
@@ -804,7 +804,7 @@ export class IgnitionModuleDeserializer {
               )
             ? (this._deserializeModuleParameterRuntimeValue(
                 serializedFuture.address
-              ) as ModuleParameterRuntimeValue<string>) // This is unsafe, but we only serialize valid valus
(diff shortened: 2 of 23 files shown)


packages/core/src/types/module.ts
```diff
@@ -344,7 +344,7 @@ export type SolidityParameterType =
  *
  * @beta
  */
-export type ModuleParameterType = SolidityParameterType;
+export type ModuleParameterType = SolidityParameterType | AccountRuntimeValue;
```
kanej, line 347: What do you think of leaving `ModuleParameterType` unchanged but altering just default value to be `ModuleParameterType | AccountRuntimeValue | undefined` or whatever?

Are there any implications to expanding the `ModuleParameterType` to include `AccountRuntimeValue` outside of default value, or in the future?

kanej (commented):
I hadn't thought about it before but are there any implications for:

* Reconciliation
* Validation
* Serialization

packages/core/src/types/module.ts
zoeyTM: Hm, that's an interesting idea I hadn't thought about. In all honesty, I'm not sure what the implications might be.

kanej: Maybe one way to think about this is  are there any places the current change would allow us to use `AccountRuntimeValue` that are unintended?
I suspect changing the default value will lead to more code changes but less "conceptual  " changes.

zoeyTM: I've added a change in https://github.com/NomicFoundation/hardhat-ignition/pull/679/commits/9e480042f745a5c0e86d49e5191ed91976e7f9b5 that fixes the one place I was able to find that was impacted by changing `ModuleParameterType` as opposed to changing the default value type. I believe I've gone over anywhere relevant and this was the only place that would have been impacted.

Issue #676, opened 2024-01-15
Unexpected next nonce error


kanej:
While testing the new `create2` functionality I hit a next nonce error. I am believe that the actual issue is that some of the transactions are reverting but somehow we are triggering the next nonce error before displaying the revert message:

![image](https://github.com/NomicFoundation/hardhat-ignition/assets/24030/34d85fa1-9b20-4d76-aa34-52822035fc25)

## Reproduction Steps

In the `create2` branch, build then `cd` to the `ENS` example, and run the deploy with the `create2` strategy:

```shell
cd ./examples/ens
npx hardhat ignition deploy ./ignition/modules/ENS.js --strategy create2
```

zoeyTM:
I have confirmed a few things:

1. There is a race condition that happens when one transaction in a batch reverts that causes the nonce error above
2. The race condition is not specific to `create2` and does exist in the `basic` strategy as well
3. The reason the `ENS` module is failing is due to a permissions issue. Specifically, the owner of the ENS contract is set to `msg.sender` which, when deploying with `create2`, is the `CreateX` contract address.

PR #682, opened 2024-01-29
Error out when tx simulation fails

zoeyTM:
This is a possible fix for #676 

I would consider this a bit of a hacky fix, but it does make the error more clear. I'm not entirely certain what a more complete fix would look like, but there appear to be some inconsistencies within our engine code. For instance, simulation errors are a journal message, but we intentionally don't log simulation errors to journal (so they also don't currently update the front end UI code). 

My guess for a more complete solution would be that futures within the same batch should probably wait until at least the simulation succeeds on the previous futures transactions before processing the next future (i.e. future1 simulation succeeds before future2 requests a nonce). This would be a more significant change to the engine flow, however, so I'm not sure that's what we want right now.

### DIFF
--- a/packages/core/src/internal/execution/future-processor/handlers/send-transaction.ts
+++ b/packages/core/src/internal/execution/future-processor/handlers/send-transaction.ts
@@ -1,3 +1,6 @@
+import { IgnitionError } from "../../../../errors";
+import { ERRORS } from "../../../../errors-list";
+import { failedEvmExecutionResultToErrorDescription } from "../../../journal/utils/failedEvmExecutionResultToErrorDescription";
 import { assertIgnitionInvariant } from "../../../utils/assertions";
 import { JsonRpcClient } from "../../jsonrpc-client";
 import { NonceManager } from "../../nonce-management/json-rpc-nonce-manager";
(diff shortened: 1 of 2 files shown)


zoeyTM:
Screen Shot 2024-01-29 at 2 04 18 AM

kanej:
I will need to look through the code to reaquaint myself with the reasons for the different treatment of simulation errors.
In the meantime, is your mental model that you have a batch with two transactions:

```mermaid
flowchart LR
  subgraph Batch
    1[transaction #1]
    2[transaction #2]
  end
```

The execution of the batch starts and each is assigned a nonce, but #1 reverts with a failed simulation. We then get an issue because nonce #2 is invalid?

packages/core/src/internal/execution/future-processor/handlers/send-transaction.ts
```diff
@@ -120,10 +122,12 @@ export async function sendTransaction(
...
+  } else {
+    throw new IgnitionError(ERRORS.EXECUTION.SIMULATION_ERROR, {
+      error:
+        typeof result.error === "string"
+          ? result.error
+          : failedEvmExecutionResultToErrorDescription(result.error),
+    });
   }
-
-  return createExecutionStateCompleteMessageForExecutionsWithOnchainInteractions(
```
kanej, line 125: As I recall (with all that implies), I think we intentionally exclude simulation errors from the journal to allow for rerunning during debugging without increasing the size of the journal.

Is it possible we can trigger the equivalent event as would be triggered if we did record it to the journal?

zoeyTM:
Closing in favor of https://github.com/NomicFoundation/hardhat-ignition/pull/683

PR #684, opened 2024-01-30
Added a test to ensure simulation errors are displayed

### LINKED ISSUE #627: gas estimation fallback simulation reason fails on testnet/main
https://github.com/NomicFoundation/hardhat-ignition/blob/4e17b268ad0409d08d9a3e21b737e4f51200c4fb/packages/core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts#L141

If the gas estimation fails, we simulate the call to get revert reason. However, we don't pass along a gas limit as by definition the gas estimation has failed. This is usually fine in a ephemeral or HH network because the standard accounts have so much eth against them. On mainnet or a testnet, you are unlikely to be so lucky, hence the simulation call will fail without enough funds.

zoeyTM:
Resolves #627 

Simulating calls doesn't require any gas, but added a test anyways just to cover the case.

### DIFF
--- a/packages/core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts
+++ b/packages/core/src/internal/execution/future-processor/helpers/network-interaction-execution.ts
@@ -133,6 +133,11 @@ export async function sendTransactionForOnchainInteraction(
   try {
     gasLimit = await client.estimateGas(estimateGasPrams);
   } catch (error) {
+    // We remove the fees before simulating the transaction since we weren't able to estimate gas
+    // and otherwise geth will try to use the block gas limit as `gas`, and most accounts don't
+    // have enough balance to pay for that.
+    const { fees: _fees, ...paramsWithoutFees } = estimateGasPrams;
+
     // If the gas estimation failed, we simulate the transaction to get information
     // about why it failed.
     //
@@ -140,7 +145,7 @@ export async function sendTransactionForOnchainInteraction(
     // too broad and make the assertion below fail. We could try to catch only
     // estimation errors.
     const failedEstimateGasSimulationResult = await client.call(
-      estimateGasPrams, // TODO: we need to set a gas limit here, or the simulation could fail due to a lack of funds
+      paramsWithoutFees,
       "pending"
     );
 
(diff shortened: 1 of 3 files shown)


kanej:
Hey, I took the Rocket example, included your reversion in the constructor and ran it against sepolia, and got an error:

image

Is this a different issue?

zoeyTM:
> Hey, I took the Rocket example, included your reversion in the constructor and ran it against sepolia, and got an error:
> 
> image
> 
> Is this a different issue?

The reason this was happening is linked to the original issue. Gas estimation fails in this particular code path, but we were still sending the estimated `fees` object along with the rest of the transaction params when attempting to simulate the transaction. Because we included `fees` but not `gasLimit`, the node (or provider?) assigned one by default and simulates the maximum possible fee (default `gasLimit` times `maxFeePerGas`), which then triggered the insufficient funds error you linked above. The reason this only happened on testnet/mainnet and not locally (as in my test case) is (I believe) because the local nodes default `gasLimit` and `fees` are drastically lower, so the max possible fee wasn't high enough to trigger insufficient funds.

My solution was to delete the `fees` from the gas estimation params before simulating the transaction. I confirmed that this fixed the problem when deploying to Sepolia, and that the insufficient funds error is no longer thrown.

kanej:
> > Hey, I took the Rocket example, included your reversion in the constructor and ran it against sepolia, and got an error:
> > image
> > Is this a different issue?
> 
> The reason this was happening is linked to the original issue. Gas estimation fails in this particular code path, but we were still sending the estimated `fees` object along with the rest of the transaction params when attempting to simulate the transaction. Because we included `fees` but not `gasLimit`, the node (or provider?) assigned one by default and simulates the maximum possible fee (default `gasLimit` times `maxFeePerGas`), which then triggered the insufficient funds error you linked above. The reason this only happened on testnet/mainnet and not locally (as in my test case) is (I believe) because the local nodes default `gasLimit` and `fees` are drastically lower, so the max possible fee wasn't high enough to trigger insufficient funds.
> 
> My solution was to delete the `fees` from the gas estimation params before simulating the transaction. I confirmed that this fixed the problem when deploying to Sepolia, and that the insufficient funds error is no longer thrown.

To check my understanding. When we submit the simulation call with fees but without a gas entry, we hit the bug, as the live network (i.e. geth) will use the block gas limit from the pending block (according to https://ethereum.org/en/developers/docs/apis/json-rpc#eth_estimategasrpc#eth_estimategas) leading to too high a fee and the error above. Whereas this is not happening against the hardhat node.

I wonder whether there is a config we can set that will allow us to simulate the error on HH so we can capture it in a test. We can check that with the Hardhat team.

PR #690, opened 2024-02-07
Validate constructor/call args against ABI types

### LINKED ISSUE #7532: Ignition - better ABI Validation: check type of constructor/call args against abi
Add to the general validation phase a check that each argument in the call/constructor args of the relevant futures matches the type expected by the abi.

## TODO

- [ ] Draw up a test plan that will effectively act as a specification
- [ ] Spike type validation for single future (i.e. named contract deploy)

zoeyTM:


kanej:
We are closing this as we are porting.

PR #696, opened 2024-02-14
Strategies config

### LINKED ISSUE #694: Create2 Configuration
The `create2` branch needs to be enhanced with a configuration option for the *salt* before it can ship (see https://github.com/NomicFoundation/hardhat-ignition/pull/626).

The design leverages strategy config stored in `hardhat.config.{js,ts}` to inject the *salt* into the `create2` strategy.

## Design

The changes would be:

- the `core` deploy function will accept a strategy object in its options that will be typed as either the `Create2Strategy | BasicStrategy`.
- The `Create2Strategy` and `BasicStrategy` will be exported from `core`.
- The `Create2Strategy` class will accept an option object in the constructor `{ salt: string }`.
- A `StrategyConfig` interface will be added that types both the basic strategy config options and the Create2 config options.
- The Ignition helper classes for `viem` and `ethers` will update their deploy functions to accept a strategy option typed as `"create2" | "basic"`. The options object will also accept a `strategyConfig` property that matches the config structure that matches the property picked under the strategy field.
- The Ignition hardhat config object will be enhanced with a strategyConfig property typed as `StrategyConfig`
- The Ignition helpers will default to the Hardhat config strategy config, but can be overriden by the passed in config
- We will add two new reconciliation rules. If a started futures strategy has changed, we fail reconciliation. Against each future we store the strategy’s config. If that config has changed we fail reconciliation.

The `StrategyConfig` interface would look like

```tsx
interface StrategyConfig {
  basic: {},
  create2: {
    salt: string;
  };
}
```

The hardhat config file would look like:

```solidity
// hardhat.config.ts
{
  ...,
  ignition: {
    strategyConfig: {
      create2: {
        salt: "test"
      }
    }
  }
}
```

Invoking core deploy would be:

```solidity
// core api
const deployPromise = deploy(
  myModule,
  {
    defaultSender: "",
    // limit the type here to a union of the two classes
    strategy: new Create2Strategy({ salt: "test" }),
  }
);
```

A programmatic invocation via an Ignition helper:

```tsx
// ignition helper
const deployPromise = deploy(
  myModule,
  {
    defaultSender: "",
    strategy: "create2",
    // this overrides the config for the strategy in hardhat.config.ts
    strategyConfig: {
      salt: "overriding-test"
    }
  }
)
```

Which would allow a Command line invocation like:

```shell
npx hardhat ignition deploy ./ignition/modules/MyModule.ts \
  --parameters ./my-module-parameters.json \
  --strategy create2
```

zoeyTM:
resolves #694


packages/core/src/deploy.ts
```diff
@@ -63,6 +64,8 @@ export async function deploy<
   defaultSender?: string;
   strategy: DeploymentStrategyType;
 }): Promise<DeploymentResult> {
+  const executionStrategy: ExecutionStrategy = strategy as ExecutionStrategy;
```
zoeyTM, line 67: What's the purpose of this type cast?

kanej: The `DeploymentStrategyType` is the reduced and simplified type we are currently exposing to the public. `ExecutionStrategy` is the full internally used type we don't want to expose, so as soon as we are behind the public api we convert from reduced to full for internal use.

packages/hardhat-plugin/src/type-extensions.ts
```diff
@@ -14,10 +14,14 @@ declare module "hardhat/types/config" {
   }
 
   export interface HardhatUserConfig {
-    ignition?: Partial<DeployConfig>;
+    ignition?: Partial<DeployConfig> & {
+      strategyConfig?: Partial<StrategyConfig>;
```
kanej, line 18: My mental model here is that because the `deploy` function of core expects the Strategy object to be newed up, the config will already have been amalgamated into the Strategy before we pass it on. Hence `DeployConfig` will never have StrategyConfig.

packages/core/src/internal/execution/execution-engine.ts
```diff
@@ -74,6 +74,11 @@ export class ExecutionEngine {
       defaultSender
     );
 
+    await this._executionStrategy.init(
```
kanej, line 77: @zoeyTM you initially had this moved up into the `deploy` function; I moved it back down to the execution engine. My reasoning is that we don't want to do an init on a strategy until we are sure that something is definitely going to be executed - so it should have passed all previous validations/checks/hurdles.
Can you recall the initial reason for moving it up to the `deploy` function?

packages/core/src/internal/execution/future-processor/helpers/build-initialize-message-for.ts
```diff
@@ -40,7 +41,8 @@ export async function buildInitializeMessageFor(
         _extendBaseInitWith(
           JournalMessageType.DEPLOYMENT_EXECUTION_STATE_INITIALIZE,
           future,
-          strategy,
+          strategy.name,
+          strategy.config,
```
kanej, line 45: When we expose the full strategy API we will have to ensure that passed arbitrary config is serializable.

packages/core/src/internal/execution/future-processor/helpers/build-initialize-message-for.ts
```diff
@@ -226,18 +234,21 @@ function _extendBaseInitWith<
 >(
   messageType: MessageT,
   future: FutureT,
-  strategy: { name: string },
+  strategy: string,
+  strategyConfig: Record<PropertyKey, string | number>,
```
kanej, line 238: This is an internal type. It covers the current possibilities, and can be expanded to recursive types with the Strategy API.

packages/core/src/internal/execution/types/execution-strategy.ts
```diff
@@ -216,7 +218,15 @@ export interface ExecutionStrategy {
...
-  init: () => Promise<void>;
+  /**
+   * The configuration options for the strategy.
+   */
+  config: Record<PropertyKey, string | number> | Record<PropertyKey, never>;
+
+  init: (
+    deploymentLoader: DeploymentLoader,
+    jsonRpcClient: JsonRpcClient
+  ) => Promise<void>;
```
kanej, line 229: This is internal and covers the current strategies. But basically the init should inject any services that a strategy might use.

packages/core/src/internal/execution/types/execution-strategy.ts
```diff
@@ -216,7 +218,15 @@ export interface ExecutionStrategy {
    */
   name: string;
 
-  init: () => Promise<void>;
+  /**
+   * The configuration options for the strategy.
+   */
+  config: Record<PropertyKey, string | number> | Record<PropertyKey, never>;
```
kanej, line 224: This union lets us cast from the reduced strategy types to the fuller strategy types.

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
```diff
@@ -0,0 +1,70 @@
...
+    return fail(
+      future,
+      `Strategy changed from "${storedStrategyName}" to "${newStrategyName}"`
+    );
+  }
+
+  const storedStrategyConfig = exState.strategyConfig;
+  const newStrategyConfig = context.strategyConfig;
+
+  const isEqual = require("lodash/isEqual") as typeof import("lodash/isEqual");
```
kanej, line 61: @zoeyTM what is the thinking behind the dynamic load here? Is it the weight of loading lodash?

packages/core/src/strategies/basic-strategy.ts
```diff
@@ -6,43 +7,79 @@ import {
...
+ * for each contract deployment, call, and send data, and a single static call
+ * for each static call execution.
+ *
+ * @beta
  */
-export class BasicExecutionStrategy implements ExecutionStrategy {
-  public readonly name: string = "basic";
+export class BasicStrategy {
+  public readonly name = "basic" as const;
+  public readonly config: Record<PropertyKey, never>;
```
kanej, line 39: We initially had `{}`, but that implies a far broader type than the empty object, hence the odd record construct above.

packages/core/src/strategies/create2-strategy.ts
```diff
@@ -82,14 +132,22 @@ export class Create2ExecutionStrategy implements ExecutionStrategy {
...
+    assertIgnitionInvariant(
+      this._deploymentLoader !== undefined && this._jsonRpcClient !== undefined,
+      "Strategy not initialized"
+    );
+
+    const artifact = await this._deploymentLoader.loadArtifact(
+      executionState.artifactId
+    );
+
+    const salt = ethers.id(this.config.salt);
```
kanej, line 150: Is this worth wrapping in a util function to encapsulate the `ethers` usage?

packages/hardhat-plugin/src/helpers.ts
```diff
@@ -1,2 +1,3 @@
 export { HardhatArtifactResolver } from "./hardhat-artifact-resolver";
 export { errorDeploymentResultToExceptionMessage } from "./utils/error-deployment-result-to-exception-message";
+export { resolveStrategy } from "./utils/resolveStrategy";
```
kanej, line 3: I think this additional export to viem and ethers variants makes sense.

packages/core/src/internal/execution/execution-engine.ts
zoeyTM: I don't think there was any important reason. I just needed the deployment loader and it was the first place I was able to pass it in. This change looks good to me.

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
zoeyTM: The weight of loading lodash, yes, but mostly just mimicking the way it was imported in the `reconcile-arguments` function.

packages/core/src/internal/execution/future-processor/helpers/build-initialize-message-for.ts
kanej: This is an addition to the journal, incase that is not immediately obvious.

packages/core/src/types/deploy.ts
```diff
@@ -210,18 +213,20 @@ export interface DeploymentParameters {
...
+ * The different types of deployment strategies.
  *
  * @beta
  */
-export enum DeploymentStrategyType {
-  /**
-   * The basic deployment strategy.
-   */
-  BASIC = "basic",
+export type DeploymentStrategyType = Create2Strategy | BasicStrategy;
```
alcuadrado, line 220: I don't understand why we are exposing this. We were previously initializing the strategy based on an enum. We discussed changing it to a key of `StrategyConfig` instead, as shown [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMpinSBzAngYQHsQZhtkBvAKGVuQCM4BnYBALkoF8AaGuhKBCwQATB2p1JyJnAA2YDkwyhsAbj61O6zlSowAriARhgxZABMIAB1mFcACgCUHAG6Fg59QaMmzlm3YAPOiYOLgAKsgQAB6QIOZMyADWELiEMGgYwnhEJGTIALzIAESMLAjFAHz2SqEQeBwh2RHc0llhuaTYje31BMRdANpNYeEAus7Ibh5ehsamIBbWtrjBvXiRMXEJyanpmXU5A-lFpcysVTXruAD8PYctbQ+dZHcHzS-Yw9fjk9PmlF0Oio-hWTnUoLs9jO5WKrQonEcEOWUOKAiEkBEcMo0jkChKzHMxWQiNUQA)

I don't get why we are now exposing the strategies, this brought about more changes than needed.

packages/core/src/deploy.ts
```diff
@@ -63,8 +62,10 @@ export async function deploy<
   deploymentParameters: DeploymentParameters;
   accounts: string[];
   defaultSender?: string;
-  strategy?: DeploymentStrategyType;
+  strategy: DeploymentStrategyType;
```
alcuadrado, line 65: See https://github.com/NomicFoundation/hardhat-ignition/pull/696/files#r1513488521

packages/core/src/index.ts
```diff
@@ -5,6 +5,8 @@ export * from "./errors";
 export { IgnitionModuleSerializer } from "./ignition-module-serializer";
 export { formatSolidityParameter } from "./internal/formatters";
 export { status } from "./status";
+export { BasicStrategy } from "./strategies/basic-strategy";
```
alcuadrado, line 8: See https://github.com/NomicFoundation/hardhat-ignition/pull/696/files#r1513488521

packages/core/src/internal/execution/types/messages.ts
```diff
@@ -79,6 +82,7 @@ export interface DeploymentExecutionStateInitializeMessage {
   futureId: string;
   futureType: DeploymentExecutionStateFutureTypes;
   strategy: string;
+  strategyConfig: PassedStrategyConfig;
```
alcuadrado, line 85: What happens if we read an existing journal that doesn't have this property?

alcuadrado: I guess the reducers should handle it? Can they?

packages/core/src/internal/execution/types/execution-state.ts
```diff
@@ -42,6 +42,12 @@ export enum ExecutionSateType {
   SEND_DATA_EXECUTION_STATE = "SEND_DATA_EXECUTION_STATE",
 }
 
+/**
+ * The Strategy configuration that the user has provided in the run
+ * the future is started. Used to reconcile the future on subsequent runs.
+ */
+export type PassedStrategyConfig = Record<string, number | string>;
```
alcuadrado, line 49: How is it that we can have this narrower type here and a more generic one everywhere else? Or am I missing something?

packages/core/src/internal/execution/types/execution-strategy.ts
alcuadrado: I'd say this is ok for now

packages/core/src/internal/execution/types/execution-state.ts
alcuadrado: Maybe a better name than "Passed" could be "ConcreteExecutionConfig" or something like that?

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
alcuadrado: I think this makes sense. If we had the slow imports plugins that luist built, it would have forced you to do this.

packages/core/src/internal/execution/types/execution-strategy.ts
alcuadrado: The only part that surprised me is that it now takes a wider type as argument when compared with the previous constructor. It used to use `LoadArtifactFunction` and now it's aware of the entire deployment.

packages/core/src/strategies/create2-strategy.ts
alcuadrado: Why are we hashing this? Doesn't createX do it? is this expected? Can't we validate that it's 32bytes instead? Doesn't this alter the input, making it impossible to control the replay protection with bit flags as createX expects?

I know the answers to some of this questions are "no", because you have some tests using those bit flags, but i don't get it.

packages/hardhat-plugin/src/type-extensions.ts
alcuadrado: I don't understand the newed up part.

packages/core/src/types/deploy.ts
kanej: I had thought it was the design we had coming out of the strategies config document discussion: https://www.notion.so/nomicfoundation/Create2-Configuration-795c08cfa9244f03b4bb9f6e4de5ba07?pvs=4

I may have misunderstood the decision made there.

Why pass a strategy object rather than `"basic" | "create2"` into core's `deploy` function? I was trying to avoid a breaking change on the core api when Strategies proper are rolled out. With a full strategy API I expect that we will pass a strategy object rather than a strategy factory + config that are unified in core.

In my mind the three options are:

1. Pass initialized strategies into `deploy`, but have to expose two core strategies now
2. Pass config and an enumeration now, unifying them within core. Then later support passing a strategy factory in addition as well as the enumeration. Basically always support the current API in addition to the new API of passing a strategy factory along with the strategy config.
3. Pass config and an enumeration now. Breaking change on the API with the full strategy API. 

Is there a better approach? Or is something like 2 what you would prefer to see?

packages/core/src/internal/execution/types/messages.ts
kanej: We can choose a behaviour to deal with the absence of a `strategyConfig`. Should reconciliation fail or pass if the previous run was on the previous version.

I think we can say it returns the `{}` and either pass or fail reconciliation on that, as the previous version is by definition `basic` with no config.

packages/core/src/internal/execution/types/execution-state.ts
kanej: Updated to `ConcreteExecutionConfig`

packages/core/src/internal/reconciliation/futures/reconcileNamedContractDeployment.ts
```diff
@@ -43,6 +44,11 @@ export async function reconcileNamedContractDeployment(
     return result;
   }
 
+  result = reconcileStrategy(future, executionState, context);
```
alcuadrado, line 47: We should only reconcile the strategy if the execution hasn't completed. i.e. if a future has already been successfully executed, we don't care about its strategy.

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
```diff
@@ -67,7 +67,7 @@ export function reconcileStrategy(
     );
   }
 
-  const storedStrategyConfig = exState.strategyConfig;
+  const storedStrategyConfig = exState.strategyConfig ?? {};
```
alcuadrado, line 70: Let's add a comment explaining it here. We may accidentally delete it, because it may look unnecesary.

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
```diff
@@ -0,0 +1,82 @@
...
+  const newStrategyName = context.strategy;
+
+  if (storedStrategyName !== newStrategyName) {
+    return fail(
+      future,
+      `Strategy changed from "${storedStrategyName}" to "${newStrategyName}"`
+    );
+  }
+
+  const storedStrategyConfig = exState.strategyConfig ?? {};
```
alcuadrado, line 70:
```suggestion
  // We may have an `undefined` strategy config when reading a journal, as
  // some previous versions of Ignition didn't set this property
  const storedStrategyConfig = exState.strategyConfig ?? {};
```

packages/core/src/internal/reconciliation/helpers/reconcile-strategy.ts
alcuadrado: see my suggestion

alcuadrado (approved):
I sent a commit adding a comment. 

LGTM

PR #709, opened 2024-03-07
Replaced process.exit() with process.exitCode()

### LINKED ISSUE #708: Don't use `process.exit` in Hardhat tasks
I just noticed that our plugins are using `process.exit` in some tasks to exit with an error code (e.g. `process.exit(1)`). Instead of doing that, we should be using [`process.exitCode`](https://nodejs.org/api/process.html#processexitcode_1)

rishabhdwivedii:
fixes #708


alcuadrado:
Hey, thanks for sending this PR, but unfortunately we got two fixing the same issue and picked the other one instead. The reason is that `process.exitCode` is not a function, so this one doesn't work.

PR #712, opened 2024-03-12
Simplify source code input for verification

### LINKED ISSUE #706: apply HH-verify input simplification
The build info used for the input to Etherscan during verification can include extraneous files for the compilation of the contract.
In Hardhat verify there is a simplification step that is attempted to reduce the nubmer of files that have to be sent to Etherscan.
We should replicate this behaviour in Ignition's verify logic.

## Questions

1. Can we get an outline of the HH-verify logic for simplifying the input to Etherscan, to determine if we can replicate it in general
2. Is it possible to reuse the HH-verify logic directly by applying it to the Etherscan class that is programmatically exposed?
3. How should Ignition, with its pre-saved artifacts and build infos recreate this behaviour?

zoeyTM:
fixes #706


packages/core/src/verify.ts
```diff
@@ -87,6 +90,23 @@ function resolveChainConfig(
...
 
+function getImportSourceNames(
+  sourceName: string,
+  buildInfo: BuildInfo
+): string[] {
+  const contractSource = buildInfo.input.sources[sourceName].content;
+  const { imports } = analyze(contractSource);
+
+  const importSources = imports.map((i) =>
+    path.join(path.dirname(sourceName), i).replaceAll("\\", "/")
```
kanej, line 101: @alcuadrado because this is our build info are we able to assume that there are no remappings that need to be considered?

alcuadrado: Does it work with global imports like `import "@openzepellin/Owner.sol";`?

alcuadrado: > @alcuadrado because this is our build info are we able to assume that there are no remappings that need to be considered?

The remappigns are already included in the buildInfo, and they are correct, because the build already succeeded.

alcuadrado: What we need to do with remappings though, is to remove the unused ones, as those may create a privacy issue.

They way to do that would be something like this:

* For each import in a file, if it's global, check if it matches any of the input remappings. If it does, flag that remapping as used.
* After processing all the files, only include the used remappings.

alcuadrado: This implements the logic of choosing which remapping affects an import, if any: https://github.com/NomicFoundation/hardhat/blob/a3ada90ad7362660c98372c36779409121af39ac/packages/hardhat-core/src/utils/remappings.ts#L5-L14

alcuadrado: Note that in that logic, the length of the "match" is used to pick the right remapping if more than one matches.

packages/core/src/verify.ts
```diff
@@ -121,20 +154,20 @@ function prepareInputBasedOn(
...
 
   if (sourceToLibraryAddresses === null) {
-    return JSON.stringify(buildInfo.input);
+    return buildInfo.input;
   }
 
   const { input } = buildInfo;
   input.settings.libraries = sourceToLibraryAddresses;
 
-  return JSON.stringify(input);
```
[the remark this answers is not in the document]
alcuadrado: If it fails we should throw an error explaining that they can try with a flag to include the unrelated contracts

alcuadrado:
Something that we discussed on a call.

Let's add a flag `includeUnderlatedContracts` to the verification function that core exposes. By default it should be false. If true, we don't minify the solc input.

Then, in the plugin, we add a `--include-unrelated-contracts` flag to verify, which enables that flag.

If the verification fails and the flag wasn't on, then we recommend the users to retry with `--include-unrelated-contracts`.

alcuadrado (approved):
LGTM

PR #718, opened 2024-03-25
fee issue reproduction

kanej:
A reproduction for the fee issue with `create2`.

image

## Reproduction steps

```shell
git clean -fdx .
npm install
npm run build
cd ./examples/viem-sample

npx hardhat vars set ALCHEMY_API_KEY XXXX
npx hardhat vars set TEST_PRIVATE_KEY XXXX

npx hardhat ignition deploy ./ignition/modules/LockModule.ts --strategy create2 --network sepolia
```

Note that on sepolia, `./ignition/modules/Rocket.ts` is deployable with `create2` - perhaps this is related to `value` passing in the LockModule when converted through create2 strategy.


alcuadrado:
I just pushed a commit fixing this issue. It doesn't have tests. 

There was actually a TODO explaining that this issue could happen (my fault probably).

Basically, if the gas estimation fails, we simulate the transaction so that we can get access to the return data. During that simulation we weren't setting the gas limit, so some clients would use the block gas limit, and most users didn't have enough eth to afford that and the simulation failed.

I fixed it by computing the max possible gas limit before running that simulation, and using that value instead of `undefined`.

We may be able to always use that one value, but it's not clear. I'm asking someone from the Geth team about it.

alcuadrado:
An alternative fix, as discussed with the Geth team member, is not setting fee information to calls, unless when we know we are setting also the gas limit. We could enforce this in the type system with something like

```ts
/**
 * The params to make an `eth_call`.
 */
export type CallParams = {
  to?: string;
  value: bigint;
  data: string;
  from: string;
  nonce?: number;
} & ({} | { fees: NetworkFees; gasLimit: bigint });
```

alcuadrado:
This had already been fixed by @zoeyTM in https://github.com/NomicFoundation/hardhat-ignition/pull/684

PR #730, opened 2024-04-04
Support zero gas fee chains

jimthematrix:
This is to add support for chains that have been set up with zero base fee, such as a private instance of blockchains using Besu.

### DIFF
--- a/packages/core/src/internal/execution/jsonrpc-client.ts
+++ b/packages/core/src/internal/execution/jsonrpc-client.ts
@@ -632,7 +632,11 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
     // We prioritize EIP-1559 fees over legacy gasPrice fees
     if (latestBlock.baseFeePerGas !== undefined) {
       // Logic copied from ethers v6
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/execution/jsonrpc-client.ts
```diff
@@ -632,7 +632,11 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
     // We prioritize EIP-1559 fees over legacy gasPrice fees
     if (latestBlock.baseFeePerGas !== undefined) {
       // Logic copied from ethers v6
-      const maxPriorityFeePerGas = 1_000_000_000n; // 1gwei
+      let maxPriorityFeePerGas = 1_000_000_000n; // 1gwei
+      if (latestBlock.baseFeePerGas === 0n) {
+        // support zero gas fee chains
```
kanej, line 637: Can you expand on this comment to give an example, i.e. private chains like besu

kanej (commented):
Thanks for this @jimthematrix, I left a small comment. I think once you rebase and update the comment this is good to go.

kanej (approved):
LGTM.

PR #731, opened 2024-04-09
Integrating hardhat-ledger support

### LINKED ISSUE #720: Provide clean UI when using ledger
With the enhancements to `hardhat-ledger` coming from https://github.com/NomicFoundation/hardhat/pull/4946, we can now control the cli UI entirely from Ignition.

## TODO

- [ ] Add `hardhat_setLedgerOutputEnabled` provider rpc call in Ignition deploy to disable ledger cli UI changes
- [ ] Add listeners to the provider that update the UI state
- [ ] Display a waiting for ledger confirmation message at the bottom of the current batch, and update it with successful or failed confirmation
- [ ] Integration test if possible (maybe fake the provider?)

## Potential UI

```shell
Batch #2
  Executed RESOLVER#PublicResolver
  Executed RESOLVER#set_subnode_owner_for_resolver
  Executed REVERSEREGISTRAR#set_subnode_owner_reverse

  Ledger: Connecting wallet
```

```shell
Batch #2
  Executed RESOLVER#PublicResolver
  Executed RESOLVER#set_subnode_owner_for_resolver
  Executed REVERSEREGISTRAR#set_subnode_owner_reverse

  Ledger: Waiting for confirmation
```

```shell
Batch #2
  Executed RESOLVER#PublicResolver
  Executed RESOLVER#set_subnode_owner_for_resolver
  Executed REVERSEREGISTRAR#set_subnode_owner_reverse

  Ledger: Transaction signed
```

zoeyTM:
fixes #720 

Waiting for confirmation from ledger device
Screen Shot 2024-04-09 at 12 18 17 AM

Ledger device confirmed and transaction was sent to network (waiting for block confirmations before continuing)
Screen Shot 2024-04-09 at 12 18 30 AM

Previous ledger messages are removed when execution continues (and are replaced by current ledger messages when needed)
Screen Shot 2024-04-09 at 12 19 21 AM

All ledger messages removed for final status display
Screen Shot 2024-04-09 at 12 20 36 AM

### DIFF
--- a/packages/hardhat-plugin/src/index.ts
+++ b/packages/hardhat-plugin/src/index.ts
@@ -262,39 +262,49 @@ ignitionScope
             false,
           ]);
 
-          hre.network.provider.once("connection_start", () =>
-            executionEventListener.ledgerConnectionStart()
+          hre.network.provider.once(
+            "connection_start",
+            executionEventListener.ledgerConnectionStart
           );
-          hre.network.provider.once("connection_success", () =>
-            executionEventListener.ledgerConnectionSuccess()
+          hre.network.provider.once(
+            "connection_success",
+            executionEventListener.ledgerConnectionSuccess
           );
-          hre.network.provider.once("connection_failure", () =>
-            executionEventListener.ledgerConnectionFailure()
+          hre.network.provider.once(
+            "connection_failure",
+            executionEventListener.ledgerConnectionFailure
           );
-          hre.network.provider.on("confirmation_start", () =>
-            executionEventListener.ledgerConfirmationStart()
+          hre.network.provider.on(
+            "confirmation_start",
+            executionEventListener.ledgerConfirmationStart
           );
-          hre.network.provider.on("confirmation_success", () =>
-            executionEventListener.ledgerConfirmationSuccess()
+          hre.network.provider.on(
+            "confirmation_success",
+            executionEventListener.ledgerConfirmationSuccess
           );
-          hre.network.provider.on("confirmation_failure", () =>
-            executionEventListener.ledgerConfirmationFailure()
+          hre.network.provider.on(
+            "confirmation_failure",
+            executionEventListener.ledgerConfirmationFailure
           );
 
           // these events aren't currently used by our UI,
           // but i'm leaving them here for future reference
 
-          // hre.network.provider.once("derivation_start", () =>
-          //   executionEventListener.ledgerDerivationStart()
+          // hre.network.provider.once(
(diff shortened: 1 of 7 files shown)


packages/hardhat-plugin/src/ui/pretty-event-handler.ts
```diff
@@ -309,6 +312,77 @@ export class PrettyEventHandler implements ExecutionEventListener {
...
+      ...this.state,
+      ledger: true,
+      ledgerMessage: "Connecting wallet",
+    };
+
+    this._redisplayCurrentBatch();
+
+    this.state = {
+      ...this.state,
+      ledgerMessageIsDisplayed: true,
```
kanej, line 326: Can you walk me through the need for the split here and have two state updates?
Why can't `ledgerMessageIsDisplayed: true` go in the first state update?

packages/hardhat-plugin/test/ui/helpers/calculate-deploying-module-panel-display.ts
```diff
@@ -21,6 +21,9 @@ describe("ui - calculate starting message display", () => {
     maxFeeBumps: 0,
     gasBumps: {},
     strategy: "basic",
+    ledger: false,
+    ledgerMessage: "",
+    ledgerMessageIsDisplayed: false,
```
kanej, line 26: Is the `ledger` property needed in addition to `ledgerMessageIsDisplayed`?
What drives the difference between the `ledger: true` state and the `ledgerMessageIsDisplayed: false` state?

packages/hardhat-plugin/src/index.ts
```diff
@@ -273,6 +314,31 @@ ignitionScope
...
         });
 
+        try {
+          await hre.network.provider.send("hardhat_setLedgerOutputEnabled", [
+            true,
+          ]);
+
+          hre.network.provider.off("connection_start", () =>
+            executionEventListener.ledgerConnectionStart()
+          );
```
kanej, line 324: The `.off` here won't work as anonymous function is not the same instance as was passed into `.on`.

I suspect that the anonymous functions should be replaced with named local functions, that way both `.on` and `.off` can be passed the same instances.

zoeyTM: would this work?

```ts
hre.network.provider.off("connection_start", executionEventListener.ledgerConnectionStart);
```

kanej: Yes

PR #739, opened 2024-04-23
Swap monorepo from npm to pnpm

### LINKED ISSUE #701: Switch to pnpm from npm
Hardhat uses `pnpm`. We should match HH in preparation for moving to the HH monorepo.

zoeyTM:
Resolves #701 

Note: Due to the fact that we will not be bringing the `examples` over when we migrate to the hardhat monorepo, I elected to leave them out of the `pnpm` upgrade. That includes not linking them to the workspace as well as removing their tests from our CI and `fullcheck` npm script.


packages/hardhat-plugin/README.md
```diff
@@ -15,10 +15,10 @@ Join the Hardhat Ignition channel of our [Hardhat Community Discord server](http
 
 ```bash
 # ethers users
-npm install --save-dev @nomicfoundation/hardhat-ignition-ethers
+pnpm install --save-dev @nomicfoundation/hardhat-ignition-ethers
```
kanej, line 18: We should leave the install instructions in the readmes as assuming `npm`

package.json
```diff
@@ -3,28 +3,18 @@
...
-      "packages/hardhat-plugin",
-      "packages/hardhat-plugin-ethers",
-      "packages/hardhat-plugin-viem",
-      "examples/*"
-    ]
-  },
   "scripts": {
-    "build": "npm run build -w packages",
-    "lint": "npm run lint --workspaces",
+    "build": "pnpm --filter \"./packages/*\" build",
```
kanej, line 7: In Harhdat the equivalent command is `pnpm run --recursive build` combined with a `pnpm-workspace.yaml` file.
Is there a tradeoff here? Is the goal to constrain down the build to only the packages subfolders?

package.json
```diff
@@ -3,28 +3,18 @@
...
-    "test": "npm run test -w packages && npm run test:integrations -w packages/core",
-    "test:coverage": "npm run test:coverage -w packages",
-    "test:examples": "npm run test -w examples --if-present",
+    "test": "pnpm --filter \"./packages/*\" test && pnpm --filter \"./packages/core\" test:integrations",
+    "test:coverage": "pnpm --filter \"./packages/*\" test:coverage",
+    "test:examples": "pnpm --filter \"./examples/*\" test",
     "watch": "tsc --build --watch packages/core packages/hardhat-plugin",
-    "fullcheck": "npm run build && npm run lint && npm run test:examples",
-    "clean": "npm run clean --workspaces --if-present"
+    "fullcheck": "pnpm build && pnpm lint",
```
kanej, line 16:
```suggestion
    "fullcheck": "pnpm build && pnpm lint && pnpm test",
```
Even if we are leaving out the example tests we should run the standard integeration tests.

packages/core/package.json
```diff
@@ -31,19 +31,19 @@
...
+    "preapi-extractor": "pnpm build",
     "api-extractor": "api-extractor run --local --verbose",
     "test": "mocha --recursive \"test/**/*.ts\"",
-    "test:debug": "DEBUG='ignition:*' npm run test",
+    "test:debug": "DEBUG='ignition:*' pnpm test",
     "test:build": "tsc --project ./test/",
     "test:coverage": "nyc mocha --recursive \"test/**/*.ts\" \"test-integrations/**/*.ts\"",
     "test:integrations": "mocha --recursive \"test-integrations/**/*.ts\"",
     "clean": "rimraf .nyc_output coverage dist tsconfig.tsbuildinfo",
-    "prepack": "npm run build"
```
kanej, line 46: I think to run pre/post scripts in pnpm you need to enable: `enable-pre-post-scripts=true` in the `.npmrc` file:

https://pnpm.io/cli/run#enable-pre-post-scripts

.github/workflows/ci.yml
```diff
@@ -13,36 +13,38 @@ jobs:
...
       - name: Install
-        run: npm install
+        run: pnpm install
       - name: Build
-        run: npm run build
+        run: pnpm build
       - name: Run tests
-        run: npm run test
-      - name: Run tests in examples
-        run: npm run test:examples
```
kanej, line 27: Is the intent that we won't run example tests at all now?

Is there anyone to keep the tests for CI time?

package.json
zoeyTM: Ah, the first time I tried that version was before I added the `pnpm-workspace.yaml` file. I already had this version in place by the time I added that file, so I didn't connect the dots that that was the specific reason it wasn't working before. I don't think there's any tradeoffs, but I'll change it to match Hardhat's style for consistency

packages/core/package.json
zoeyTM: `enable-pre-post-scripts` is enabled by default as of version 9

.github/workflows/ci.yml
zoeyTM: My understanding is that we are not bringing the examples over to the Hardhat repo, so, since there were non-trivial `node_modules` resolution problems specific to the examples blocking my previous attempt at this PR, I decided it wasn't worth dedicating dev time to solving those issues since we won't have the examples available in that way soon anyways. 

There may be other ways we can bring them over to the Hardhat repo in a different form, but I would consider that to be a separate issue.

kanej: I am going to miss them.

I pushed a commit that updates the publishing instructions to say:

1. install with verdaccio
2. install in each of the examples
3. run test

It doesn't solve the developer experience of hacking on changes in the examples, but at least we still check that the examples work before a release.

kanej (approved):
LGTM.

PR #740, opened 2024-04-24
Custom logic for Polygon and support for `maxPriorityFeePerGas`

### LINKED ISSUE #728: Add maxPriorityFeePerGas param to config
We utilize your tool for deployments, but encounter an issue when attempting to deploy our contracts to Polygon - there's a maxPriorityFeePerGas limit of 30 GWei on the mainnet. Below is a snippet of your code for configuring transaction fees (please refer to the EIP1193JsonRpcClient in the ignition-core):
```
// We prioritize EIP-1559 fees over legacy gasPrice fees
        if (latestBlock.baseFeePerGas !== undefined) {
            // Logic copied from ethers v6
            const maxPriorityFeePerGas = 1_000_000_000n; // 1gwei
            const maxFeePerGas = latestBlock.baseFeePerGas * 2n + maxPriorityFeePerGas;
            return {
                maxFeePerGas,
                maxPriorityFeePerGas,
            };
        }
```
Due to this limitation, we're unable to configure maxPriorityFeePerGas, resulting in all our transactions being stuck in the mempool and never executed. Is it possible to either add maxPriorityFeePerGas to the config file or calculate it using the fee from the previous block?"

zoeyTM:
- adds `maxPriorityFeePerGas` to user config
- returns legacy fees when deploying to Polygon
  - manually tested and confirmed that this does indeed fix deployments on Polygon

resolves #728

### DIFF
--- a/packages/core/src/deploy.ts
+++ b/packages/core/src/deploy.ts
@@ -52,6 +52,7 @@ export async function deploy<
   strategy,
   strategyConfig,
   maxFeePerGasLimit,
+  maxPriorityFeePerGas,
 }: {
   config?: Partial<DeployConfig>;
   artifactResolver: ArtifactResolver;
@@ -69,6 +70,7 @@ export async function deploy<
   strategy?: StrategyT;
   strategyConfig?: StrategyConfig[StrategyT];
   maxFeePerGasLimit?: bigint;
+  maxPriorityFeePerGas?: bigint;
 }): Promise<DeploymentResult> {
   const executionStrategy: ExecutionStrategy = resolveStrategy(
     strategy,
@@ -114,6 +116,7 @@ export async function deploy<
 
   const jsonRpcClient = new EIP1193JsonRpcClient(provider, {
     maxFeePerGasLimit,
+    maxPriorityFeePerGas,
   });
 
   const isAutominedNetwork = await checkAutominedNetwork(provider);
--- a/packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
+++ b/packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
@@ -145,6 +145,12 @@ export class EthersIgnitionHelper {
       defaultSender,
       strategy,
       strategyConfig: resolvedStrategyConfig,
+      maxFeePerGasLimit:
+        this._hre.config.networks[this._hre.network.name]?.ignition
+          .maxFeePerGasLimit,
+      maxPriorityFeePerGas:
+        this._hre.config.networks[this._hre.network.name]?.ignition
+          .maxPriorityFeePerGas,
     });
 
     if (result.type !== DeploymentResultType.SUCCESSFUL_DEPLOYMENT) {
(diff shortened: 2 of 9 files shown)


packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
```diff
@@ -145,6 +145,12 @@ export class EthersIgnitionHelper {
       defaultSender,
       strategy,
       strategyConfig: resolvedStrategyConfig,
+      maxFeePerGasLimit:
+        this._hre.config.networks[this._hre.network.name]?.ignition
+          .maxFeePerGasLimit,
```
kanej, line 150:
```suggestion
          ?.maxFeePerGasLimit,
```
Is this needed (and on the next one) for the case where ignition is defined but the `maxFeePerGasLimit` isn't?

packages/core/src/deploy.ts
```diff
@@ -69,6 +70,7 @@ export async function deploy<
   strategy?: StrategyT;
   strategyConfig?: StrategyConfig[StrategyT];
   maxFeePerGasLimit?: bigint;
+  maxPriorityFeePerGas?: bigint;
```
kanej, line 73: Should this not be part of the `config: DeployConfig`. What is our rule for moving it up to the top level?

packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
```diff
@@ -145,6 +145,12 @@ export class EthersIgnitionHelper {
...
+      maxPriorityFeePerGas:
+        this._hre.config.networks[this._hre.network.name]?.ignition
```
kanej, line 152: We should consider whether the deploy on the ethers/viem helpers should expose an override for both `maxFeePerGasLimit` and `maxPrioirityFeePerGas`.
  
Is it just unlikely to come up in test scenarios, given they are against Hardhat node?

packages/hardhat-plugin/src/index.ts
```diff
@@ -271,6 +272,9 @@ ignitionScope
           strategyConfig,
           maxFeePerGasLimit:
             hre.config.networks[hre.network.name]?.ignition.maxFeePerGasLimit,
+          maxPriorityFeePerGas:
```
kanej, line 275: Are there integration tests that are need here around config?

packages/core/src/deploy.ts
zoeyTM: `DeployConfig` is meant for configuring ignition's behavior. `maxFeePerGasLimit` and `maxPriorityFeePerGas` are not part of `DeployConfig` because they can be configured differently on a per network basis.

packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
zoeyTM: I think it was simply an oversight that they were not already exposing these overrides. I think both helpers should expose any options available for normal deployment, specifically because of script based deployments.

zoeyTM: or are you suggesting adding additional overrides that can be passed in via the `deploy` function itself? I'm less inclined to provide this pattern unless it's requested by users.

kanej: Lets leave this for now. We can add it later if we get requests.

kanej (approved):
LGTM!

PR #761, opened 2024-05-16
Add function to encode a function call

### LINKED ISSUE #760: Add function to encode function call args
```typescript
m.encodeFunctionCall<
  ContractNameT extends string,
  FunctionNameT extends string
>(
  contract: CallableContractFuture<ContractNameT>,
  functionName: FunctionNameT,
  args: ArgumentType[]
): EncodeFunctionCallFuture

interface EncodeFunctionCallFuture<
  ContractNameT extends string,
  FunctionNameT extends string
> {
  type: FutureType.ENCODE_FUNCTION_CALL;
  id: string;
  module: IgnitionModule;
  dependencies: Set<Future>;
  contract: ContractFuture<ContractNameT>;
  functionName: FunctionNameT;
  args: ArgumentType[];
}
```

`args` should accept the same types as `m.call`

`EncodeFunctionCallFuture` should be able to be passed as an arg to:
- contract constructors

- `m.call` args

- `m.staticCall` args

- `m.send` data 

Validations and reconciliations should be same as for `m.call` , minus things they don’t have in common like `value` and `from`

Usage:

example using proxy initialization function
```typescript
const proxyAdminOwner = m.getAccount(0);

const contract = m.contract("Contract");

const data = m.encodeFunctionCall(contract, "initialize", [proxyAdminOwner])

const proxy = m.contract("TransparentUpgradeableProxy", [..., data]);

// etc.
```

zoeyTM:
resolves #760


packages/core/src/types/module-builder.ts
```diff
@@ -368,6 +386,31 @@ export interface IgnitionModuleBuilder {
     options?: StaticCallOptions
   ): StaticCallFuture<ContractNameT, FunctionNameT>;
 
+  /**
+   * Encode a function call.
```
kanej, line 390:
```suggestion
   * ABI encode a function call, including both the function's name and the parameters it is being called with. This is useful when sending a raw transaction to invoke a smart contract or when invoking a smart contract proxied through an intermediary function.
```

I am equivocal here. `Encode a function call` is a bit laconic. But I am not convinced I have improved on it. @zoeyTM what do you think?

packages/core/src/types/module-builder.ts
```diff
@@ -368,6 +386,31 @@ export interface IgnitionModuleBuilder {
...
+   *
+   * @param contractFuture - The contract ABI to encode with
```
kanej, line 392:
```suggestion
   * @param contractFuture - The contract that the ABI for encoding will be taken from
```

packages/core/src/types/module-builder.ts
```diff
@@ -368,6 +386,31 @@ export interface IgnitionModuleBuilder {
...
+   * @param functionName - The name of the function
+   * @param args - The arguments to pass to the function
+   * @param options - The options for the call
+   *
+   * @example
+   * ```
+   * const myContract = m.contract("MyContract");
+   * const data = m.encodeFunctionCall(myContract, "updateCounter", [100]);
+   * m.send("callFunctionOnContract", myContract, 0n, data);
```
kanej, line 401:
```suggestion
   * m.send("callUpdateCounter", myContract, 0n, data);
```

examples/complete/test/Complete.js
```diff
@@ -41,4 +41,10 @@ describe("Complete", function () {
 
     expect(await withLib.readonlyFunction(40)).to.equal(42);
   });
+
+  it("Should call the otherFunction with 42", async () => {
+    const { duplicate } = await loadFixture(deployCompleteFixture);
+
+    expect(await duplicate.savedArg()).to.equal(42);
```
kanej, line 48: Yes, but what is the question?

packages/core/src/internal/module-builder.ts
```diff
@@ -610,6 +614,68 @@ class IgnitionModuleBuilderImplementation<
...
+    }
+
+    if (typeof options !== "object") {
+      this._throwErrorWithStackTrace(
+        `Invalid parameter "options" provided to encodeFunctionCall "${functionName}" in module "${this._module.id}"`,
+        this.encodeFunctionCall
+      );
+    }
+
+    const futureId = toCallFutureId(
```
kanej, line 640: I am not sure about this. It is not obvious that this is not a call but the encoding of a call:

image

Maybe something like `Apollo#Rocket.encode[launch]`. I mean even writing it I hate it.

packages/core/test/send.ts
```diff
@@ -236,6 +236,33 @@ describe("send", () => {
...
+    const sendFuture = [...moduleWithDependentContracts.futures].find(
+      ({ id }) => id === "Module1#test_send"
+    );
+
+    if (!(sendFuture instanceof SendDataFutureImplementation)) {
+      assert.fail("Not a send data future");
+    }
+
+    assert.equal(sendFuture.dependencies.size, 2);
+    assert(sendFuture.dependencies.has(exampleFuture!));
```
kanej, line 263: Can we add the check for encode function call future.

packages/core/src/internal/validation/futures/validateNamedEncodeFunctionCall.ts
```diff
@@ -0,0 +1,66 @@
...
+  isModuleParameterRuntimeValue,
+} from "../../../type-guards";
+import { ArtifactResolver } from "../../../types/artifact";
+import { DeploymentParameters } from "../../../types/deploy";
+import { EncodeFunctionCallFuture } from "../../../types/module";
+import { ERRORS } from "../../errors-list";
+import { validateArtifactFunction } from "../../execution/abi";
+import { retrieveNestedRuntimeValues } from "../utils";
+
+export async function validateNamedEncodeFunctionCall(
```
kanej, line 13: I was going to query why this was not just `validateEncodeFunctionCall`, but it is following the pattern of `validateNamedStaticCall`. That raises the further question of whether we intended this pattern for the calls or whether it is just an accident of transcription.
I don't think we solve this here by the way, but do we need another PR for these?

packages/core/test/reconciliation/futures/reconcileNamedEncodeFunctionCall.ts
```diff
@@ -0,0 +1,376 @@
...
+
+    assert.deepStrictEqual(reconiliationResult.reconciliationFailures, [
+      {
+        futureId: "Module#Contract1.function1",
+        failure: "Argument at index 0 has been changed",
+      },
+    ]);
+  });
+
+  it("should reconcile an address arg with entirely different casing", async () => {
```
kanej, line 214: Lets drop this test. I suspect it is pulled over, but I don't think it adds here. Same for the next test.

packages/core/test/reconciliation/futures/reconcileNamedEncodeFunctionCall.ts
```diff
@@ -0,0 +1,376 @@
...
+          futureType: FutureType.ENCODE_FUNCTION_CALL,
+          status: ExecutionStatus.SUCCESS,
+          functionName: "function1",
+          args: [1, "a", differentAddress],
+        }
+      )
+    );
+  });
+
+  it("should find changes to contract unreconciliable", async () => {
```
kanej, line 92:
```suggestion
  it("should find changes to future depenedencies unreconciliable", async () => {
```

packages/core/test/encodeFunctionCall.ts
```diff
@@ -0,0 +1,754 @@
...
+
+              m.encodeFunctionCall(call as any, "test");
+
+              return { another };
+            }),
+          /Invalid contract given/
+        );
+      });
+
+      it("should not validate a library", () => {
```
kanej, line 358: Nice

packages/core/test/encodeFunctionCall.ts
```diff
@@ -0,0 +1,754 @@
...
+              m.encodeFunctionCall(another as any, "test");
+
+              return { another };
+            }),
+          /Invalid contract given/
+        );
+      });
+    });
+
+    it("should not validate a non-existant hardhat contract", async () => {
```
kanej, line 373: Is this not testing the `m.contract` validation?

packages/core/test/encodeFunctionCall.ts
```diff
@@ -0,0 +1,754 @@
...
+      );
+    });
+
+    it("should not validate a missing module parameter (deeply nested)", async () => {
+      const module = buildModule("Module1", (m) => {
+        const p = m.getParameter("p");
+
+        const another = m.contract("Another", fakeArtifact, []);
+        m.encodeFunctionCall(another, "test", [
+          [123, { really: { deeply: { nested: [p] } } }],
```
kanej, line 601: My God we are thorough.

packages/core/src/internal/execution/future-processor/helpers/future-resolvers.ts
```diff
@@ -117,6 +116,25 @@ export function resolveFutureFrom(
...
+  if (data === undefined) {
+    return "0x";
+  }
+
+  if (typeof data === "string") {
+    return data;
+  }
+
+  // this type coercion is safe because we know the type of data is EncodeFunctionCallFuture
+  return findResultForFutureById(deploymentState, data.id) as string;
```
kanej, line 135: Lets initially return a `result` and then do a `assertIgnitionInvariant` to ensure it is a string.

packages/core/src/internal/execution/types/execution-state.ts
```diff
@@ -131,6 +133,25 @@ export interface StaticCallExecutionState
...
+ */
+export interface EncodeFunctionCallExecutionState
+  extends BaseExecutionState<
+    ExecutionSateType.ENCODE_FUNCTION_CALL_EXECUTION_STATE,
+    FutureType.ENCODE_FUNCTION_CALL
+  > {
+  artifactId: string;
+  functionName: string;
+  args: SolidityParameterType[];
+  result?: string;
```
kanej, line 152: Should this just be `result`, like readEvent?

packages/core/src/internal/reconciliation/helpers/reconcile-data.ts
```diff
@@ -10,7 +11,16 @@ import { compare } from "./compare";
...
-  return compare(future, "Data", exState.data, future.data ?? "0x");
+  if (typeof future.data === "string" || future.data === undefined) {
+    return compare(future, "Data", exState.data, future.data ?? "0x");
+  }
+
+  return compare(
+    future,
+    "Data",
+    exState.data,
+    findResultForFutureById(context.deploymentState, future.data.id) as string
```
kanej, line 24: Can we do the `assertIgnitionInvariant` thing here to avoid the cast.

kanej (changes requested):
I left a few comments and nit picks. The only real question is around the id and default id.
Otherwise this is really solid work. Nicely done.

packages/core/src/internal/module-builder.ts
alcuadrado: I think not clashing it would be better. What about something like `Apollo#encodeFunctionCall(<functionCallId>)`?

kanej: Should the underlying contract name be included as well? Is `functionCallId`:
1. `launch`
2. `Rocket.launch` or using some other separator?

packages/core/test/encodeFunctionCall.ts
zoeyTM: No, this validation (as well as the normal function call) has its own version of the check that this is testing. 

That being said, this comment made me notice that I somehow forgot to swap the validation functions when I copied these tests from the normal function call tests, so I'll push a commit with that change.

kanej (approved):
Looks good:

image

PR #755, opened 2024-05-09
Exclude BNB Chain from zero fee configuration in gas fee logic

magicsih:
This commit updates the gas fee configuration logic to accommodate the BNB Chain (chainId 56). While the BNB Network supports EIP-1559, it reports a `baseFeePerGas` of zero(https://github.com/bnb-chain/bsc/issues/2062). This adjustment ensures that transactions on the BNB Chain do not mistakenly receive a max fee configuration of zero, which is not appropriate despite the reported base fee. This change maintains the functionality for other chains such as Polygon (chainId 137) that require legacy gasPrice fees, and excludes only BNB Chain from the zero max fee logic. This is crucial for proper transaction fee handling on networks supporting EIP-1559 but reporting a base fee of zero.

Deployments on the BNB Chain were failing. The required tip for transactions should be n wei, but because the `baseFeePerGas` was mistakenly set to 0, the actual tip was also calculated as 0. This resulted in deployment errors that prevented successful transaction submissions. This commit corrects this misconfiguration, ensuring that the correct transaction fees are applied, thus enabling proper deployments on the BNB Chain.

PR #768, opened 2024-05-25
Exclude BNB Test Chain from zero fee configuration in gas fee logic

### LINKED ISSUE #763: Exclude BNB Test Chain from zero fee configuration in gas fee logic
Following on from: https://github.com/NomicFoundation/hardhat-ignition/pull/755

We should also exclude BNB Testnet Chain with chainId 97 and add a test to cover this case:

1. Enhance this guard to include `97`

https://github.com/NomicFoundation/hardhat-ignition/blob/355112935e732269abf05c8e775a1a146f634c94/packages/core/src/internal/execution/jsonrpc-client.ts#L643

2. Add an equivalent test to this:

https://github.com/NomicFoundation/hardhat-ignition/blob/355112935e732269abf05c8e775a1a146f634c94/packages/core/test-integrations/new-api/internal/new-execution/jsonrpc-client.ts#L188

MukulKolpe:
Following on from #755, Updated the gas fee configuration logic to accommodate the BNB Test Chain (chainId 97) to exclude from zero fee configuration.

Resolves #763

### DIFF
--- a/packages/core/src/internal/execution/jsonrpc-client.ts
+++ b/packages/core/src/internal/execution/jsonrpc-client.ts
@@ -645,7 +645,7 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
       // of blockchains using Besu. We explicitly exclude BNB
       // Smartchain (chainId 56) from this logic as it is EIP-1559
       // compliant but only sets a maxPriorityFeePerGas.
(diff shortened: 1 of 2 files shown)


kanej (approved):
Looks good. Thanks @MukulKolpe!

PR #773, opened 2024-06-03
Fix: handle anvil `hardhat_setBalance` response for create2

### LINKED ISSUE #764: `hardhat_setBalance` blocking create2 deploys on anvil
This was first raised in the discord channel: https://discord.com/channels/750408878008827925/1153426756901032037/1243258480874360883

If you use `create2` as a strategy when deploying against anvil, then because anvil has a chain id of 31337, the same as Hardhat, we attempt to deploy a `createX` factory if it is not present. This requires a `hardhat_setBalance` call to be made to the account that is used for deploying the `createX` factory.

In Hardhat, the rpc call `hardhat_setBalance` returns a boolean. It appears not to in anvil. Ignition on not getting the expected return result throws an error:

image

## Approach

We should loosen the `hardhat_setBalance` return type check to allow anvil to pass. In Hardhat we always return true, or throw if we where unable to set the balance; so the check on the Ignition side is redundant.

## Testing

This is a small fix, but we should double check `create2` runs against anvil manually.

## Workaround

In the meantime, [manually deploying the `createX`](https://github.com/pcaversaccio/createx?tab=readme-ov-file#new-deployments) factory then running an Ignition deploy with `create2` strategy should work.

zoeyTM:
resolves #764 

manually verified working on anvil

### DIFF
--- a/packages/core/src/internal/execution/jsonrpc-client.ts
+++ b/packages/core/src/internal/execution/jsonrpc-client.ts
@@ -286,6 +286,11 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
       params: [address, balanceHex],
     });
 
+    // anvil supports this method, but returns `null` instead of a boolean
+    if (returnedBalance === null) {
+      return true;
+    }
+
     assertResponseType(
       "hardhat_setBalance",
       returnedBalance,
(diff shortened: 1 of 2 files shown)


kanej (approved):
LGTM.

PR #774, opened 2024-06-04
Adjust regex to handle square brackets

### LINKED ISSUE #762: Throwing error when calling overloaded functions in Hardhat-Ignition module.
Here's the minimal example:

**contracts/OverloadSum.sol**

```js
contract OverloadSum {
    function sum(uint256 a, uint256 b) external pure returns (uint256) {
        return a + b;
    }

    function sum(uint256[] memory _array) external pure returns (uint256) {
        return _array[0] + _array[1];
    }
}
```

**ignition/modules/OverloadSum.js**

```js
module.exports = buildModule("OverloadSum", (m) => {

  const overloadSum = m.contract("OverloadSum");
  
   m.call(overloadSum, "sum(uint256,uint256)", [1, 2], { id: "MyUniqueId1" })
  m.call(overloadSum, "sum(uint256[])", [[1, 2]], { id: "MyUniqueId2" })

  return { overloadSum };
});
```

When I'm executing this command:

```js
npx hardhat ignition deploy ignition/modules/OverloadSum.js
```

Getting this error:

```js
IgnitionError: IGN702: Module validation failed with reason: The function name "sum(uint256[])" is invalid, make sure you use a valid identifier.
```

zoeyTM:
fixes #762 

Adjusted the regex to be more strict in general as there were some invalid cases that were being allowed with the previous regex (even though they would be caught later in the validation process). 

Added tests for relevant passing and failing cases

--

I also verified manually that square brackets are allowed in filenames in both unix and windows systems, so that is a non-issue.

### DIFF
--- a/packages/core/src/internal/utils/identifier-validators.ts
+++ b/packages/core/src/internal/utils/identifier-validators.ts
@@ -13,8 +13,10 @@ const solidityIdentifierRegex = /^[a-zA-Z$_][a-zA-Z0-9$_]*$/;
 /**
  * A regex capturing the solidity identifier rule but extended to support
  * the `myfun(uint256,bool)` parameter syntax
+ *
+ * This *could* be even stricter, but it works for now and covers obvious mistakes
  */
-const functionNameRegex = /^[a-zA-Z$_][a-zA-Z0-9$_,()]*$/;
+const functionNameRegex = /^[a-zA-Z0-9$_]*(\([a-zA-Z0-9$_,\[\]]*\))?$/;
 
 /**
  * Does the identifier match Ignition's rules for ids. Specifically that they
--- a/packages/core/src/internal/utils/identifier-validators.ts
+++ b/packages/core/src/internal/utils/identifier-validators.ts
@@ -16,7 +16,7 @@ const solidityIdentifierRegex = /^[a-zA-Z$_][a-zA-Z0-9$_]*$/;
  *
  * This *could* be even stricter, but it works for now and covers obvious mistakes
  */
-const functionNameRegex = /^[a-zA-Z0-9$_]*(\([a-zA-Z0-9$_,\[\]]*\))?$/;
+const functionNameRegex = /^[a-zA-Z0-9$_]+(\([a-zA-Z0-9$_,\[\]]*\))?$/;
 
 /**
  * Does the identifier match Ignition's rules for ids. Specifically that they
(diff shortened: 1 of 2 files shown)


packages/core/src/internal/utils/identifier-validators.ts
```diff
@@ -13,8 +13,10 @@ const solidityIdentifierRegex = /^[a-zA-Z$_][a-zA-Z0-9$_]*$/;
 /**
  * A regex capturing the solidity identifier rule but extended to support
  * the `myfun(uint256,bool)` parameter syntax
+ *
+ * This *could* be even stricter, but it works for now and covers obvious mistakes
  */
-const functionNameRegex = /^[a-zA-Z$_][a-zA-Z0-9$_,()]*$/;
+const functionNameRegex = /^[a-zA-Z0-9$_]*(\([a-zA-Z0-9$_,\[\]]*\))?$/;
```
kanej, line 19: Nice catch.
Are there any backwards compatibility concerns here?

kanej: Should the `*` be a `+`?
Would the following incorrectly pass `(uint256)`?

kanej (commented):
I left a few comments, but this looks good.

packages/core/src/internal/utils/identifier-validators.ts
zoeyTM: > Should the `*` be a `+`?

Ah that's a great catch, I just pushed an update to fix that (with a new test case to verify)

> Are there any backwards compatibility concerns here?

I don't believe so. The biggest potential impacts I was aware of while writing this were:

1. The rightmost `*` needs to stay an `*` and not be changed to a `+`. This change would break `myFunc()` which currently works
2. Due to how the regex was previously written, `0-9` was allowed to be in the function name, which looks unintentional, but is actually correct as that is valid solidity. I added `0-9` to the first character set to continue this behavior.

I believe these are the only two potential breaking changes of things that should work before and after the change. This technically has a breaking change in that new invalid patterns will fail, but they would have failed later in the code anyways previously, so I don't believe this is actually a breaking change. They just properly throw where intended now.

kanej: Ah I follow now. The move is all about reworking the parameter list.
We support `[]` in the parameter list and this is allows additional valid identifiers, so no breaking change there - agreed.
I sharpened up the first part in `firstpart(secondparts[])`, so that the first character can't be a number: https://docs.soliditylang.org/en/v0.8.26/grammar.html#a4.SolidityLexer.Identifier.
It could be argued that we could sharpen up the parameter list as well, but I am happy with the current level of precision. What do you think?

kanej (approved):
LGTM!

Issue #782, opened 2024-06-14
Error with viem and custom chainId (for example, a tenderly fork)


johnpm-12:
### What happened?

When using viem, and a network with a unique/custom chainId, the error NetworkNotFoundError is thrown.
The normal solution is to pass in a custom viem chain, but `hre.ignition.deploy` does not allow passing in a chain.
See the error message:
![image](https://github.com/NomicFoundation/hardhat-ignition/assets/39016062/f6f676b2-965c-42d8-8cfb-0f04748f9211)

I propose adding a parameter or field to an existing parameter on hre.ignition.deploy that allows passing a custom viem chain.

### Minimal reproduction steps

See https://github.com/johnpm-12/hardhat-ignition-tenderly-test

`npm run deploySameId` works as expected, because the chainId is the same as one in `viem/chains`

`npm run deployUniqueId` throws NetworkNotFoundError

### Search terms

NetworkNotFoundError

kanej:
Thanks for the reproduction repo, it really helps.
I think the additional `chain` parameter makes sense as a solution.

johnpm-12:
@kanej see #783 I put up a PR solving this

PR #783, opened 2024-06-14
add optional chain param to viem deploy

johnpm-12:
see #782

### DIFF
--- a/packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
+++ b/packages/hardhat-plugin-viem/src/viem-ignition-helper.ts
@@ -252,8 +252,12 @@ export class ViemIgnitionHelper {
     deployedContract: { address: string },
     chain?: Chain
   ): Promise<GetContractReturnType> {
-    const publicClient = await hre.viem.getPublicClient({ chain });
-    const [walletClient] = await hre.viem.getWalletClients({ chain });
+    const publicClient = await hre.viem.getPublicClient(
+      chain ? { chain } : undefined
+    );
+    const [walletClient] = await hre.viem.getWalletClients(
+      chain ? { chain } : undefined
+    );
 
     if (walletClient === undefined) {
       throw new HardhatPluginError(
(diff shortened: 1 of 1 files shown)


zoeyTM:
I actually don't think this is needed at all. We have access to the chain ID, either by querying the network or via the custom one configured in hardhat config. All we need to do is pass the existing chain ID into the viem functions as well. No need for a new parameter. @kanej

alcuadrado:
Agreed with @zoeyTM. 

Can you explain the motivation behind a new param? Is it because of a type-level limitation?

johnpm-12:
> I actually don't think this is needed at all. We have access to the chain ID, either by querying the network or via the custom one configured in hardhat config. All we need to do is pass the existing chain ID into the viem functions as well. No need for a new parameter. @kanej

I originally considered defining the custom chain inside the deploy function automatically if the chain id doesn't match a supported viem chain. But you have to make assumptions about native currency name, symbol, and decimals. They're required params when defining a viem chain.

zoeyTM:
> > I actually don't think this is needed at all. We have access to the chain ID, either by querying the network or via the custom one configured in hardhat config. All we need to do is pass the existing chain ID into the viem functions as well. No need for a new parameter. @kanej
> 
> I originally considered defining the custom chain inside the deploy function automatically if the chain id doesn't match a supported viem chain. But you have to make assumptions about native currency name, symbol, and decimals. They're required params when defining a viem chain.
> 
> https://viem.sh/docs/chains/introduction#custom-chains
> 
> EDIT: also I figured it should support functionality similar to the other viem hardhat functions (getPublicClient, getWalletClients, etc). They solve this problem with an optional chain parameter

I think this functionality would be better served via extending hardhat config, adding the necessary fields to custom chains there, then pulling the info from `hre.config` inside the ignition helper.

johnpm-12:
> > > I actually don't think this is needed at all. We have access to the chain ID, either by querying the network or via the custom one configured in hardhat config. All we need to do is pass the existing chain ID into the viem functions as well. No need for a new parameter. @kanej
> > 
> > 
> > I originally considered defining the custom chain inside the deploy function automatically if the chain id doesn't match a supported viem chain. But you have to make assumptions about native currency name, symbol, and decimals. They're required params when defining a viem chain.
> > https://viem.sh/docs/chains/introduction#custom-chains
> > EDIT: also I figured it should support functionality similar to the other viem hardhat functions (getPublicClient, getWalletClients, etc). They solve this problem with an optional chain parameter
> 
> I think this functionality would be better served via extending hardhat config, adding the necessary fields to custom chains there, then pulling the info from `hre.config` inside the ignition helper.

And generate the custom chain internally if it's defined in the hardhat config?
What if a custom chain is not fully defined in the hardhat config? Throw an error?
I assume getpublicClient, getWalletClients, and the other hardhat viem functions would need to be updated to do this, too?

I can do it this way, I just want to make sure I understand how ya'll want it done before I get started

kanej:
I am going to close this PR as we are porting Ignition to the Hardhat repo.

We are also reworking the entirety of `hardhat.config.{js,ts}` for Hardhat 3. I think after the update of Ignition to support Hardhat 3 is the time to tackle configuration options for custom chains in `hardhat-plugin-viem`.

PR #796, opened 2024-08-22
chore:Fix the typo and replace 'this' with the class in a static method

iosh:
- Fix a typo 
- Replace 'this' with the class in a static method. For more information, refer to: https://github.com/mysticatea/eslint-plugin/blob/master/docs/rules/no-this-in-static.md.

I have a question. I'm interested in contributing to Hardhat and would like to know what I can to do.


kanej (approved):
Nice catch on the static calls. We will look and see options on adding an eslint rule.

PR #798, opened 2024-08-30
Adds deployment stamp modification

3commascapital:
useful context is lacking from artifacts. addresses are very useful, but not sufficient for all situations, such as indexers, which often desire a start block.

this will require a docs update, which i am happy to do as soon as it is communicated that this feature is desired.


kanej:
Our apologies @3commascapital, we let this one slip through the cracks.

Providing more transaction details to enable use cases like indexing is something we want to add, if you are interested in talking through the use case and the approach in this PR you can ping us [on discord](https://hardhat.org/discord).

kanej:
I am going to close this PR in preparation for the port of Hardhat Ignition over to the Hardhat repo.
The discussion on this topic continued, our next steps are looking to enhancing the `status` command and function in core to support the extra data needed here. We will pick that discussion up in the Hardhat repo.

PR #799, opened 2024-09-01
feat: JSON5 support for module parameters

### LINKED ISSUE #790: feat: Allow `... ether` to be parsed in parameters
### Describe the feature

Since we already have a custom deserializer for parameters (https://github.com/NomicFoundation/hardhat-ignition/issues/663 to parse bigint) i think it makes most sense to also parse w.r.t `parseEther` of ethers, so that instead of writing `"1000000000000000000n"` as parameter which is definitely not human-readable, I could type `1.0 ether`, based on `.endsWith` keyword it can directly parse the value.

In this way, `1.0 gwei` could also be parsed and such.

P.S. would love to work on this!

### Search terms

parameters

erhant:
- Added JSON5 parser support to be used for module parameters.
- Resolves #790 as per the discussion within the issue.
- Added integration test for JSON5 parameter parser.

JSON5 parser is used for both JSON and JSON5 files, as JSON5 is compatible with that as well. If there are any performance concerns on this, I can edit the parsing logic to adhere to file extension and use the respective parser, but I doubt performance is a concern for module parameters at this level.

> [!NOTE]
> 
> Although I didnt touch the code in any other part, test from other packages _randomly_ failed for me, so at some point I ran the JSON5 tests with: 
> 
> ```sh
> pnpm --filter ./packages/hardhat-plugin test 
> ```


packages/hardhat-plugin/src/index.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from "fs-extra";
 import { extendConfig, extendEnvironment, scope } from "hardhat/config";
 import { NomicLabsHardhatPluginError } from "hardhat/plugins";
+import { parse as json5parse } from "json5";
```
erhant, line 18: I had to import like this to make the linter happy, I hope that is okay

zoeyTM: Can you move this down into the deploy task function instead? It should go on line 119 above the `prompts` import like this: 

```typescript
const { default: chalk } = await import("chalk");
const { parse: json5Parse } = await import("json5");
const { default: Prompt } = await import("prompts");
```

packages/hardhat-plugin/src/index.ts
```diff
@@ -672,7 +676,7 @@ async function resolveConfigPath(
   try {
     const rawFile = await readFile(filepath);
 
-    return JSON.parse(rawFile.toString(), bigintReviver);
+    return await json5parse(rawFile.toString(), bigintReviver);
```
zoeyTM, line 679: small swap to camelCase to match the updated import above

```suggestion
    return await json5Parse(rawFile.toString(), bigintReviver);
```

zoeyTM (changes requested):
This looks great, thanks for submitting the PR! Just two small changes then this should be good to merge 👍

zoeyTM:
Can you also pull the latest `development` branch from the hardhat-ignition repo and rebase your branch onto it. That should fix the CI

packages/hardhat-plugin/src/index.ts
erhant: thanks for the review & precise instructions!

im not sure about the import part, the ignition scope at line 119 does not reach the function `resolveConfigPath` at line 672 and it gives an error due to scope.

i can add it around line 676, but im not entirely familiar with dynamic imports, not sure if it causes any trouble 🤔 

also, the function `resolveParametersString` seems to parse JSON as well, should this use JSON5 too?

zoeyTM: Ah! My mistake. That's what I get for reviewing a PR at 3am 😅

Don't worry about the import, it's fine where it is. Could you change the JSON parsing in `resolveParametersString` as well, then this should be ready to merge assuming the CI passes. Thanks!

erhant:
updated! hoping that CLI passes 🙏🏻

zoeyTM:
@erhant Can you rebase your branch so that your two commits are the only commits in the PR please? It looks like something went wrong when you tried to merge our dev branch into yours and that's why the CI is failing.

zoeyTM:
to be clear, these are the only two commits that should be in the PR:

- [added json5 parsing with test](https://github.com/NomicFoundation/hardhat-ignition/pull/799/commits/c8abb526bec775884a61d9c61c0d70c04c25070b)
- [casing + resolveParametersString](https://github.com/NomicFoundation/hardhat-ignition/pull/799/commits/1ff5e404fa992b335414f011ecf132c81dc2a1a5)

erhant:
hmm okay so by syncing the fork using `merge` alone I may have caused some trouble for the rebase on my end, I should have done `git pull --rebase` most likely.

I will simply close this PR and re-open it with a cleaner branch, with a single commit on top of [f3dc8ad](https://github.com/NomicFoundation/hardhat-ignition/commit/f3dc8ad288654085868aebfd8336a21f5dcaaeed) that is the latest `development` branch.

Here is the new PR: https://github.com/NomicFoundation/hardhat-ignition/pull/800

PR #800, opened 2024-09-05
feat: JSON5 support for module parameters

### LINKED ISSUE #790: feat: Allow `... ether` to be parsed in parameters

erhant:
**Re-opened PR with updated base**

Original PR message from #799 below:

- Added JSON5 parser support to be used for module parameters.
- Resolves #790 as per the discussion within the issue.
- Added integration test for JSON5 parameter parser.

JSON5 parser is used for both JSON and JSON5 files, as JSON5 is compatible with that as well. If there are any performance concerns on this, I can edit the parsing logic to adhere to file extension and use the respective parser, but I doubt performance is a concern for module parameters at this level.

> [!NOTE]
> 
> Although I didnt touch the code in any other part, test from other packages _randomly_ failed for me, so at some point I ran the JSON5 tests with: 
> 
> ```sh
> pnpm --filter ./packages/hardhat-plugin test 
> ```


zoeyTM (approved):
LGTM 👍 Thanks for fixing the commit history. I added a commit fixing the lockfile, so once the CI passes I'll go ahead and merge!

Issue #791, opened 2024-08-05
Default deployment


SebastienGllmt:
### Describe the feature

Right now, when testing locally, I always have to do

```bash
npx hardhat node
npx hardhat ignition deploy ./ignition/modules/... --network localhost
```

it's annoying to have to open two terminals and do an extra command just for this, and it's easy to accidentally forget to run the deploy script (esp. for new contributors to our project trying to setup their env for the first time)

Ideally, there would be a way in my hardhat config to specify a default hardhat ignition deployment to run when using the localhost network so that people don't have to type this manually

This is somewhat philosophically similar to #752 

### Search terms

deploy, node, default

kanej:
I like the idea of starting a local Hardhat node already setup with an Ignition module.
This would need some design work around how it composes with the Hardhat accounts system (we are looking at reworking this in the next major Hardhat upgrade).

fvictorio:
As a temporary workaround, you can do something like this:

```js
const { subtask } = require("hardhat/config")
const { TASK_NODE_SERVER_READY } = require("hardhat/builtin-tasks/task-names");
const path = require("path")

subtask(TASK_NODE_SERVER_READY, async (_, hre, runSuper) => {
  const result = await runSuper();

  await hre.run({
    scope: "ignition",
    task: "deploy"
  }, {
    modulePath: path.resolve(__dirname, "ignition", "modules", "Foo.js")
  })

  return result;
});
```

This will only work for the `hh node` task, so you won't get deployments for, say, `hh console`, but it might help until we have a better solution.

SebastienGllmt:
I tried the workaround suggested, and the issues I have is that calling this task directly does not generate a `deployments` folder, because it uses `EphemeralDeploymentLoader` instead of `FileDeploymentLoader`

This isn't good for me use-case because I need to get the addresses of the deployed contracts, but `_deployedAddresses` is not exposed anywhere in the `EphemeralDeploymentLoader` case (ideally I would either have a file on-disk to read, or a deployment-agnostic way to get the list of addresses deployed so I can write a tool that processes these files regardless of it this is a localhost deployment or a mainnet deployment)

fvictorio:
Ah, interesting!

@kanej is there an option to "force" ignition to write deployments to disk, even if the chain is the in-process network?

kanej:
We don't have a way of forcing Ignition to ignore that it is running against the Hardhat node.
[...]

PR #809, opened 2024-09-21
Force write the deployment information to disk

SebastienGllmt:
This PR is to help with #791 

Notably, with this code change, you can now force writing the deployment information to disk (even when using the hardhat network) by overriding the deploy task like this

```ts
function forceToDisk() {
  scope('ignition').task(
    'deploy',
    async (taskArguments: IgnitionDeployParameters, hre, runSuper) => {
      const network = hre.hardhatArguments.network ?? hre.config.defaultNetwork ?? 'hardhat';
      // see https://github.com/NomicFoundation/hardhat-ignition/issues/791
      if (taskArguments.deploymentId == null && network === 'hardhat') {
        taskArguments.deploymentId = taskArguments.deploymentId ?? 'chain-31337';
      }
      const result = await runSuper(taskArguments);
      return result;
    }
  );
}
```

### DIFF
--- a/packages/hardhat-plugin/src/index.ts
+++ b/packages/hardhat-plugin/src/index.ts
@@ -147,7 +147,8 @@ ignitionScope
       const deploymentId = resolveDeploymentId(givenDeploymentId, chainId);
 
       const deploymentDir =
-        hre.network.name === "hardhat"
+        // allow overriding givenDeploymentId to force writing to disk in in-memory mode
+        hre.network.name === "hardhat" && givenDeploymentId === undefined
           ? undefined
           : path.join(hre.config.paths.ignition, "deployments", deploymentId);
 


zoeyTM (changes requested):
I think this is a bit unclear and overcomplicated for a solution. I think a much simpler solution would be to just add a flag instead, something like this:

```typescript
// hardhat-plugin/src/type-extensions.ts
    ignition?: Partial<DeployConfig> & {
      strategyConfig?: Partial<StrategyConfig>;
      writeLocalhostDeployment?: boolean;
    };
  }

  export interface HardhatConfig {
    ignition: Partial<DeployConfig> & {
      strategyConfig?: Partial<StrategyConfig>;
      writeLocalhostDeployment?: boolean;
    };
  }
```

alcuadrado:
Maybe `.addFlag()` in the task itself instead?

zoeyTM:
Closing in favor of https://github.com/NomicFoundation/hardhat-ignition/pull/816

PR #810, opened 2024-09-23
SVG zoom for visualization graph

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/806
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/807

- using `svg-pan-zoom` library to zoom and pan mermaid diagram
- set `maxTextSize` to `5e5` instead of the default of `5e4`

### DIFF
--- a/packages/ui/src/components/mermaid.tsx
+++ b/packages/ui/src/components/mermaid.tsx
@@ -1,11 +1,13 @@
 import React, { useEffect, useMemo } from "react";
 import styled from "styled-components";
+import svgPanZoom from "svg-pan-zoom";
 
 import {
   IgnitionModule,
   IgnitionModuleResult,
 } from "@nomicfoundation/ignition-core/ui-helpers";
 import mermaid from "mermaid";
+
 import { toMermaid } from "../utils/to-mermaid";
 
 export const Mermaid: React.FC<{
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/mermaid.tsx
```diff
@@ -23,6 +25,7 @@ export const Mermaid: React.FC<{
 
   useEffect(() => {
     mermaid.initialize({
+      maxTextSize: 500000,
```
alcuadrado, line 28:
```suggestion
      maxTextSize: 5000000,
```

Should we do something larger? AFAIK this is to prevent slow rendered, but I guess ignition users, being devs, would have ok-ish computers.

packages/ui/src/components/mermaid.tsx
```diff
@@ -31,6 +34,20 @@ export const Mermaid: React.FC<{
     mermaid.contentLoaded();
   });
 
+  // requestAnimationFrame is used to ensure that the svgPanZoom is called after the svg is rendered
+  useEffect(() => {
+    requestAnimationFrame(() => {
```
kanej, line 39: Can you walk me through this one.

Is the problem that you want to hold on the `svgPanZoom` invocation until you are sure that the diagram svg is rendered?

alcuadrado:
I tried this manually and it works, but I added this

```ts
  for (let i = 1; i < 20; i++) {
    basicContracts.push(
      m.contractAt("BasicContract", basicContracts[i - 1], { id: "basic" + i })
    );
  }
```

to `examples/complete/ignition/modules/CompleteModule.js`, and there are two issues:

1. The batches "buttons" overlap with each other.
2. The diagram is still 300px tall, so it's hard to read.

packages/ui/src/components/mermaid.tsx
zoeyTM: Correct. If `svgPanZoom` is run before its rendered the function will error. This ensures that its run after rendering

alcuadrado (approved):
🚀

PR #816, opened 2024-09-24
add writeLocalhostDeployment to deploy command

SebastienGllmt:
This is an alternative to #809 to also solve #791


zoeyTM (changes requested):
This looks good overall! I did add one small suggestion for the wording of the flag description. Also, can you add a unit test for this case as well. Once those two things are taken care of this looks good to merge :+1:

zoeyTM (approved):
Looks great, thanks for the PR!

SebastienGllmt:
Thanks for merging! Excited for this to be included in the next release so we can use it easily!

PR #819, opened 2024-09-30
Add support for global module parameters

### LINKED ISSUE #803: Allow creating global parameters
### Problem
Module parameters are great but sometimes you can have multiple modules requiring the exact same parameter which can become a bit tedious to manage.

For example, if I wanted to set the owner for a bunch of contracts on different modules I would have a config file like this:
```
{
  "GraphProxyAdmin": {
    "owner": "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0",
    "pauseGuardian": "0x95cED938F7991cd0dFcb48F0a06a40FA1aF46EBC"
  },
  "Controller": {
    "owner": "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0",
  },
  "RewardsManager": {
    "owner": "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0",
    "issuancePerBlock": "114155251141552511415n"
  }
}
```

### Solution
It would be nice to have global parameters that could be accessed from any module. Maybe use the `global` keyword to group them.

The config file in this case would be much cleaner, and the global parameters would immediately be recognized as such:

```
{
  "global": {
    "owner": "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0",
    "pauseGuardian": "0x95cED938F7991cd0dFcb48F0a06a40FA1aF46EBC"
  },
  "RewardsManager": {
    "issuancePerBlock": "114155251141552511415n"
  }
}
```

Then for accessing the global parameters the API could be something like:
```ts
const owner = m.getGlobalParameter('owner')
```

### Search terms

global, parameter, module parameter

zoeyTM:
resolves #803

### DIFF
--- a/packages/core/src/internal/utils/resolve-module-parameter.ts
+++ b/packages/core/src/internal/utils/resolve-module-parameter.ts
@@ -23,7 +23,8 @@ export function resolveModuleParameter(
   }
 
   const moduleParameters =
-    context.deploymentParameters[moduleParamRuntimeValue.moduleId];
+    context.deploymentParameters[moduleParamRuntimeValue.moduleId] ??
+    context.deploymentParameters.$global;
 
   if (moduleParameters === undefined) {
     assertIgnitionInvariant(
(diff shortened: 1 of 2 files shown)


alcuadrado (approved):
🚀

PR #821, opened 2024-10-08
Add CLI command to list transactions for a given deployment

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/805

Screen Shot 2024-10-08 at 4 25 53 AM
Screen Shot 2024-10-08 at 4 26 14 AM


packages/core/src/list-transactions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+            break;
+          }
+          case ExecutionSateType.CALL_EXECUTION_STATE: {
+            transactions.push({
+              type: exState.type,
+              from: exState.from,
+              txHash: transaction.hash,
+              status: getTransactionStatus(transaction),
+              name: exState.functionName,
```
alcuadrado, line 94: Can we also display the contract name?

packages/core/src/list-transactions.ts
```diff
@@ -0,0 +1,144 @@
...
+    exState.type === ExecutionSateType.DEPLOYMENT_EXECUTION_STATE ||
+    exState.type === ExecutionSateType.CALL_EXECUTION_STATE ||
+    exState.type === ExecutionSateType.SEND_DATA_EXECUTION_STATE
+  );
+}
+
+function getTransactionStatus(transaction: Transaction): TransactionStatus {
+  const status =
+    transaction.receipt === undefined
+      ? TransactionStatus.DROPPED
```
alcuadrado, line 138: Can't this also mean pending? I guess dropped would be if that networkInteraction has another transaction after this one.

alcuadrado (commented):
This is looking good. I left 3 comments.

packages/core/src/list-transactions.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const deploymentState = await loadDeploymentState(deploymentLoader);
+
+  if (deploymentState === undefined) {
+    throw new IgnitionError(ERRORS.LIST_TRANSACTIONS.UNINITIALIZED_DEPLOYMENT, {
+      deploymentDir,
+    });
+  }
+
+  const executionOrder = await getExecutionOrder(deploymentLoader);
```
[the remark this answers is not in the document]
zoeyTM: I'm confused because I'm pretty sure the logic you wrote is doing the exact same thing as the logic I wrote? 

```
for (const journalMessage of journal) {
   if (!isSendTrasanction(journalMessage))  continue;
   const [future, networkInteraction, transaction] = somehowGetTheData(journalMessage);
   // same logic
}
```

```typescript
export async function getExecutionOrder(
  deploymentLoader: DeploymentLoader
): Promise<string[]> {
  const futureIds: string[] = [];

  // get each journal message in the order they were written, same as above
  for await (const message of deploymentLoader.readFromJournal()) {
    // ensures the message is relevant and not journal metadata. also makes sure each entry to the array is unique
    if ("futureId" in message && !futureIds.includes(message.futureId)) {
      futureIds.push(message.futureId);
    }
  }

  return futureIds;
}
```

```typescript
// iterating through the future IDs we just retrieved, in the same order they were written to the journal
 for (const futureId of executionOrder) {
    // this is the "somehowGetTheData" part of your snippet
    const exState = deploymentState.executionStates[futureId];

    // skip if this future doesn't send transactions, just like your snippet
    if (!doesSendTransactions(exState)) {
      continue;
    }
```

am I misunderstanding?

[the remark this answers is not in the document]
zoeyTM: update: i was in fact misunderstanding :)

packages/core/src/list-transactions.ts
```diff
@@ -48,86 +48,94 @@ export async function listTransactions(
...
-        }
+    const exState = deploymentState.executionStates[message.futureId];
+
+    assertIgnitionInvariant(
+      doesSendTransactions(exState),
+      "Expected execution state to be a type that sends transactions"
+    );
+
+    const networkInteraction =
+      exState.networkInteractions[message.networkInteractionId - 1];
```
alcuadrado, line 66: This `-1` caught me by surprise tbh. Do we always have the network interaction ids be 1, 2, 3, ...

TBH I don't remember, and I couldn't find it easily in the code.

alcuadrado: I search the code, and we normally do

```ts
const onchainInteraction = executionState.networkInteractions.find(
    (interaction) => interaction.id === networkInteractionId
  );
```

zoeyTM: The `-1` is because i'm indexing it directly, not using a `find`. As far as I know, we've always started the networkInteraction ID's at 1

PR #822, opened 2024-10-09
fix: fqn resolution of artifacts

### LINKED ISSUE #778: Fully-qualified artifact name resolution seems to not actually work
### What happened?

When I have two contracts with the same name (let's say "MyContract", in different .sol files in different paths) the resolution in the `m.contract("MyContract", [], ...)` future prompts to use fully-qualified artifact names, but then when I try that it is not accepted/found by ignition. So there's no way to discriminate between artifacts with the same name (I agree that, however, having artifacts with the same name is not necessarily a good practice, but they might come somehow from different libraries).

### Minimal reproduction steps

Hardhat version: 2.22.5.
Hardhat Ignition version: 0.15.4 (at least by checking the package-lock.json).
Also using: `@openzeppelin/contracts`.

1. I created a contract named MyOwnedERC20, located at contracts/MyOwnedERC20.sol. I don't believe the contents matter but for the purpose of this discovery the contents are these:

```
// SPDX-License-Identifier: MIT
pragma solidity 0.8.24;
import "@openzeppelin/contracts/access/Ownable.sol";
import "@openzeppelin/contracts/token/ERC20/ERC20.sol";


contract MyOwnedERC20 is ERC20, Ownable {
    constructor() ERC20("My Token", "MYTK") Ownable(msg.sender) {}
    function mint(address account, uint256 amount) onlyOwner public {
        _mint(account, amount);
    }
}
```

2. Out of curiosity, I copied the contract exactly into a new directory (yes, two copies of the same file): contracts/subdirectory/MyOwnedERC20.sol.

3. Then I built: `npx hardhat compile`.

4. Then, create a hardhat ignition module. In this example, ignition/modules/MyOwnedERC20.js. The contents are simple:

```
const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");

module.exports = buildModule("MyOwnedERC20", (m) => {
  // const contract = m.contract(
  //     "contracts/subdirectory/MyOwnedERC20.sol:MyOwnedERC20", [], {id: "MyOwnedERC20_1"}
  // );
  const contract = m.contract(
      "MyOwnedERC20", [], {id: "MyOwnedERC20_1"}
  );

  return { contract };
});
```

Notice how there are 3 lines intentionally commented. We'll make use of it in the next steps.

5. Try it: `npx hardhat ignition deploy ignition/modules/MyOwnedERC20.js`. You'll see an error:

```
Error HH701: There are multiple artifacts for contract "MyOwnedERC20", please use a fully qualified name.

Please replace MyOwnedERC20 for one of these options wherever you are trying to read its artifact:

contracts/MyOwnedERC20.sol:MyOwnedERC20
contracts/subdirectory/MyOwnedERC20.sol:MyOwnedERC20


For more info go to https://hardhat.org/HH701 or run Hardhat with --show-stack-traces
```

(this is a standard hardhat error and it is understandable, so no problem here)

6. Edit the MyOwnedERC20.js deployment file. Comment the m.contract invocation and uncomment the other m.contract invocation:

```
const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");

module.exports = buildModule("MyOwnedERC20", (m) => {
  const contract = m.contract(
      "contracts/subdirectory/MyOwnedERC20.sol:MyOwnedERC20", [], {id: "MyOwnedERC20_1"}
  );
  // const contract = m.contract(
  //     "MyOwnedERC20", [], {id: "MyOwnedERC20_1"}
  // );

  return { contract };
});
```

7. Try it: `npx hardhat ignition deploy ignition/modules/MyOwnedERC20.js`. You'll see another error now:

```


Hardhat Ignition 🚀

Resuming existing deployment from ./ignition/deployments/chain-31337

Deploying [ MyOwnedERC20 ]

Batch #1
  Executing MyOwnedERC20#MyOwnedERC20_1...

Error in plugin hardhat-ignition: Artifact path not found for contracts/subdirectory/MyOwnedERC20.sol:MyOwnedERC20

For more info run Hardhat with --show-stack-traces
```

**I expect**: the fully-qualified artifact path to work.

### Search terms

duplicate artifact name

kanej:
In the artifact resolver detect whether a FQN is being used to refer to the contract and use Hardhat's own build info resolution method (which assumes an FQN).

This is now covered with an additional integration test for repos with multiple contracts with the same name in different files and an Ignition module where those contracts are referred to by FQNs.

Fixes #778.

### DIFF
--- a/packages/hardhat-plugin/src/hardhat-artifact-resolver.ts
+++ b/packages/hardhat-plugin/src/hardhat-artifact-resolver.ts
@@ -14,6 +14,17 @@ export class HardhatArtifactResolver implements ArtifactResolver {
   public async getBuildInfo(
     contractName: string
   ): Promise<BuildInfo | undefined> {
+    // If a fully qualified name is used, we can can
+    // leverage the artifact manager directly to load the build
+    // info.
+    if (this._isFullyQualifiedName(contractName)) {
+      return this._hre.artifacts.getBuildInfo(contractName);
+    }
+
+    // Otherwise we have only the contract name, and need to
+    // resolve the artifact for the contract ourselves.
+    // We can build on the assumption that the contract name
+    // is unique based on Module validation.
     const artifactPath = await this._resolvePath(contractName);
 
     if (artifactPath === undefined) {
@@ -50,5 +61,15 @@ export class HardhatArtifactResolver implements ArtifactResolver {
 
   public loadArtifact(contractName: string): Promise<Artifact> {
     return this._hre.artifacts.readArtifact(contractName);
+  }
+
+  /**
+   * Returns true if a name is fully qualified, and not just a bare contract name.
+   *
+   * This is based on Hardhat's own test for fully qualified names, taken
+   * from `contract-names.ts` in `hardhat-core` utils.
+   */
+  private _isFullyQualifiedName(contractName: string): boolean {
+    return contractName.includes(":");
   }
 }
--- a/packages/hardhat-plugin/test/fixture-projects/multiple-contracts-with-same-name/contracts/Rocket1.sol
+++ b/packages/hardhat-plugin/test/fixture-projects/multiple-contracts-with-same-name/contracts/Rocket1.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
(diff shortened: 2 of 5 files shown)


zoeyTM (approved):
LGTM!

PR #828, opened 2024-11-01
Allow ignition modules as a dependency for futures

### LINKED ISSUE #812: `Future` after `IgnitionModule`
## Problem

In some situations, a user needs to express that a `Future` should only be executed after an `IgnitionModule` has been executed.

We have already covered this for one case, where the `Future` depends on a `ContractFuture` exported by the `IgnitionModule`, as the `Future` implicitly depends on the entire `IgnitionModule` (what we jokingly call The Malaga rule).

There are cases where no `ContractFuture` is exported from the `IgnitionModule,` and you still need to be able to express this. Right now, this leads to a workaround where a `ContractFuture` is unnecessarily defined and exported, as shown by The Graph's module definitions.

## Solution

We could have an explicit way to express "after module".

A possible syntax could be something like:

```ts
const MyModule = require("./module.js");

// ...
const foo = m.contract("Foo", [], {after: [MyModule]});
```

This could be quickly done by adding all the futures as dependencies of `foo`, but it is probably worth making it a first-class citizen to be able to properly communicate and display it in the `visualize` task.

## For future reference

I already validated this with the team at The Graph [here](https://github.com/graphprotocol/contracts/pull/1025#discussion_r1761946130).

### Search terms

modules

zoeyTM:
resolves #812

### DIFF
--- a/packages/core/src/internal/batcher.ts
+++ b/packages/core/src/internal/batcher.ts
@@ -130,7 +130,28 @@ export class Batcher {
   ): boolean {
     const dependencies = batchState.adjacencyList.getDependenciesFor(futureId);
 
-    return [...dependencies].every(
+    return [...dependencies].every((depId) => {
+      if (/#/.test(depId)) {
+        return batchState.visitState[depId] === VisitStatus.VISITED;
+      }
+
+      return this._checkModuleDependencyIsComplete(depId, batchState);
+    });
+  }
+
+  /**
+   * This is needed because moduleIds are not present in the visit state
+   * causing an infinite loop when checking whether a depenedency is visited if that dependency is a module.
+   */
+  private static _checkModuleDependencyIsComplete(
+    moduleId: string,
+    batchState: BatchState
+  ) {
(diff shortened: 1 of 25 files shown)


FKNCOIN (commented):
K

FKNCOIN (commented):
sudo apt install ./<file>.deb

# If you're on an older Linux distribution, you will need to run this instead:
# sudo dpkg -i <file>.deb
# sudo apt-get install -f # Install dependencies

packages/core/src/internal/batcher.ts
```diff
@@ -130,7 +130,28 @@ export class Batcher {
   ): boolean {
     const dependencies = batchState.adjacencyList.getDependenciesFor(futureId);
 
-    return [...dependencies].every(
+    return [...dependencies].every((depId) => {
+      if (/#/.test(depId)) {
```
alcuadrado, line 134:
```suggestion
      // We distinguish between module and future ids here, as the future's always have `#` and the modules don't.
      if (/#/.test(depId)) {
```

alcuadrado (approved):

PR #832, opened 2024-11-15
fix: add additional validation for global parameters

kanej:
Default values for parameters masked the test cases for our global parameters feature.

This commit adds additional tests for parameters without default values and enhances the validations to support Global Parameters as an option.

## TODO

- [x] Add tests to confirm whether a module specific module parameter will work even if there is a global module parameter but that doesn't have the specified property
- [x] Look to refactor out duplicated code in the validation steps

### DIFF
--- a/packages/core/src/internal/utils/resolve-module-parameter.ts
+++ b/packages/core/src/internal/utils/resolve-module-parameter.ts
@@ -13,35 +13,28 @@ export function resolveModuleParameter(
   moduleParamRuntimeValue: ModuleParameterRuntimeValue<ModuleParameterType>,
   context: { deploymentParameters: DeploymentParameters; accounts: string[] }
 ): SolidityParameterType {
-  if (context.deploymentParameters === undefined) {
-    assertIgnitionInvariant(
-      moduleParamRuntimeValue.defaultValue !== undefined,
-      `No default value provided for module parameter ${moduleParamRuntimeValue.moduleId}/${moduleParamRuntimeValue.name}`
-    );
+  const potentialParamAtModuleLevel =
+    context.deploymentParameters?.[moduleParamRuntimeValue.moduleId]?.[
+      moduleParamRuntimeValue.name
+    ];
 
-    return _resolveDefaultValue(moduleParamRuntimeValue, context.accounts);
+  if (potentialParamAtModuleLevel !== undefined) {
+    return potentialParamAtModuleLevel;
   }
 
-  const moduleParameters =
-    context.deploymentParameters[moduleParamRuntimeValue.moduleId] ??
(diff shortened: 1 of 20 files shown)


packages/core/src/internal/utils/resolve-module-parameter.ts
```diff
@@ -13,35 +13,28 @@ export function resolveModuleParameter(
   moduleParamRuntimeValue: ModuleParameterRuntimeValue<ModuleParameterType>,
   context: { deploymentParameters: DeploymentParameters; accounts: string[] }
 ): SolidityParameterType {
-  if (context.deploymentParameters === undefined) {
-    assertIgnitionInvariant(
-      moduleParamRuntimeValue.defaultValue !== undefined,
-      `No default value provided for module parameter ${moduleParamRuntimeValue.moduleId}/${moduleParamRuntimeValue.name}`
-    );
+  const potentialParamAtModuleLevel =
```
kanej, line 16: I have switched the logic here.

Instead of looking up the module parameters and falling back to the global parameters then check if the named parameter exists within the object, we change to:

* look up the named parameter in the module level params
* look up the named parameter in the global level params
* finally use the default value

packages/core/src/internal/validation/futures/validateArtifactContractAt.ts
```diff
@@ -16,9 +17,11 @@ export async function validateArtifactContractAt(
   /* stage two */
 
   if (isModuleParameterRuntimeValue(future.address)) {
-    const param =
-      deploymentParameters[future.address.moduleId]?.[future.address.name] ??
-      future.address.defaultValue;
+    const param = resolvePotentialModuleParameterValueFrom(
```
kanej, line 20: I refactored out the logic of module parameter resolution given deployed parameters and a ModuleParameterRuntimeValue.

packages/core/test/call.ts
```diff
@@ -762,14 +762,61 @@ m.call(..., { id: "MyUniqueId"})`
         (v) => v.type === FutureType.CONTRACT_CALL
       );
 
-      await assert.isFulfilled(
```
kanej, line 765: `assert.isFulfilled` would succeed even if the promise errored (as technically the promise is complete). I switched to getting the result and asserting that there were no validation failures.

zoeyTM (approved):
Left one comment, but I don't think any action needs to be taken. This looks good to merge.

packages/core/test/call.ts
kanej: You are correct I mispoke. The issue is that the underlying validation could fail and the test could still pass.
Switching to asserting on the return result allowed the test to fail in the presence of unexpected validation failures.

PR #836, opened 2024-11-20
Fix validation for global params with no default value

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/834


packages/core/src/internal/validation/futures/validateNamedContractDeployment.ts
```diff
@@ -65,7 +65,8 @@ export async function validateNamedContractDeployment(
 
   const missingParams = moduleParams.filter(
     (param) =>
-      deploymentParameters[param.moduleId]?.[param.name] === undefined &&
+      (deploymentParameters[param.moduleId]?.[param.name] === undefined &&
+        deploymentParameters.$global[param.name]) === undefined &&
```
kanej, line 69: I think this is a bug, the additional brackets are not going to give the value expected.

packages/core/src/internal/validation/futures/validateNamedContractDeployment.ts
```diff
@@ -80,6 +81,7 @@ export async function validateNamedContractDeployment(
   if (isModuleParameterRuntimeValue(future.value)) {
     const param =
       deploymentParameters[future.value.moduleId]?.[future.value.name] ??
+      deploymentParameters.$global[future.value.name] ??
```
kanej, line 84:
```suggestion
      deploymentParameters.$global?.[future.value.name] ??
```

I think we need to guard against there being no global parameters passed.

kanej (changes requested):
We need tests around the new validation paths.
We should also add tests explicitly for the no default value in a module parameter but a valid entry in the global module parameters.

kanej:
I closing this in favour of: https://github.com/NomicFoundation/hardhat-ignition/pull/832

PR #837, opened 2024-11-20
fix serialization of bigints in transacitons command

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/830

### DIFF
--- a/packages/core/src/list-transactions.ts
+++ b/packages/core/src/list-transactions.ts
@@ -95,8 +95,10 @@ export async function listTransactions(
                 ? exState.result.address
                 : undefined
               : undefined,
-          params: exState.constructorArgs,
-          value: networkInteraction.value,
+          params: exState.constructorArgs.map((arg) =>
+            typeof arg === "bigint" ? arg.toString() : arg
+          ),
+          value: networkInteraction.value.toString(),
         });
 
         break;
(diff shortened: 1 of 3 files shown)


packages/core/src/list-transactions.ts
```diff
@@ -95,8 +95,10 @@ export async function listTransactions(
                 ? exState.result.address
                 : undefined
               : undefined,
-          params: exState.constructorArgs,
-          value: networkInteraction.value,
+          params: exState.constructorArgs.map((arg) =>
```
kanej, line 98: The type of `params` is `SolidityParameterType[]`, which supports `bigint`.

I think we should leave the mapping here as is and deal with the issue of bigin serialization in `JSON.stringify` in:

https://github.com/NomicFoundation/hardhat-ignition/blob/44ebf75ad3d0ef47f6acc7022ceac03f0e8454c0/packages/hardhat-plugin/src/ui/helpers/calculate-list-transactions-display.ts#L30

packages/core/src/list-transactions.ts
```diff
@@ -132,7 +136,7 @@ export async function listTransactions(
             lastTxIndex === networkInteraction.transactions.length - 1
           ),
           to: networkInteraction.to,
-          value: networkInteraction.value,
+          value: networkInteraction.value.toString(),
```
kanej, line 139: Similarly I think the value of network interaction is a `bigint`, lets keep that when we are passing data around rather than letting a UI concern leak in here.

kanej (changes requested):
Lets add a test explicitly for bigints in the parameters of a transaction.

kanej (approved):
I think we could consider exploring other ways of out putting the params in the future rather than stringify. We convert bigints to strings here in a way that may be un-intuitive to users.

PR #843, opened 2024-11-27
Enable ignition UI for deployments via hardhat scripts

zoeyTM:
Resolves https://github.com/NomicFoundation/hardhat-ignition/issues/839

Tested whatever situations I could think of and everything looked okay. Below is a sample output of a run on a local non-ephemeral network with an immediate follow up run that deploys nothing:

Screen Shot 2024-11-27 at 3 02 42 AM

### DIFF
--- a/packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
+++ b/packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
@@ -1,5 +1,6 @@
 import {
   HardhatArtifactResolver,
+  PrettyEventHandler,
   errorDeploymentResultToExceptionMessage,
   resolveDeploymentId,
 } from "@nomicfoundation/hardhat-ignition/helpers";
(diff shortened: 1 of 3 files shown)


alcuadrado:
Zoey and I reviewed this, and decided we should add a boolean flag to control whether the UI should be displayed or not, which should default to false. We'll document this in the scripts guide, but it will work out of the box for test use cases (i.e. no unexpected output in tests).

packages/hardhat-plugin-ethers/src/ethers-ignition-helper.ts
```diff
@@ -134,10 +135,13 @@ export class EthersIgnitionHelper {
             deploymentId
           );
 
+    const executionEventListener = new PrettyEventHandler();
```
alcuadrado, line 138: The flag would control if this is initialized or not.

Same for the viem plugin.

alcuadrado (approved):

PR #846, opened 2024-12-04
Fix status command when using external artifacts

### LINKED ISSUE #797: Ignition status command does not work with external artifacts deployed
### Version of Hardhat

2.22.8

### What happened?

Same setup as in NomicFoundation/hardhat-ignition#823. I'm using ignition to deploy external artifacts.

After successful deployment I ran `npx hardhat ignition status op-sepolia-1` I received this output:
```
Error HH700: Artifact for contract "Orec" not found. 

For more info go to https://hardhat.org/HH700 or run Hardhat with --show-stack-traces
```

If I run the deploy command again it says `Nothing new to deploy based on previous execution stored in ./ignition/deployments/op-sepolia-1" (as expected) so it seems that deploy command finds the required artifacts but status command does not?

Expected status command to print addresses of deployed contracts.

### Minimal reproduction steps

1. Create sample hardhat project 1, compile it;
2. Create hardhat project 2, create ignition module that tries to deploy artifact from project 1 (as described [here](https://hardhat.org/ignition/docs/guides/creating-modules#using-existing-artifacts));
3. Use ignition to deploy to testnet (I was deploying to OP Sepolia)
4. Try `ignition status <deployment-id>` command 

### Search terms

ignition

zoeyTM:
resolves #797 

Fallback to `deploymentLoader` artifact resolution for external artifact resolution


packages/core/src/status.ts
```diff
@@ -38,9 +38,19 @@ export async function status(
...
-    );
+    let artifact: Artifact<Abi>;
+
+    try {
+      artifact = await artifactResolver.loadArtifact(
+        deployedContract.contractName
+      );
+    } catch (e) {
+      if (e instanceof Error && /HH700/g.test(e.message)) {
+        artifact = await deploymentLoader.loadArtifact(deployedContract.id);
```
alcuadrado, line 49: We could instead always call the deploymentLoader as the status is about a deployment, and the artifacts are gonna be stored already.

alcuadrado: We can change `artifactResolver: Omit<ArtifactResolver, "getBuildInfo">` to `_artifactResolver?: Omit<ArtifactResolver, "getBuildInfo">` so we keep it backward compatible and also indicate that it won't be used.

PR #866, opened 2025-06-11
Exclude Core Chain and its testnet from zero fee configuration in gas fee logic

eric0x01:
Following on from #755, Updated the gas fee configuration logic to accommodate the Core Chain (chainId 1116) and testnet (chainId 1114) to exclude from zero fee configuration.


zoeyTM:
Thanks for the PR! 

I'm closing this in favor of [a more flexible solution](https://github.com/NomicFoundation/hardhat/pull/6857) we've been meaning to add for a while that will let users set their own gas settings when needed, instead of relying on submitting PRs for each relevant chain. 

This should be available in the next Hardhat Ignition release this week!

